PR #65, opened 2022-04-25
Setup relay

marcellmueller:
Added basic example page with create application mutation you can access at http://localhost:3000/relay-example


BryanGK (approved):
Looks good to me

PR #66, opened 2022-04-26
Node convict

marcellmueller:
@matthieu-foucault Thanks for the review, I have made the changes you requested and also refactored setup-pg url builder using the node-convict vars.


app/config.ts
```diff
@@ -0,0 +1,69 @@
+'use strict';
```
matthieu-foucault, line 1: You may want to keep this file a `js` file and a CommonJS module (i.e use require/module.exports instead of import/export) in case you need to pass some config to the client via the `publicRuntimeConfig` in the `next.config.js` (in which case this file wouldn't go through babel/tsc)

app/config.ts
```diff
@@ -0,0 +1,69 @@
...
+  },
+  PGHOST: {
+    doc: 'Postgres host',
+    format: String,
+    default: 'localhost',
+    env: 'PGHOST',
+  },
+  PGPORT: {
+    doc: 'Application port',
+    format: Number,
```
matthieu-foucault, line 45:
```suggestion
    format: 'port',
```

matthieu-foucault (commented):
As you were mentioning, it seems a bit cumbersome at first glance. Removing the defaults handling from your server files should improve this. 
The main benefit is to have a single file documenting all the env variables and arguments that are supported by your application. This will be extremely valuable when new developers onboard this repo, as they can easily discover the configuration options.

The config file should also be the one place where you define the configuration options validation and rules. For instance, we have testing feature flags in our application to bypass authorization, and we enforce that those flags are not set if a `NAMESPACE` env variable is define (i.e. if the application is deployed on openshift)

app/backend/lib/setup-pg.ts
```diff
@@ -1,24 +1,14 @@
...
-  databaseURL += '@';
-
-  databaseURL += process.env.PGHOST || 'localhost';
-  if (process.env.PGPORT) {
-    databaseURL += `:${process.env.PGPORT}`;
-  }
-
-  databaseURL += '/';
-  databaseURL += process.env.PGDATABASE || 'ccbc';
+  const databaseURL = `postgres://${PGUSER}:${PGPASSWORD}@${PGHOST}:${PGPORT}/${PGDATABASE}`;
```
matthieu-foucault, line 11:
```suggestion
  const databaseURL = `postgres://${PGUSER}${PGPASSWORD ? ':' : ''}${PGPASSWORD}@${PGHOST}:${PGPORT}/${PGDATABASE}`;
```

The `:` between user and password should only be added if there is a password, which typically won't be the case in your local development environment

app/config.js
```diff
@@ -0,0 +1,67 @@
...
+  },
+});
+
+// Load environment dependent configuration
+const env = config.get('NODE_ENV');
+config.loadFile('./' + env + '.json');
+
+config.validate({ allowed: 'warn' });
+
+export default config;
```
matthieu-foucault, line 67:
```suggestion
module.exports = config;
```

app/lib/relayEnvironment.ts
```diff
@@ -1,4 +1,4 @@
-const { Environment, Network, RecordSource, Store } = require('relay-runtime');
+import { Environment, Network, RecordSource, Store } = require('relay-runtime');
```
matthieu-foucault, line 1:
```suggestion
import { Environment, Network, RecordSource, Store } from 'relay-runtime';
```

eazeredo:
@marcellmueller , can you connect this PR to its issue?

marcellmueller:
> @marcellmueller , can you connect this PR to its issue?

Done :sweat_smile:

eazeredo:
Did you connect?

Is the #73?

BryanGK (approved):

PR #102, opened 2022-05-03
70 create helm chart

marcellmueller:
The main app pod only has 1 replica for now due to resource limits.


BryanGK (approved):
This looks great. Good work, Marcel!

PR #109, opened 2022-05-06
Navbar and Footer

BryanGK:
This PR implements the navbar and footer for the app. Also, is properly implementing the _document.tsx and moves Head into _app.tsx, favicons and app title.


marcellmueller (approved):
Looking good :sunglasses:

PR #113, opened 2022-05-09
Form Spike

BryanGK:
This PR is for researching and setting up the Form using @rjsf/core taken from the `cas-cif` project


app/tests/components/Form/RadioWidget.test.tsx
```diff
@@ -0,0 +1,10 @@
+import RadioWidget from '../../../lib/theme/widgets/RadioWidget';
+
+describe('RadioWidget', () => {
```
BryanGK, line 3: @wenzowski @marcellmueller I would like to get your input on these 3 tests I framed out. Would you add/remove any? Thoughts on testing strategies?

marcellmueller (approved):

PR #115, opened 2022-05-09
74 bc gov typography

marcellmueller:
Okay I added the typography and did a small refactor, I hope you don't mind but I think it's best practice to keep `_app.tsx` nice and clean. I pulled the header and nav/footer stuff into the component `Layout.tsx` and moved the `GlobalStyles` into it's own file in the `/styles` folder.

I also added an `index.ts` into the `/components` folder so we have the option to do a destructured module import on one line and keep our imports cleaner if we wish. 

Let me know if you have any feedback or would like any changes, none of this is set in stone and I'm happy to revert some things if you want.


app/types.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-typography';
```
BryanGK, line 1: What is the idea behind this module declaration? I was having some trouble using the 'react-typography' module, was this needed to make it work?

BryanGK (approved):
This looks great!

PR #127, opened 2022-05-13
Form widget testing

BryanGK:


marcellmueller:
@BryanGK I'm far from an expert but these look like some good tests to build around these! Maybe add some basic style tests? Here is a super simple one I did:

https://github.com/bcgov/CONN-CCBC-portal/blob/main/app/tests/components/StyledGovButton.test.tsx

Another style test I thought might be good for things like text areas would be to see if the on focus styles such as borders work?

wenzowski:
A bit hard to comment on this one since the tests are already merged. Perhaps we should look at adopting our own [git style guide](https://github.com/pagarme/git-style-guide/blob/master/README.md)?

Your default example (going via `createMocks`) is required. So you're covering that use case in the first two proposed tests (with and without errors). Do we know what errors we're trying to test? Given it's a radio presumably the only validation is that we've filled the required field, right? We probably want to contemplate any other errors that could be presented to the user in the first test.

I'm guessing what we really want to test without required is that we can skip the field?

Wouldn't we need to assert that all items in the `enumOptions` array are rendered as part of all prior tests? Pseudocode comments laying out our test plan for each test can help with this. Same with returning the correct value, wouldn't we want to assert that we get the correct value in each test (we get a value when there are errors too) and assert there's no value when the field isn't required?

BryanGK:
> A bit hard to comment on this one since the tests are already merged. Perhaps we should look at adopting our own [git style guide](https://github.com/pagarme/git-style-guide/blob/master/README.md)?
>
Why does having the tests merged make it difficult to comment? And how would having the style guide help that?
> 
> Your default example (going via `createMocks`) is required. So you're covering that use case in the first two proposed tests (with and without errors). Do we know what errors we're trying to test? Given it's a radio presumably the only validation is that we've filled the required field, right? We probably want to contemplate any other errors that could be presented to the user in the first test.
> 
> I'm guessing what we really want to test without required is that we can skip the field?
> 
That seems a reasonable plan, can you show me an example of what a test of that sort may look like or how best to go about thinking about implementation?
> Wouldn't we need to assert that all items in the `enumOptions` array are rendered as part of all prior tests? Pseudocode comments laying out our test plan for each test can help with this. Same with returning the correct value, wouldn't we want to assert that we get the correct value in each test (we get a value when there are errors too) and assert there's no value when the field isn't required?
>
Again this seem a reasonable approach, any tips for implementing the tests?

PR #135, opened 2022-05-16
creates dev help request issue template

BryanGK:


.github/ISSUE_TEMPLATE/dev-help-request-template.md
```diff
@@ -0,0 +1,44 @@
+---
+name: Developer Help Request Template
+about: Default developer help request template template
```
marcellmueller, line 3: @BryanGK 

Yeah I think this is a great idea :tada: Even just taking the time to clearly out what our issue is will help us to understand it better, and likely get better help. I guess my only question is the workflow for this is to open up and issue and get help/comments on it then close it?

Just noticed template written twice on this line as well as the spelling mistake in the commit title to cleanup.

BryanGK: Yeah we would probably need our own column in zenhub for these issues so they don't muddy the waters

BryanGK: Then close the issue once they get resolved

marcellmueller (approved):
Nice one! Minor request but can you squash this on merge because of that minor spelling error  :sweat_smile:

PR #136, opened 2022-05-16
Form save

marcellmueller:
@BryanGK Okay you might as well review this for now. I ended up disabling a bunch of the form stuff that requires the database as that is the issue right now, it is just running the docker file with no database so it can't continue. 

That is likely out of the scope of this sprint but I am going to look into it, I saw they do it in cas so there is an example at least.


BryanGK:
> @BryanGK Okay you might as well review this for now. I ended up disabling a bunch of the form stuff that requires the database as that is the issue right now, it is just running the docker file with no database so it can't continue.
> 
> That is likely out of the scope of this sprint but I am going to look into it, I saw they do it in cas so there is an example at least.

Was that related to the e2e tests?

marcellmueller:
> > @BryanGK Okay you might as well review this for now. I ended up disabling a bunch of the form stuff that requires the database as that is the issue right now, it is just running the docker file with no database so it can't continue.
> > That is likely out of the scope of this sprint but I am going to look into it, I saw they do it in cas so there is an example at least.
> 
> Was that related to the e2e tests?

Not 100% sure because they used to run locally with the Docker image. Though due to changes made in this such as creating an application in the db when the start form button is clicked it just flat out won't work. I suppose I should open up an issue because this is some tech debt I knew was incoming.

app/schema/mutations/createApplicationMutation.ts
```diff
@@ -4,38 +4,28 @@ import { commitMutation, graphql } from 'react-relay';
...
   };
-
   commitMutation(relayEnvironment, {
     mutation,
     variables,
     onError: () => {
       return console.log('CreateApplicationMutation failed');
     },
+    onCompleted: (response) => {
+      console.log(response);
```
BryanGK, line 27: Is there a better practice for the `onCompleted` callback as opposed to a console.log?

BryanGK (changes requested):
I'm just curious about the console.log in the createApplicationMutation.ts before approving

app/schema/mutations/createApplicationMutation.ts
marcellmueller: @BryanGK yes, probably, that is actually left in there from some debugging. I couldn't figure out how to get it to actually return the response for some reason though since it wasn't a requirement for the sprint I moved on. I can clean it up though if you'd like.

BryanGK: nah it's good, we can leave it there for now as I'm sure it'll be used for something come launch

BryanGK (approved):
Let's ship it!

marcellmueller:
> Let's ship it!

:partying_face:

PR #149, opened 2022-05-19
docs: moves PR template to .github

BryanGK:


marcellmueller (approved):
From some articles I looked up it seems like this should work, if not an all caps one in the root of /.github

PR #158, opened 2022-05-27
Update relay structure

marcellmueller:
Alright, take a look at this and let's merge it in so we can work together on getting the login working today. This contains not only the updated relay stuff but also some updated CI/CD stuff I took from cif as well.


BryanGK (approved):
This looks great! Ship it!

Issue #180, opened 2022-06-02
NextJS/Relay SSR session handling


marcellmueller:
- [x] Implement SSR session handling
- [x] Use current users sub to create and load forms
- [x] Use session to conditionally render login/logout button in nav
- [x] Check if user is logged in on page load and redirect to login if not

PR #193, opened 2022-06-03
Relay ssr

marcellmueller:
This is for issue #180. It should send any non logged in user to the home page if they try and access other routes. We also now pass the session down to the `Layout` component so we can conditionally render login buttons.


BryanGK (approved):
Made some excellent improvements here, nice work!

Issue #145, opened 2022-05-19
Feature Flag


eazeredo:
## Acceptance Criteria

- [X] Determine what Feature Flag framework/system we will use. (https://www.growthbook.io/)
- [x] Implement the feature flag
- [x] "Flag" the pages that are unpolished (which are ...)
- [x] Implement the Dashboard for the PO enable/disable features

PR #201, opened 2022-06-07
145 growthbook

marcellmueller:
This is for #145 feature flagging. Configured growthbook and added individual form page feature flagging. I can walk you through it but you can see how I use the features in ApplicationForm and pass them to the schema where we conditionally render pages.

If you log into growthbook and click features you can see how to toggle them on and off. Seems to require either refreshing the page a few times or opening it in a new browser window and starting the form again before they turn on/off.


BryanGK (approved):
Looks good!

Issue #197, opened 2022-06-06
Implement Relay jest testing setup


marcellmueller:
We will need to implement Relay jest testing setup for pages and reference how cif does it.

PR #211, opened 2022-06-09
feat: add basic pages ssr tests

marcellmueller:
These are very basic jest tests since the old ones broke once we implemented Relay SSR. Currently only doing basic unauthenticated tests, I will likely have to do up a mock relay auth at some point.


eazeredo:
Is this related to card #197?

AntBush:
Looks good to me, good first step for relay testing

PR #218, opened 2022-06-13
feat: update growthbook features on route change

marcellmueller:
Update Growthbook features on route change. Still seems to take about 10-30 seconds after I've turned a feature on/off but I've tested it a bunch and it seems to work well, should make for a better experience.


BryanGK (approved):
Way to go! This is a great improvement!

PR #227, opened 2022-06-14
GraphQL Schema Sync Guardrails

BryanGK:
This PR is ensure that we validate the graphql schema is in sync with the sqitch schema definitions + graphile plugins


BryanGK:
@wenzowski do you have examples of guardrails we can use to ensure the schema remains in sync?

wenzowski:
@BryanGK yes, though it's [a yaml for circleci](https://github.com/bcgov/cas-ciip-portal/blob/develop/.circleci/config.yml#L193) not a github action the steps to run should be roughly the same: boot the db, deploy the schema, regenerate the schema using checked in app code, diff generated schema with checked in schema (error if different)

marcellmueller:
Added a new test to our workflow that deploys the database schema, builds the graphql schema and does a `git diff` to ensure the schema committed to github is up to date.

matthieu-foucault (approved):
LGTM, maybe some refactoring to make it more efficient would be possible

.github/workflows/tests.yaml
```diff
@@ -81,3 +81,57 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -yqq sqitch libdbd-pg-perl postgresql-client
+          asdf plugin-add postgres || true
+          asdf plugin-update --all
+          asdf install postgres 14.1
+          asdf reshim
+          pg_ctl start
+          createdb ${{ env.PGDATABASE }} || true
+          sqitch rebase -y
```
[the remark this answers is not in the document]
marcellmueller: I'll remove that thanks.

marcellmueller:
> LGTM, maybe some refactoring to make it more efficient would be possible

Sounds good. Also I was wondering if there was potentially some issues using the same cache with same sha as the pgtap test would potentially cause any issues, or if I should make the schema test wait for the pgtap test so it can start up quick with the same cache.

matthieu-foucault:
> I was wondering if there was potentially some issues using the same cache with same sha as the pgtap test would potentially cause any issues, or if I should make the schema test wait for the pgtap test so it can start up quick with the same cache.

You might want a separate job, with your two jobs (pgtap and schema) depending on it.
See https://github.com/bcgov/cas-cif/blob/b90fd4352f26d82df710c9ce4dc9071da8251546/.github/workflows/test.yml#L13 for instance

.github/actions/test-env-setup/action.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -yqq sqitch libdbd-pg-perl postgresql-client
+        asdf plugin-add postgres || true
+        asdf plugin-update --all
+        asdf install postgres 14.1
+        asdf reshim
+        pg_ctl start
+        createdb ${{ env.PGDATABASE }} || true
+        sqitch rebase -y
```
marcellmueller, line 29: @matthieu-foucault I refactored this but it will currently break without the rebase. At some point we added the rebase in as a fix to a broken deployment as we figured doing a rebase wasn't a huge deal in a CI build.

matthieu-foucault: What happens without the rebase? Do you have any error logs you can share?

marcellmueller: `Cannot find change 5a5afdf81cce0a6b89b18da78f556baf0f655522 (tables/connect_session) in sqitch.plan
Error: Process completed with exit code 2.`

marcellmueller: I wonder if it would be as simple as renaming the asdf postgres cache, I remember this issue came up when Bryan did a fairly large database rework

matthieu-foucault: Right, this was an issue because the job where you created your cache was the same job that deployed the changes, effectively committing those changes in the cache. This shouldn't be an issue now that you have an `install-test-env` job

.github/actions/test-env-setup/action.yml
```diff
@@ -0,0 +1,45 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -yqq sqitch libdbd-pg-perl postgresql-client
+        asdf plugin-add postgres || true
+        asdf plugin-update --all
+        asdf install postgres 14.1
+        asdf reshim
+        pg_ctl start
+        createdb ${{ env.PGDATABASE }} || true
+        sqitch deploy
```
matthieu-foucault, line 29: With a `sqitch deploy` here, you're going to save the state of the database in your cache, which means that, because our cache key won't change when sqitch migrations change, it won't redeploy changes, any you'll have a stale schema, and , if we delete/rename a change, sqitch will exit with an error.

This action should only install the required apt packages, set env variables, and set up the caches, as it does in cas-cif: https://github.com/bcgov/cas-cif/blob/develop/.github/actions/dev-env-setup/action.yml

You should have a [separate action](https://github.com/bcgov/cas-cif/blob/develop/.github/actions/local-app-run/action.yaml) to deploy the migrations and start the app, ideally by running the docker containers that we're building instead of the apt-installed sqitch

.github/workflows/tests.yaml
```diff
@@ -8,66 +8,49 @@ env:
...
 
-      - name: Install Dependencies
+  jest:
+    needs: [install-test-env]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: dev env setup
+        uses: ./.github/actions/test-env-setup
+      - run: yarn install --frozen-lockfile
```
matthieu-foucault, line 25: yarn install should only be done once, in your `install-test-env` job. When you then run the `test-env-setup` action, your packages will be added to the job by restoring the yarn cache

matthieu-foucault (changes requested):

matthieu-foucault:
@marcellmueller I left some change requests earlier this morning

.github/workflows/tests.yaml
marcellmueller: Okay I removed `test-env-setup` action and am using `dev-env-setup` action we are using in the end to end tests, as well as do our yarn install in that action.

matthieu-foucault (approved):

matthieu-foucault (approved):

Issue #173, opened 2022-06-01
Validate Data types and MaxLength on Org profil & address


MarsRomer:
As ISED
I need to make sure that the character limits and data types are enforced
so that my data import works with the schema I have in Salesforce

This ticket is limited to the forms that have already been created or will be created in the test environment during this sprint

### Description
https://regexlib.com/REDetails.aspx?regexp_id=2790
https://owasp.org/www-community/OWASP_Validation_Regex_Repository

### Acceptance Criteria
**Organization Profile Page**
- Type of org
    - [x] Required
#
Please describe your organization type
   - [x] Required

band number
- [x] number
- [x] Required

Org name
- [x] Maxlength: 200 characters
- [x] Required

Is operating name same as legal name?
- [x] Required

Operating name
- [x] Required

Applicant organization a subsidiary?
- [x] Required

Please enter the name of the parent organization
- [x] Required

organization an Indigenous entity?
- [x] Required

short description of the Indigenous entity
- [x] maxlength: 75 characters
- [x] Required

overview of the organization
- [x] maxlength 3500 characters
- [x] Required

Date of incorporation or registration
- [x] datepicker [month & year drop down](https://reactdatepicker.com/#example-year-select-dropdown) instead where it is more clear that it is an input field?
- [x] Required

Applicant business number (9-digit business identifier provided by Canada Revenue Agency)
- [x] type: number
- [x] maxlength: 9
- [x] Required

**Organization Location Page**
Unit number
- [x] type: text input
- [x] maxlength: 9
- [x]  Optional

Street Number
- [x]  Validation: 1-9 digits and or letter, no symbols
- [ ]  Error message: Must be 9 or fewer numbers or letters, don't include symbols.
- [x]  Required

Street name
- [x]  Required

PO Box
- [x]  Validation: 1-6 digits
- [ ]  Error message: Must be 9 or fewer numbers, don't include letters or symbols.
- [x]  Optional

City
- [x]  Required

Province

- [x]  Dropdown options: "Alberta", "British Columbia", "Manitoba", "New Brunswick", "Newfoundland", "Northwest Territories", "Nova Scotia", "Nunavut", "Ontario", "Prince Edward Island", "Quebec", "Saskatchewan", "Yukon Territories"
- [x]  Required

Postal Code

- [ ]  Validation: see comments for potential solution
- [ ]  Error message: Must be in format H0H 0H0 with no spaces.
- [x]  Required

Mailing address same?

- [x]  Required
[...]

PR #229, opened 2022-06-15
Fix/misc form

BryanGK:
This PR is to fix some minor form issues that cropped up in separating out individual schema files

ref #173


wenzowski (approved):
looking great

PR #248, opened 2022-06-20
feat: add mock login config

marcellmueller:
This enables a mock login if the env variable `ENABLE_MOCK_AUTH` is set to true. Will be helpful for those end to end tests that have been falling behind since auth was implemented.

### DIFF
--- a/app/backend/lib/sso-middleware.ts
+++ b/app/backend/lib/sso-middleware.ts
@@ -8,6 +8,9 @@ const baseUrl =
     : `http://${config.get('HOST')}:${config.get('PORT') || 3000}`;
 
 let oidcIssuer: string;
+
+const mockAuth = config.get('ENABLE_MOCK_AUTH');
+
 if (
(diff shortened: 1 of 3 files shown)


app/backend/lib/sso-middleware.ts
```diff
@@ -23,6 +26,10 @@ export default async function ssoMiddleware() {
     getLandingRoute: () => {
       return '/dashboard';
     },
+    bypassAuthentication: {
+      login: mockAuth,
+      sessionIdleRemainingTime: mockAuth,
```
AntBush, line 31: Do we want to turn both options on automatically when mockAuth = true? Is there ever a case where we want `sessionIdleRemainingTime` to be `false` while `login` is `true`? From looking at the SSO repo, it seems that the check to the sessionIdleRemainingTime would return a `0`, so it would be useful to have some testing for when a user's session expires. 

Not needed for this PR I think, but something to think about nonetheless.

PR #250, opened 2022-06-21
84 budget details

marcellmueller:


app/lib/theme/ObjectFieldTemplate.tsx
```diff
@@ -1,40 +1,104 @@
...
+
+const ObjectFieldTemplate = (props: ObjectFieldTemplateProps) => {
   const DescriptionField = props.DescriptionField || DefaultDescriptionField;
+  const uiInline = props.uiSchema['ui:inline'];
+
+  const getInlineKeys = () => {
+    // Get array of inline keys so we can see if field exists in grid so we don't render it twice.
+    const inlineKeys: string[] = [];
+
+    uiInline &&
```
AntBush, line 35: Could have also just used a `props.uiSchema['ui:inline']?.map` instead of the check, but either still works.

AntBush (commented):
Looks good and everything working!

app/lib/theme/ObjectFieldTemplate.tsx
marcellmueller: For sure and I considered that but I kind of think using a logical short circuit in jsx is more readable than using the optional chaining but I like optional chaining outside of that.

I honestly don't feel super strongly either way though.

AntBush: Neither do I, I'm alright with the short circuit. In the future, we should extend the ObjectFieldTemplateProps to include the `ui:inline` key officially rather than letting it default to `any` by the type. That or, if we know what type it is, mark it as such with an `as`

marcellmueller: Sounds good thanks!

PR #252, opened 2022-06-21
Form e2e testing

marcellmueller:
Getting caught up on the form end to end test. Will keep this in draft until the remaining form pages are in and also until I refactor some things such as the `cy.wait()` I am using throughout since the page changes would break without it. Likely there is a better way to go about it.

Set environment variable `ENABLE_MOCK_AUTH=true` to test this.

### DIFF
--- a/app/cypress/integration/index.spec.js
+++ b/app/cypress/integration/index.spec.js
@@ -1,3 +1,4 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
 context('Homepage', () => {
   beforeEach(function () {
     cy.visit('/');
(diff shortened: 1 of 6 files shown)


marcellmueller:
@matthieu-foucault I rebased and updated this end to end test. It currently goes through the entire form along with most of the fields aside from some such as I haven't brought any file upload testing into this yet. 

No rush obviously but I would appreciate you taking a look some time this week and maybe some suggestions about how to refrain from relying on `cy.wait()` in between route changes. It looks like it passed this time but I imagine it will fail in CI sometimes depending on how fast github is at the moment.

app/cypress/integration/index.spec.js
```diff
@@ -10,50 +11,597 @@ context('Homepage', () => {
...
-    //     cy.get('h1').contains('Organization location');
-    //   });
-
-    // cy.get('body').happoScreenshot();
+    // Todo: find a way around using these wait
+    cy.wait(2000);
+
+    cy.get('button').contains('Go to dashboard').should('be.visible').click();
+
+    cy.waitUntil(() => cy.url().should('contain', '/dashboard'));
```
matthieu-foucault, line 19:
```suggestion
    cy.url().should('contain', '/dashboard');
```
This should do the same thing, no? Cypress will retry the assertion for a while, until it is true

app/cypress/integration/index.spec.js
```diff
@@ -10,50 +11,597 @@ context('Homepage', () => {
...
-    // cy.get('button')
-    //   .contains('Continue')
-    //   .click()
-    //   .then(() => {
-    //     cy.get('h1').contains('Organization location');
-    //   });
-
-    // cy.get('body').happoScreenshot();
+    // Todo: find a way around using these wait
+    cy.wait(2000);
```
matthieu-foucault, line 15: This one shouldn't be needed. If that's not the case we can investigate why we need to wait 2s

app/cypress/integration/index.spec.js
```diff
@@ -10,50 +11,597 @@ context('Homepage', () => {
...
+
+    // Project information page
+
+    cy.waitUntil(() =>
+      cy.get('h1').contains('Project information').should('be.visible')
+    );
+
+    cy.waitUntil(() => cy.get('[id="root_projectTitle"]').should('be.visible'));
+
+    cy.get('[id="root_geographicAreaDescription"]')
```
matthieu-foucault, line 37: I would recommend using [`@testing-library/cypress`](https://testing-library.com/docs/cypress-testing-library/intro/) to select our inputs. This will be more flexible and have the added benefit of ensuring that inputs are properly labelled

app/cypress/integration/index.spec.js
```diff
@@ -10,50 +11,597 @@ context('Homepage', () => {
...
+
+    cy.get('button').contains('Go to dashboard').should('be.visible').click();
+
+    cy.waitUntil(() => cy.url().should('contain', '/dashboard'));
+
+    // Dashboard page
+
+    cy.waitUntil(() => cy.get('h1').contains('Dashboard').should('be.visible'));
+
+    cy.get('button').contains('New application').should('be.visible').click();
```
matthieu-foucault, line 25:
```suggestion
    cy.get('button').contains('New application').click();
```
Cypress will not click on invisible elements by default, so you don't need this assertion

matthieu-foucault (commented):
We should be able to simplify that cypress test, let's take a look a that together?

app/cypress/integration/index.spec.js
```diff
@@ -10,50 +11,485 @@ context('Homepage', () => {
...
-    // cy.get('[id="root_isLegalPrimaryName-1"]');
-
-    // cy.get('button')
-    //   .contains('Continue')
-    //   .click()
-    //   .then(() => {
-    //     cy.get('h1').contains('Organization location');
-    //   });
-
-    // cy.get('body').happoScreenshot();
```
matthieu-foucault, line 50: remove commented out code

VB-at-Bis (approved):
LGTM

app/cypress/integration/index.spec.js
marcellmueller: @matthieu-foucault I'm confused, this is removed?

matthieu-foucault (approved):

PR #254, opened 2022-06-22
238 other funding sources

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/238

This one wasn't as bad as I thought, the friction was mostly figuring out that I had to nest the `uiSchema` stuff similar to how the `otherFundingSources` page schema was nested. This page is more complex than others because it dynamically allows us to add as many new groups of fields as we want with the `Add another funding source` button, as well as remove ones though that button only shows up on the second group onward.

This required me to add the custom `ArrayFieldTemplate` renderer. Even though we are unlikely to use this for another page in this project I added custom button labels using the `ui:array-buttons` prop to make it a little more generic than just for this page.

Also, while dynamically showing the form once you hit yes on the initial radio buttons isn't shown in the card I confirmed with Zoey that is how we are going to show it.

### DIFF
--- a/app/formSchema/pages/index.ts
+++ b/app/formSchema/pages/index.ts
@@ -10,6 +10,7 @@ export { default as existingNetworkCoverage } from './existingNetworkCoverage';
 export { default as mapping } from './mapping';
 export { default as organizationLocation } from './organizationLocation';
 export { default as organizationProfile } from './organizationProfile';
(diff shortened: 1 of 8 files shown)


AntBush (commented):
Looks good! Only those two comments

PR #259, opened 2022-06-28
fix: form text formatting

marcellmueller:
This fixes some weird text formatting issues due to my use of `white-space: pre line` so we could have multi line titles as required in some form pages. This just removes the line breaks and turns template literals into regular strings. 

Also changed instances of the term `Applicant` to lowercase and removed oxford commas as these aren't used in government/

Screenshot at 2022-06-28 13-23-14

### DIFF
--- a/.github/actions/app/action.yaml
+++ b/.github/actions/app/action.yaml
@@ -29,7 +29,7 @@ inputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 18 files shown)


AntBush (approved):
LGTM, sad about the loss of the oxford comma

PR #289, opened 2022-07-04
263 required fields

marcellmueller:


app/styles/GlobalTheme.tsx
```diff
@@ -15,6 +15,7 @@ const theme = {
     backgroundGrey: '#F2F2F2',
     components: '#606060',
     error: '#D8292F',
+    errorBackground: 'rgba(248, 214, 203, 0.4)',
```
VB-at-Bis, line 18: should we be consistent and use #f8d6cb66 ?

VB-at-Bis (commented):
Looks good

PR #302, opened 2022-07-12
fix: add labels to PostgresCluster and update metabase network policy

matthieu-foucault:


helm/app/templates/networkPolicies/metabase-ingress.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
 spec:
   podSelector:
     matchLabels:
-      postgres-operator.crunchydata.com/instance-set: {{ .Release.Name }}-hippo-ha
+      app.kubernetes.io/instance: {{ .Release.Name }}
   ingress:
```
marcellmueller, line 11: https://github.com/bcgov/CONN-CCBC-portal/commit/9aa2f9c692c8c262d8aaf751d667ccfc056adbcc

I guess we just needed to remove the `-hippo-ha`? Should we try a test deployment to dev before we merge this in?

matthieu-foucault: > I guess we just needed to remove the `-hippo-ha` ?

We also needed to add labels to the PostgresCluster, which were missing

> Should we try a test deployment to dev before we merge this in?

Sure, that would be better than making another PR if it doesn't work

PR #312, opened 2022-07-15
214 wording updates

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/214

Added a new folder called `CustomTitles` to render more complex field titles in the design that had things like lists. They are imported and rendered in the `FieldTemplate` component. In most cases `label: false` should be passing into `ui:options` in `uiSchema` so that the title/field key isn't rendered as well.


VB-at-Bis (approved):
Looks good

PR #316, opened 2022-07-18
291 inputs and masks

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/291

### DIFF
--- a/app/components/Form/Label.tsx
+++ b/app/components/Form/Label.tsx
@@ -1,4 +1,5 @@
 import styled from 'styled-components';
+
 const StyledLabel = styled('label')`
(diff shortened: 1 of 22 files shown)


marcellmueller:
> Looks good. Couple of comments (non-blocking):
> 
> * should we move colors (i.e. #e71f1f) into shared/global constants?
> * should we have tests for error values in TextInput widget?

Yeah I think moving colours into a constants file is a great idea to do in the future since Jest doesn't play well with the global themes but I will save that for another PR.

Good idea on the tests, I will add a test for error values before I merge this thank you.

Issue #291, opened 2022-07-05
Inputs and masks 👺 🎭


elliottSlee:
### User Story
As an applicant
I want the input fields to detect whether the information I enter is formatted correctly and give me guidance on how to input the correct information
so that I don't make preventable errors.

### Description
In usability testing, all users expected dollar, phone, and postal code inputs to give guidance on how to respond and, critically, to prevent them from inputting incorrect information.

Errors should be formatted by the input box changing to a 2px red box (#E71F1F) and red warning message (#E71F1F) per the [BC Design System](https://developer.gov.bc.ca/Forms).

### Acceptance Criteria:
<details>
  <summary><strong>Budget details page</strong></summary>

<br>

**Total eligible costs & Total project cost**
- [x] Only accepts numbers
- [x] Mask $###,###,###.##

</details>
<br/>

<details>
  <summary><strong>Project funding page</strong></summary>

<br>

**All input fields**
- [x] Only accepts numbers
- [x] Mask $###,###,###.##

</details>
<br/>

<details>
  <summary><strong>Other funding sources page</strong></summary>

<br>

**Amount requested under funding partner** (input fields for the next 5 years and total)
- [x] Only accepts numbers
- [x] Mask $###,###,###.##

</details>
<br/>

<details>
  <summary><strong>Benefits</strong></summary>

<br>

**Final number of Eligible Households targeted by this proposal**
- [x] Only accepts a whole number

</details>
<br/>

<details>
  <summary><strong>Estimated project employment</strong></summary>

<br>

**Current employment**
- [x] Only accepts a whole number
- [x] >= 1M
- [x] Mask ###,###

**Hours of employment per week** 
- [x] Only accepts numbers, up to 1 decimal place

**Average months per year of full time employment that will be created**
- [x] Only accepts numbers, up to 1 decimal place

</details>
<br/>

<details>
  <summary><strong>Organization Contact Information</strong></summary>

<br>

**Telephone**
- [x] Only accepts numbers
- [x] Input mask telephone as "###-###-####" e.g., 555-555-5555
#
**Extension**
- [x] Only accepts numbers
- [x] maxlength: 9
</details>
<br/>

<details>
  <summary><strong>Authorized & alternate contact pages</strong></summary>

<br>

**Telephone**
- [x] Only accepts numbers
- [x] Input mask telephone as "###-###-####" e.g., 555-555-5555
#
**Extension**
- [x] Validation: number
- [x] maxlength: 9
</details>

[...]

PR #320, opened 2022-07-19
260 add indigenous field

marcellmueller:
This is branched from #291 and is a small PR to add a new field to the benefits page.

https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/260


VB-at-Bis (approved):
Looks good

PR #326, opened 2022-07-20
Aws assume role

marcellmueller:
We had to update our`s3client` to use the AWS `RoleArn` credentials in our configuration.


VB-at-Bis (approved):
Perfect

PR #332, opened 2022-07-21
Terraform script to provision s3 bucket

VB-at-Bis:
Updated Terraform script to provision private s3 bucket


marcellmueller (approved):
Looks great, thanks again for this one!

PR #338, opened 2022-07-22
Fix uploads

marcellmueller:
Fix for file widget using proper form id and fix for review page upload fields file names not rendering.


VB-at-Bis (approved):
Looks good

PR #340, opened 2022-07-22
chore: fix relay ssr setup

marcellmueller:
Our relay setup wasn't quite properly implemented plus using a mix of `usePreloadedQuery` and `useLazyLoadedQuery` was causing issues. This fixes most of the SSR issues, I just need to fix the queries on the `[page]` and `/dashboard` pages. They are currently partially implemented though I need to figure out how to pass in query variables.


VB-at-Bis (approved):
Looks good. I rebuild app and tested locally.

PR #346, opened 2022-07-25
274 delete uploaded file

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/274

Might want to actually test this one whoever is reviewing this. I made it so once the column `is_deleted` is set to true the `deleted_at` and `deleted_by` gets updated using the `update_timestamps` trigger function. Then there is some logic on the front end to remove it from the file from our RJSF field file list.


VB-at-Bis (approved):
Looks good.
Cancel image can be better :)

PR #349, opened 2022-07-25
feat: add ignoreoptional option

marcellmueller:
Updated the postal code example formatting and added a new `ui:options` attribute `ignoreOptional` which removes the `(optional)` to the end of non required fields if desired.


VB-at-Bis (approved):
Looks good

PR #353, opened 2022-07-26
detect and persist identity provider

VB-at-Bis:
Discussed ticket 169 with Junmin. He suggested to use `identity_provider` and `preferredUsername` as main identifiers, and do not store idir, bceid_basic or bceid_business because Keycloak is using `preferredUsername` as `keycloak_username`, and it is unique user identifier.
Code changes:
- added `identity_provider` in keycloak_jwt type
- updated code to populate `identity_provider` from JWT
- updated revert action for `applications` table that was blocking `sqitch revert` and `sqitch deploy`

### DIFF
--- a/db/revert/tables/applications/remove_unique_owner.sql
+++ b/db/revert/tables/applications/remove_unique_owner.sql
@@ -2,6 +2,7 @@
 
 BEGIN;
 
-ALTER TABLE ccbc_public.applications ADD CONSTRAINT applications_owner_key UNIQUE(owner);
+-- commented out becuase it causes errors on `sqitch revert`
+-- ALTER TABLE ccbc_public.applications ADD CONSTRAINT applications_owner_key UNIQUE(owner);
 
 COMMIT;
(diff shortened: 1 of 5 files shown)


db/revert/tables/applications/remove_unique_owner.sql
```diff
@@ -2,6 +2,7 @@
 
 BEGIN;
 
-ALTER TABLE ccbc_public.applications ADD CONSTRAINT applications_owner_key UNIQUE(owner);
+-- commented out becuase it causes errors on `sqitch revert`
+-- ALTER TABLE ccbc_public.applications ADD CONSTRAINT applications_owner_key UNIQUE(owner);
```
AntBush, line 6: Do we want to keep this commented out? If we keep this commented out, when someone attempts to revert before this they won't be in the same state as before this was deployed. 

I'm not sure what the proper solution for this should be. Can @wenzowski provide some suggestions? The problem is when the db is populated, this causes the revert to fail.

VB-at-Bis: I would suggest to keep it commented out. It does not harm existing or new users as it result in `ccbc_public.applications` without the constraint - which is the requirement to allow multiple applications for the same user.

AntBush: Agreed, I'll go ahead and approve this then

AntBush (approved):

PR #354, opened 2022-07-26
Fix migration

marcellmueller:
Added some new columns to `attachment` table and sqitch didn't know about them. Instead of rebasing I am trying a way that Dylan suggested that will be good if we have to add columns when in prod. Basically make a numbered file in sqitch `attachment_001` `with alter table add column` and add to sqitch plan.

I tested pushing this to dev so it works, will need to merge in so we can deploy db to dev again (sorry I didn't know a better way).


VB-at-Bis (approved):
Looks good

PR #355, opened 2022-07-27
92 mark as complete

AntBush:
### Description: 

Adds CCBC_ID to the DB and hooks the function to the application

References # 92

### Type of change:

- [X] New feature (non-breaking change which adds functionality)


AntBush:
This is currently blocked due to some relay stuff. 

When working on this, there are 2 considerations that should be noted the first: the function to add a reference number to the application is not using the lock I set in place due to a permission issue. I think the best course of action would be to grant lock permission to the ccbc_auth_user to resolve this as otherwise, we can run into race conditions. Though there may be other security considerations surrounding this. 

The second consideration: when deploying these changes the intake table is empty, but adding an intake relies on an existing row in the intake table (one that has both open_timestamp and close_timestamp filled with a valid time that is between the current time). You should be sure to add in an intake time, maybe even a bogus intake that is valid from CURRENT_TIMESTAMP -> the day before release.

VB-at-Bis:
- back merged `main`;
- added fallback for missing record in `intake` table;
- fixed CCBC Id column in Application list;

marcellmueller (approved):
Looks good :+1:

PR #368, opened 2022-08-02
28 session expiration handling

VB-at-Bis:
- added SessionExpirtHandler (cloned for cas-cif);
- added tests
- back-merged main


marcellmueller:
@VB-at-Bis This looks good, I tried it and the modal is working. The e2e tests seem to be working, I tried running locally and rerunning here and they are fine unless I'm missing something.

marcellmueller (approved):

PR #379, opened 2022-08-04
39-application-withdraw

marcellmueller:


app/pages/form/[id]/[page].tsx
```diff
@@ -56,13 +62,19 @@ const FormPage = ({ preloadedQuery }: RelayProps<{}, PageQuery>) => {
   return (
     <Layout session={session} title="Connecting Communities BC">
       <FormDiv>
+        {status === 'withdrawn' && (
```
Comment (VB-at-Bis) [inline app/pages/form/[id]/[page].tsx:65]: .. I wish we did not use string comparison...

app/components/Dashboard/Table.tsx
```diff
@@ -69,60 +79,75 @@ const Table = ({ applications }: Props) => {
...
+            const intakeClosingDate = intakeByIntakeId?.closeTimestamp;
+            const isIntakeClosed = intakeClosingDate
+              ? Date.parse(intakeClosingDate) < Date.now()
+              : false;
+
+            const isWithdrawn = application.status === 'withdrawn';
+
+            const getApplicationUrl = (status: string) => {
+              if (status === 'withdrawn') {
+                return `/form/${application.rowId}/19`;
```
VB-at-Bis, line 114: should we move `19` into a const? what if order of pages changes again?

VB-at-Bis (approved):
Looks good; added some minor comments

app/components/Dashboard/Table.tsx
```diff
@@ -62,8 +62,12 @@ const Table = ({ applications }: Props) => {
   const applicationNodes = applications.allApplications.nodes;
   const router = useRouter();
 
+  const formPages = Object.keys(schema({}).properties);
+
+  const reviewPage = formPages.indexOf('review') + 1;
```
marcellmueller, line 67: @VB-at-Bis Ok I am getting the review page here now instead of using number or const.

PR #385, opened 2022-08-08
262 automate calculations

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/262

### DIFF
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -1,7 +1,6 @@
 import { useState } from 'react';
 import { useRouter } from 'next/router';
 import Button from '@button-inc/bcgov-theme/Button';
-import Alert from '@button-inc/bcgov-theme/Alert';
 import type { JSONSchema7 } from 'json-schema';
 import { CalculationForm, FormBase } from '.';
 import uiSchema from '../../formSchema/uiSchema';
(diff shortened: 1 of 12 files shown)


app/components/Form/ApplicationForm.tsx
```diff
@@ -177,12 +179,52 @@ const ApplicationForm: React.FC<Props> = ({
...
-      <Button variant="secondary" style={{ marginLeft: '20px' }}>
-        Save
-      </Button> */}
-    </FormBase>
+    </>
+  );
+
+  return (
+    <>
+      {isCalculatedPage ? (
```
marcellmueller, line 187: @VB-at-Bis @AntBush what do you guys think about this? I wasn't sure if there is a cleaner way to go about using the `CalculatedForm` renderer. This way we only use it on the pages that use the calculations but it feels a bit messy somehow.

VB-at-Bis: It looks good for me :) .. not messy at all

AntBush: I think this way for one or two it's fine, but we'll definitely have to think of a different way to do this if we have many custom forms we want to display.

AntBush (approved):
Looks good to me!

app/components/Form/ApplicationForm.tsx
matthieu-foucault: Did you check how calculated fields are handled in CIIP and CIF. Adding some section-specific branching in the `ApplicationForm` page is not a maintainable solution. The rjsf way of doing this is to define custom fields for those items that require calculations

PR #387, opened 2022-08-09
Applications Exposed when logged off

AntBush:
The crux of this issue is that the `applications` table was given guest select permissions, where that is not wanted/needed in this case. This PR is to rectify that. This PR still requires tests - and as such is not ready to be merged in


matthieu-foucault:
@AntBush https://github.com/bcgov/cas-cif/blob/develop/schema/test/unit/tables/project_test.sql has some permissions test examples.

Something like this would work. You'll want to add tests for insert, update and delete as well, and for the `ccbc_auth_user` role.
`delete` should be forbidden for all roles, as we only want to allow soft-delete.
```sql
set role ccbc_guest;
select throws_like(
  $$
    select * from ccbc_public.applications
  $$,
  'permission denied%',
  'ccbc_guest cannot select rows from table applications'
);
```

VB-at-Bis:
Dashboard link will not be visible for non-logged in users with  https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/294 . Direct access to /dashboard URL as a bookmark or browser link redirects user to Home page. Therefore, there are no steps in UI that can lead to the issue of `Applications Exposed when logged off`

matthieu-foucault:
Although the UI doesn't expose applications for the `ccbc_guest` we still want to ensure that database permissions are correct and tested, so this change is still needed

VB-at-Bis:
Tests added and verified locally:
```vb@vbpro:/mnt/c/Workspace/Button/CONN-CCBC-portal/db/test/unit/tables$ pg_prove -d ccbc ccbc_user_test.sql
ccbc_user_test.sql .. ok
All tests successful.
Files=1, Tests=21,  0 wallclock secs ( 0.03 usr  0.00 sys +  0.05 cusr  0.01 csys =  0.09 CPU)
Result: PASS
```

db/test/unit/tables/ccbc_user_test.sql
```diff
@@ -116,5 +116,13 @@ select throws_like(
...
 );
 
+select throws_like(
+  $$
+    select * from ccbc_public.applications
+  $$,
+  'permission denied%',
+    'ccbc_guest cannot select rows from table_applications'
+);
+
```
matthieu-foucault, line 126: can we move this to a new `application_test.sql` file? 
We will want to test the other permissions as well (insert, update, delete)

matthieu-foucault (dismissed):

db/test/unit/tables/ccbc_user_test.sql
VB-at-Bis: Moved tests to `ccbc_guest_test.sql`

db/test/unit/tables/ccbc_guest_test.sql
```diff
@@ -0,0 +1,41 @@
+begin;
```
marcellmueller, line 1: I believe this should all stay in the same file. This test is in the `/tables` folder which was named after `ccbc_user` table.

It is modelled after this one but ours needs work: 
https://github.com/bcgov/cas-cif/blob/develop/schema/test/unit/tables/cif_user_test.sql

This morning in our meeting Alec asked us to rename ours for consistency here:
https://github.com/bcgov/CONN-CCBC-portal/pull/429/files

Though ours is a work in progress.

marcellmueller (approved):

db/test/unit/tables/ccbc_guest_test.sql
VB-at-Bis: Ok, lets merge this as is and re-factor tests in another ticket :)

marcellmueller (approved):

PR #389, opened 2022-08-09
Dashboard fixes and testing

marcellmueller:
This PR adds Relay testing helpers from cif and unit tests for the dashboard. It also fixes to the dashboard withdraw button z-index and also minor fixes to the edit/view button and url handling.


app/components/Dashboard/Table.tsx
```diff
@@ -110,17 +111,30 @@ const Table = ({ applications }: Props) => {
...
             };
 
+            const formatEditButton = () => {
+              if (isWithdrawn) {
+                return 'View';
+              } else if (isSubmitted && isIntakeClosed) {
+                return 'View';
+              } else {
+                return 'Edit';
+              }
```
wenzowski, line 135: We've got this logic duplicated a bit. Can we consider an `isEditable()` helper to refactor out the duplicated chained elif?

app/tests/components/Dashboard/Dashboard.test.tsx
```diff
@@ -0,0 +1,248 @@
...
+      },
+    };
+    componentTestingHelper.loadQuery(payload);
+    componentTestingHelper.renderComponent();
+    expect(screen.getByText('CCBC-010005')).toBeInTheDocument();
+    expect(screen.getByText('withdrawn')).toBeInTheDocument();
+    expect(screen.getByText('View')).toBeInTheDocument();
+    expect(screen.queryByTestId('withdraw-btn-test')).toBeNull();
+  });
+});
```
wenzowski, line 248: these are all valuable test cases; what's missing now is db unit tests to confirm the expected permissions for mutability and access are respected. These tests confirm that the user can see whether something should be editable or view only but not whether those applications are actually edit-protected or view-protected. The additional tests should be done at pgtap level.

wenzowski (changes requested):
One small opportunity for reafactor but still missing pgtap validation of api permissions.

VB-at-Bis (approved):
Looks good

PR #392, opened 2022-08-10
Remove attachment columns

marcellmueller:
Remove columns `is_deleted`, `deleted_by`,and  `deleted_at` and use `archived_by` and `archived_at` instead.

Update for:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/274


marcellmueller:
@matthieu-foucault I'm not sure the proper way to go about this. I removed the `attachment_001` files from sqtich as well as removed the line in `sqitch.plan`. It works fine for a fresh `sqitch deploy` but is broken on an already deployed database.

matthieu-foucault:
> @matthieu-foucault I'm not sure the proper way to go about this. I removed the `attachment_001` files from sqtich as well as removed the line in `sqitch.plan`. It works fine for a fresh `sqitch deploy` but is broken on an already deployed database.

This sounds right. Our deployments in dev, and in test until we cut a release, should wipe the database before deploying the sqitch changes. Once we have our first release, and our first tag in the sqitch plan, we'll start doing incremental changes, but until then we want to update existing changes to avoid adding unnecessary complexity.

app/lib/theme/widgets/FileWidget.tsx
```diff
@@ -178,7 +178,7 @@ const FileWidget: React.FC<WidgetProps> = ({
     const variables = {
       input: {
         attachmentPatch: {
-          isDeleted: true,
+          archivedAt: new Date().toISOString(),
```
VB-at-Bis, line 181: toISOString The timezone is always zero UTC offset. Are we string dates in UTC? 
... just want to make sure that we are nor required to use PST(PDT)

VB-at-Bis (approved):
Looks good

PR #409, opened 2022-08-11
expiryHandler settings

VB-at-Bis:
- updated `enabled` flag;
- changed redirect URL to home page;

### DIFF
--- a/app/components/SessionExpiryHandler.tsx
+++ b/app/components/SessionExpiryHandler.tsx
@@ -45,7 +45,7 @@ const SessionExpiryHandler: React.FC = () => {
 
     if (!logOutURL) {
       router.push({
-        pathname: '/login-redirect',
+        pathname: '/',
         query: {
           redirectTo: router.asPath,
           sessionIdled: true,
@@ -64,7 +64,7 @@ const SessionExpiryHandler: React.FC = () => {
           modalDisplaySecondsBeforeLogout={120}
           onSessionExpired={handleSessionExpired}
           extendSessionOnEvents={{
-            enabled: false,
+            enabled: true,
             throttleTime: 300000,
             events: ['keydown', 'scroll'],
           }}


app/components/SessionExpiryHandler.tsx
```diff
@@ -45,7 +45,7 @@ const SessionExpiryHandler: React.FC = () => {
 
     if (!logOutURL) {
       router.push({
-        pathname: '/login-redirect',
+        pathname: '/',
```
matthieu-foucault, line 48: If we want the "After re-login, you are redirected to the previous page" acceptance criteria to be met, we need that `login-redirect` page, or, at least, the login button and the sso middleware need to handle the `redirectTo` query parameter.

Looking at CIF, you can see that the [`LoginForm`](https://github.com/bcgov/cas-cif/blob/4fc8f2bd987416b0c82601625a4a25595bb42d18/app/components/Session/LoginForm.tsx) passes the `redirectTo` param to the `/login` endpoint, and that it is handled by the [`ssoMiddleware`](https://github.com/bcgov/cas-cif/blob/752cad1410765cf9b9693382f9b48757fb65ef59/app/server/middleware/sso.ts#L23) to determine the landing route.

Our pages there also redirect users to the `login-redirect` page, [setting the `redirectTo` param](https://github.com/bcgov/cas-cif/blob/78605b8ce6df8b21a02e9ddf8d4ac05dd857ed54/app/lib/relay/withRelayOptions.tsx#L37), when they are not logged in  (using the next.js [`getServerSideProps` redirect response](https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#redirect))


[Searching for `redirectTo`](https://github.com/bcgov/cas-cif/search?q=redirectTo) in the cif repo should point you to all the necessary code and tests.

matthieu-foucault (changes requested):

app/components/SessionExpiryHandler.tsx
matthieu-foucault: As per @MarsRomer 's comments this change is not needed, and redirecting users to the previous page after re-login is not a feature we need now

matthieu-foucault (approved):

PR #429, opened 2022-08-15
increase pgtap coverage

marcellmueller:
- pgtap tests
- removed an `it.only` and improved some jest assertion for mutation assertions. Unrelated to pgtap, but I'm adding this as a bonus techdebt item :)


app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -141,7 +141,7 @@ describe('The application form', () => {
     });
   });
 
-  it.only('displays the correct button label for withdrawn applications', async () => {
```
marcellmueller, line 144: :sweat_smile:

matthieu-foucault (approved):

PR #437, opened 2022-08-16
Fix/withdraw

marcellmueller:
This PR fixes the withdraw not working which I think got reverted during a rebase. I also fixed what I assume is a separate rebase issue that snuck in our `app.tsx` file.

### DIFF
--- a/app/components/Dashboard/Table.tsx
+++ b/app/components/Dashboard/Table.tsx
@@ -58,7 +58,7 @@ type Props = {
 };
 
 const Table = ({ applications }: Props) => {
-  const [withdrawRowId, setWithdrawRowId] = useState(0);
+  const [withdrawId, setWithdrawId] = useState('');
 
   const applicationNodes = applications.allApplications.nodes;
   const router = useRouter();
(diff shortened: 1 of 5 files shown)


app/components/Dashboard/Table.tsx
```diff
@@ -156,7 +156,7 @@ const Table = ({ applications }: Props) => {
         </tbody>
       </StyledTable>
 
-      <Modal id={withdrawRowId} />
+      <Modal id={withdrawId} />
```
matthieu-foucault, line 159: Adding some integration tests for our dashboard page, ensuring that we can withdraw an application (i.e. call the updateApplicationMutation with the expected input), would ensure what this bug isn't reintroduced. Happy to pair with you to guide you in writing that test if needed

matthieu-foucault (changes requested):

app/tests/components/Dashboard/Dashboard.test.tsx
```diff
@@ -172,12 +173,67 @@ describe('The Dashboard', () => {
...
+      },
+    };
+    componentTestingHelper.loadQuery(payload);
+    const user = userEvent.setup();
+
+    componentTestingHelper.renderComponent();
+    expect(screen.getByText('CCBC-010005')).toBeInTheDocument();
+    expect(screen.getByText('submitted')).toBeInTheDocument();
+    expect(screen.getByText('Edit')).toBeInTheDocument();
+    expect(screen.getByTestId('withdraw-btn-test')).toBeInTheDocument();
```
matthieu-foucault, line 216: using test ids should be a [last resort](https://testing-library.com/docs/queries/bytestid), we should use [`getByRole`](https://testing-library.com/docs/queries/byrole) instead

matthieu-foucault (approved):

PR #440, opened 2022-08-17
fix: filewidget delete empty array validation bug

marcellmueller:
Fixes a bug where if a user deletes a file it would still store an empty array in the field, causing required fields to incorrectly validate as filled. This is what caused the missing required field Mars flagged on the review page in this ticket: 

https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/277

### DIFF
--- a/app/components/Review/Review.tsx
+++ b/app/components/Review/Review.tsx
@@ -45,21 +45,6 @@ const StyledAlert = styled(Alert)`
   margin-bottom: 32px;
 `;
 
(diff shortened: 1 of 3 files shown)


VB-at-Bis (approved):
Looks good

AntBush (approved):
Looks good to me!

PR #452, opened 2022-08-18
fix: 413 payload too large

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/444

This should fix this error, though we should still likely set more max lengths on input fields because I suspect he is pasting in massive amounts of text.

### DIFF
--- a/app/server.ts
+++ b/app/server.ts
@@ -19,7 +19,7 @@ const app = next({ dev });
 const handle = app.getRequestHandler();
 
 const { json, urlencoded } = bodyParser;
-
+const bodyParserLimit = '50mb';
 app.prepare().then(async () => {
   const server = express();
 
@@ -32,9 +32,9 @@ app.prepare().then(async () => {
     await pgPool.end();
   });
 
-  server.use(json());
+  server.use(json({ limit: bodyParserLimit }));
 
-  server.use(urlencoded({ extended: false }));
+  server.use(urlencoded({ extended: false, limit: bodyParserLimit }));
 
   server.use(cookieParser());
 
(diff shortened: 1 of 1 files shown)


app/server.ts
```diff
@@ -19,7 +19,7 @@ const app = next({ dev });
 const handle = app.getRequestHandler();
 
 const { json, urlencoded } = bodyParser;
-
+const bodyParserLimit = '50mb';
```
matthieu-foucault, line 22: where does 50mb come from? 
Can we calculate the maximum expected payload size, assuming 32k characters for each of those inputs that don't have a max length?

marcellmueller: 50mb comes from various examples around the web as well as it is the limit that is set in cif.

https://github.com/bcgov/cas-cif/blob/develop/app/server/index.ts

matthieu-foucault: Given our limited storage sizes, we might want to lower that to prevent users from filling the database by forging a 50Mb JSON. 5Mb might be more than enough?

@wenzowski, any recommendation here?

wenzowski: What error do we get when we exceed the `bodyParserLimit`? How can we monitor for users hitting the parse limit in the normal course of system operation?

AntBush: We get a 413 error returned back from the server.

matthieu-foucault: @marcellmueller, can you change this to 5Mb? We should be able to monitor for those errors with sysdig

matthieu-foucault (approved):

PR #453, opened 2022-08-18
421 - error saving

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/421

Adds max lengths to all fields in authorized and alternate contact pages, as well as a max length of 32000 for any text fields that don't have one set.


VB-at-Bis (approved):
Looks good

PR #455, opened 2022-08-18
296 - migrate keycloak

marcellmueller:
Started implementing this via the JSON Alec added to 1pass though currently I'm blocked by my lack of IDIR.


matthieu-foucault (changes requested):

marcellmueller:
@matthieu-foucault this is ready for re-review.

matthieu-foucault (approved):
@marcellmueller , can you update 1 password to have the new client secret for the dev env?

The logout issue was an actual issue with the redirect URIs configuration. I have updated the CSS app to fix that

PR #465, opened 2022-08-22
fix: remove optional on read-only widgets under estimatedPorjectEmployment

AntBush:
Removes optional as per issue 464


marcellmueller (approved):
Looks good. In hindsight I wish I named it `hideOptional` though :sweat_smile:

PR #479, opened 2022-08-23
Create submitApplication mutation

matthieu-foucault:
This changes the handling of application submission and generation of the ccbc number.

When intake records are created we use a trigger to generate a sequence corresponding to the intake.
This sequence is then used to generate the ccbc number upon submission of an application.
We use a single `submitApplication` mutation to resolve the current issue of needing two mutations to submit an application


AntBush (approved):
Looks good to me, and confirmed working!

PR #481, opened 2022-08-23
View withdrawn application continue button fix

marcellmueller:
This PR changes the `Save and continue` button text to `Continue` if the application status is withdrawn as well as fixes the page routing that was refactored to happen upon completion of the update application mutation. The button text will remain `Continue` even on the final submission page so there isn't any confusion.

### DIFF
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -295,25 +295,19 @@ const ApplicationForm: React.FC<Props> = ({
   };
 
   const handleDisabled = (page: string, noErrors: boolean) => {
-    let disabled = false;
     switch (true) {
       case isWithdrawn:
-        disabled = false;
-        break;
+        return false;
       case page === 'review' && noErrors:
-        disabled = false;
-        break;
+        return false;
       case page === 'review':
-        disabled = !reviewConfirm;
-        break;
+        return !reviewConfirm;
       case page === 'acknowledgements':
-        disabled = !areAllAcknowledgementsChecked;
-        break;
+        return !areAllAcknowledgementsChecked;
       case page === 'submission':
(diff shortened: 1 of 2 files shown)


app/components/Form/ApplicationForm.tsx
```diff
@@ -292,6 +298,9 @@ const ApplicationForm: React.FC<Props> = ({
   const handleDisabled = (page: string, noErrors: boolean) => {
     let disabled = false;
     switch (true) {
+      case isWithdrawn:
+        disabled = false;
```
AntBush, line 302: Sonarcloud flagged this as redundant, we can remove the disabled assignment here and just break.

AntBush (changes requested):
Just the minor change flagged by sonarcloud

AntBush:
Additionally, do we want to still go to the success page on a withdrawn application? Maybe change that last `Continue` to go back to dashboard or something? I know that's mostly up to the designers though.

app/components/Form/ApplicationForm.tsx
marcellmueller: Alright I refactored this and removed the let altogether, sonar looks happy now.

AntBush:
> Though I think that is out of scope for this ticket.

Absolutely, just thought it was worth bringing up while it was being worked on.

AntBush (approved):
LGTM

marcellmueller:
> > Though I think that is out of scope for this ticket.
> 
> Absolutely, just thought it was worth bringing up while it was being worked on.



> > Though I think that is out of scope for this ticket.
> 
> Absolutely, just thought it was worth bringing up while it was being worked on.

For sure, we can bring it up in sprint planning or something.

PR #491, opened 2022-08-24
Sonarcloud CI

AntBush:
Sonarcloud CI


matthieu-foucault (changes requested):

.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - name: setup dev env
+      uses: ./.github/actions/dev-env-setup
+    - name: Run Jest tests
+      working-directory: ./app
+      run: yarn test --coverage
+    - name: SonarCloud Scan
+      uses: sonarsource/sonarcloud-github-action@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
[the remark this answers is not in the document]
AntBush: Fixed. Is the pre-commit hook active now? I thought it wasn't working

.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,27 @@
+on:
+  # Trigger analysis when pushing in master or pull requests, and when creating
+  # a pull request.
+  push:
+    branches:
+      - master
```
matthieu-foucault, line 6:
```suggestion
      - main
```

.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        # Disabling shallow clone is recommended for improving relevancy of reporting
+        fetch-depth: 0
+    - name: setup dev env
+      uses: ./.github/actions/dev-env-setup
+    - name: Run Jest tests
+      working-directory: ./app
+      run: yarn test --coverage
```
matthieu-foucault, line 22: We should update the `jest` job in the tests workflow, rather than adding another workflow that runs `yarn test` ?

.github/workflows/sonarcloud.yaml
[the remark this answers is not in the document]
matthieu-foucault: It's not, but it should :)

AntBush:
> @AntBush I see we don't have coverage or duplication information according to the sonarcloud bot, not sure this is working

Since none of the .ts/.js files were edited, there is no change in coverage information. You'll see in the overview it'll mention the expected amount after merging.

image

.github/workflows/sonarcloud.yaml
AntBush: Fixed

.github/workflows/sonarcloud.yaml
AntBush: I'm probably going about this the wrong way, but when attempting to reuse the jests workflow the coverage data is wiped on the new checkout, and it isn't persisted between jobs. Any guidance on this?

AntBush: Changed the jest tests to an action and reusing that, let me know if that's sufficient

.github/workflows/tests.yaml
```diff
@@ -22,11 +22,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: dev env setup
         uses: ./.github/actions/dev-env-setup
-      - name: run app locally
-        uses: ./.github/actions/local-app-run
       - name: Run Jest tests
-        working-directory: ./app
-        run: yarn test
+        uses: ./.github/actions/jest-test
```
matthieu-foucault, line 26: We're running jest tests twice in our CI. Can't we just have the sonarcloud step here?

.github/workflows/tests.yaml
```diff
@@ -22,11 +22,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: dev env setup
         uses: ./.github/actions/dev-env-setup
-      - name: run app locally
-        uses: ./.github/actions/local-app-run
```
matthieu-foucault, line 26: If you want to try removing this from the jobs below too, I don't think it's needed

AntBush: The local-app-run deploys the DB, so it's needed for the schema and pgtap tests. It should be refactored, but if we can leave it for another ticket since in the meantime sonarcloud isn't being run on other PRs

.github/workflows/tests.yaml
AntBush: Good point, yeah I can go ahead and move the sonarcloud run in the tests workflow

matthieu-foucault (approved):

PR #496, opened 2022-08-25
chore: add xlx file types to fields with xlsx

marcellmueller:


app/formSchema/uiSchema.ts
```diff
@@ -603,7 +603,7 @@ const uiSchema = {
     'ui:options': {
       maxLength: MAX_LONG_INPUT_LENGTH,
       label: false,
-      fileTypes: '.xlsx',
+      fileTypes: 'xls, .xlsx',
```
matthieu-foucault, line 606:
```suggestion
      fileTypes: '.xls, .xlsx',
```
Should there be a period in front of each of those? 
Additionally, can we create an `EXCEL_FILE_EXTENSIONS` constant in this file to ensure consistency?

matthieu-foucault (changes requested):

app/formSchema/uiSchema.ts
```diff
@@ -549,7 +551,7 @@ const uiSchema = {
     'ui:options': {
       allowMultipleFiles: true,
       maxLength: MAX_LONG_INPUT_LENGTH,
-      fileTypes: '.xlsx, .mpp',
+      fileTypes: '.xls, .xlsx, .mpp',
```
matthieu-foucault, line 554:
```suggestion
      fileTypes: `${EXCEL_FILE_EXTENSIONS}, .mpp`,
```

app/formSchema/uiSchema.ts
```diff
@@ -575,7 +577,7 @@ const uiSchema = {
     'ui:options': {
       maxLength: MAX_LONG_INPUT_LENGTH,
       fileTypes:
-        '.eml, .msg, .txt, .pdf, .doc, .docx, .xlsx, .xlsx, .xml, .jpg, .jpeg, .png',
+        '.eml, .msg, .txt, .pdf, .doc, .docx, .xls, .xlsx, .xml, .jpg, .jpeg, .png',
```
matthieu-foucault, line 580:
```suggestion
        `.eml, .msg, .txt, .pdf, .doc, .docx, .xml, .jpg, .jpeg, .png, ${EXCEL_FILE_EXTENSIONS}`,
```

matthieu-foucault (changes requested):

app/formSchema/uiSchema.ts
marcellmueller: I dunno I considered that but it seems messy and less readable.

matthieu-foucault: It's a bit more verbose but it ensures that you're consistent. Ideally we'd have all of these extensions in named constants and we'd just join all of them. Some of the extensions listed in this file are not necessarily well known (e.g. `.mpp`) and should really be in a constant.

matthieu-foucault (approved):

PR #513, opened 2022-08-26
Chore/438-rename-applications

AntBush:
Renames applications to application

### DIFF
--- a/db/test/unit/tables/applications_test.sql
+++ b/db/test/unit/tables/applications_test.sql
@@ -8,7 +8,7 @@ SELECT * FROM no_plan();
 -- Table exists
 select has_table(
   'ccbc_public', 'applications',
-  'ccbc_public.applications should exist and be a table'
+  'ccbc_public.application should exist and be a table'
 );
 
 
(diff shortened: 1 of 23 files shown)


matthieu-foucault (changes requested):
looks like there's more `ccbc_public.applications` left in the code

db/test/unit/tables/applications_test.sql
```diff
@@ -8,7 +8,7 @@ SELECT * FROM no_plan();
 -- Table exists
 select has_table(
   'ccbc_public', 'applications',
-  'ccbc_public.applications should exist and be a table'
+  'ccbc_public.application should exist and be a table'
 );
```
matthieu-foucault, line 14: The columns tests below need to be updated too

matthieu-foucault (changes requested):

db/test/unit/tables/applications_test.sql
AntBush: 🤦 Thanks

matthieu-foucault (approved):

Issue #500, opened 2022-08-25
OWASP scan


MarsRomer:

eazeredo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/500?planning-poker) with Zenhub @AntBush @marcellmueller @matthieu-foucault @wenzowski

wenzowski:
https://github.com/bcgov/pipeline-templates/blob/main/docs/github.md#owasp-scan

wenzowski:
https://github.com/bcgov/cas-cif/blob/ccb9ee4a4c622137ded61ca33625128a564da543/.github/workflows/test.yml#L173

PR #519, opened 2022-08-26
feat: add zap scan

wenzowski:
This PR turns on both baseline and full scans, and throws failing status checks for any new security scan violations found. Unfortunately, 19 rule violations had to be squelched, and each will need to be individually reviewed and either remediated or accepted.

for #500


marcellmueller (approved):
Looks good to me!

matthieu-foucault (approved):

PR #522, opened 2022-08-26
feat: trivy container scanning

wenzowski:
#366

Issue #366, opened 2022-08-02
Container Scanning


MarsRomer:
Discuss with Nick Corcoran about container scanning.  
This will scan the OS level code that we have imported.

MarsRomer:
Discuss with Nick Corcoran about container scanning

MarsRomer:
Blocked until I talk to Nick about this

MarsRomer:
@wenzowski  container scanning results = Redhat ACS right?

MarsRomer:
3 things were decided from meeting with Nick:
1. Dev team should monitor the ACS dashboard 
Action:  Mars to facilitate getting access for all developers on our team
2. Scrum team responsible for managing the vulnerabilities and violations 
Action: Mars to set aside a portion of the sprint devoted to addressing the highest priority vulnerabilities and violations.   
Action:  Team to agree on how we are going to bring these into the sprint

PR #524, opened 2022-08-27
feat: tune trivy scanning severity

wenzowski:
continues #522 to increase sensitivity and integrate with the github security tab
ref #366


marcellmueller (approved):
Looks good but seeing the results might keep me up at night. Does this mean we will have failing CI until we do some serious dependency and code cleanup?

wenzowski:
We'll need to keep an eye on the Security tab moving forward, and we'll triage on a sprint basis. The goal in merging this is to avoid introducing a regression. It looks really bad here because there's no data on the main branch yet, but once there is our checks will fail based on the diff between the PR and the main branch (ie new issues will cause failure).

PR #525, opened 2022-08-27
feat: trivy sensitivy to critical

wenzowski:
Now that the main branch has a comprehensive baseline, let's set PRs so they error out if they introduce new critical vulnerabilities where there is a known fix.

continues #524
ref #366


wenzowski:
Branch diffing is working correctly; this may not be needed. Closing for now.

PR #527, opened 2022-08-27
node 16

wenzowski:
looks like we need to upgrade node?
/leans on ci to tell us which lts versions work


wenzowski:
Note to self: land #528 before attempting to land this

VB-at-Bis (approved):
manual tests did not reveal any regressions

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -282,7 +282,7 @@ describe('The application form', () => {
 
     // luxon renders this differently in test?
     expect(
-      screen.getByText(/August 27, 2022, 9:51 a.m. PDT/)
+      screen.getByText(/August 27, 2022 at 9:51 a.m. PDT/)
```
wenzowski, line 285: It's a slight change, but it does affect copy so making sure we confirmed this adjustment with PO?
```diff
- August 27, 2022, 9:51 a.m. PDT
+ August 27, 2022 at 9:51 a.m. PDT
```

app/tests/components/Form/ApplicationFormStatus.test.tsx
```diff
@@ -65,7 +65,7 @@ describe('The application form', () => {
     componentTestingHelper.renderComponent();
 
     expect(screen.queryByText('Saving')).toBeNull();
-    expect(screen.getByText('Last saved: Jan. 1')).toBeInTheDocument();
+    expect(screen.getByText('Last saved: Jan 1')).toBeInTheDocument();
```
wenzowski, line 68: Another slight change, but also affects copy so making sure we confirmed this adjustment with PO?
```diff
- Last saved: Jan. 1
+ Last saved: Jan 1
```

VB-at-Bis: I'm currently trying to update code set explicit settings for Luxon to use defaultLocale="en-CA" which (hopefully) will solve formatting issues and allow us to revert those slight changes

VB-at-Bis: Actually, `en-CA` short month format does not have dot. 
https://metacpan.org/dist/DateTime-Locale/view/lib/DateTime/Locale/en_CA.pod#Medium and https://github.com/moment/luxon/blob/master/docs/formatting.md  (`DATE_MED`)

wenzowski: I don't personally have strong feelings about keeping/removing the dot either way, just making sure we've looped in PO in the migration to the new format.

VB-at-Bis:
@MarsRomer Quick question if you are Ok with minor change of the date formatting (caused by upgrade of node version). It seems `en-CA` locale does not have dot after month abbreviation:
Was: `Last saved: Jan. 1` now: `Last saved: Jan 1`

MarsRomer:
@VB-at-Bis  thanks for the flag, where would I see the difference in the dates?

VB-at-Bis:
On any page in top right corner when it is opened next day - `Last saved Sep 15` 
and on Homepage : extra seconds : August 19, 2027, 9:00**:00** a.m. PDT (see https://happo.io/a/336/p/1172/compare/d0d79504e3535cbc035ed575a794d7ccd990ad72/cd1df8dd7df6dda0feca55fb8835fc6eb7c8d040#Homepage%5C%20should%5C%20start%5C%20and%5C%20fill%5C%20the%5C%20first%5C%20page%5C%20of%5C%20the%5C%20form-default-2)

MarsRomer:
I think that the new format is ok.  Thanks for checking with me!

app/tests/components/Form/ApplicationFormStatus.test.tsx
```diff
@@ -59,22 +59,23 @@ describe('The application form', () => {
...
   });
 
   it('displays the error message if provided', () => {
     const mockCurrentTime = DateTime.utc(2020, 1, 2, 0, 0);
     Settings.now = () => mockCurrentTime.toMillis();
-    Settings.defaultZone = 'UTC';
+    Settings.defaultZone = "America/Vancouver";
```
matthieu-foucault, line 75: This line can probably be removed now that `renderComponent` will set the default date and zone

app/components/Form/ApplicationForm.tsx
```diff
@@ -155,6 +156,9 @@ const ApplicationForm: React.FC<Props> = ({
     query
   );
   const { id, rowId, formData, status } = application;
+  
+  Settings.defaultZone = "America/Vancouver";
+  Settings.defaultLocale = "en-CA";
```
matthieu-foucault, line 161: should we do this in `pages/_app`, so that those settings are applied for every page?

app/tests/components/TimeTravel.test.tsx
```diff
@@ -48,7 +48,7 @@ describe('The TimeTravel component', () => {
     fireEvent.mouseDown(datePicker);
     fireEvent.change(datePicker, { target: { value: "2020-01-02" } });
     await waitFor(() => {
-      // difference due to timezone
+      // difference due to timezone
```
matthieu-foucault, line 51: If we set the default time zone in this test, we won't need this one-day difference

app/tests/utils/componentTestingHelper.tsx
```diff
@@ -76,10 +77,16 @@ class ComponentTestingHelper<
   };
 
   public renderComponent(
+    useDefaultDateSettings: boolean,
     getPropsFromTestQuery: (data: TQuery['response']) => any = this.options
       .getPropsFromTestQuery,
     extraProps: any = this.options.defaultComponentProps
+
```
matthieu-foucault, line 84:
```suggestion
    useDefaultDateSettings: boolean,
    getPropsFromTestQuery: (data: TQuery['response']) => any = this.options
      .getPropsFromTestQuery,
    extraProps: any = this.options.defaultComponentProps
    useDefaultDateSettings = true
```
Setting it to `true` by default would make all of our test use the default zone and local, which is what we want. Not sure if this actually needs to be a parameter, do we have test cases that need a different default zone or locale?

VB-at-Bis: We have in tests: 
```   const mockCurrentTime = DateTime.utc(2020, 1, 1, 5, 0);
    Settings.now = () => mockCurrentTime.toMillis();
    Settings.defaultZone = "UTC";

    componentTestingHelper.loadQuery();
    componentTestingHelper.renderComponent(false); ```

Comment (matthieu-foucault) [reply app/tests/utils/componentTestingHelper.tsx:84]: Is there a reason why this specific test uses UTC, does it work if we set to the pacific time zone instead? I may have written this test and I think this was an arbitrary choice

File: app/tests/components/TimeTravel.test.tsx (hunk 5 above)
Comment (VB-at-Bis) [reply app/tests/components/TimeTravel.test.tsx]: For some reason test is failing
```  <input
                class="TimeTravel__StyledDatePicker-sc-wl9pxw-3 kIPtRb form-control"
                placeholder="YYYY-MM-DD"
                type="text"
                value="2020-01-01"
              />
            </div>
          </div>
        </div>
      </div>
    </body>

      48 |     fireEvent.mouseDown(datePicker);
      49 |     fireEvent.change(datePicker, { target: { value: "2020-01-02" } });
    > 50 |     await waitFor(() => {
         |           ^
      51 |       expect(screen.getByText(`Current date is: 2020-01-02`)).toBeInTheDocument();
      52 |     }); 
      53 |   });
```
while on local machine it passes

matthieu-foucault: Do we need to explicitly set the Luxon settings in this test?

app/tests/utils/componentTestingHelper.tsx
VB-at-Bis: ok, let me try to change from DateTime.utc(2020, 1, 2, 0, 0); to DateTime.local(2020, 1, 2, 0, { zone: "America/Vancouver" })

VB-at-Bis: Fixed

app/tests/components/TimeTravel.test.tsx
VB-at-Bis: Fixed by changing how we parse Date in TimeTravel (instead of using dateFormat use luxon .fromJSDate

app/components/Form/ApplicationForm.tsx
VB-at-Bis: Done

app/components/Form/ApplicationForm.tsx
```diff
@@ -157,6 +158,9 @@ const ApplicationForm: React.FC<Props> = ({
     query
   );
   const { id, rowId, formData, status, updatedAt } = application;
+  
+  Settings.defaultZone = "America/Vancouver";
+  Settings.defaultLocale = "en_CA";
```
matthieu-foucault, line 163: Can these lines be removed now that we're setting in in `_app`?

matthieu-foucault (changes requested):

app/components/Form/ApplicationForm.tsx
VB-at-Bis: Removed

matthieu-foucault (approved):

PR #531, opened 2022-08-29
Fix install-test-env CI step

matthieu-foucault:


.github/actions/dev-env-setup/action.yml
```diff
@@ -17,26 +17,23 @@ runs:
...
         echo "/home/runner/perl5/bin" >> $GITHUB_PATH
-    - uses: actions/cache@v2
+    - uses: actions/cache@v3
       id: asdf-cache
       with:
         path: |
           ~/.asdf
-        key: ${{ runner.os }}-asdf-cache-${{ hashFiles('.tool-versions') }}
-    - uses: actions/cache@v2
+        key: ${{ runner.os }}-asdf-cache-${{ hashFiles('.tool-versions') }}-v2
```
wenzowski, line 25: clobbering the old one didn't work? needed a fresh start?

.github/workflows/publish_deploy_image.yaml
```diff
@@ -25,21 +25,21 @@ jobs:
...
+          scan-type: fs
+          format: sarif
+          output: trivy-results.sarif
+          exit-code: '0'
+          ignore-unfixed: false
+          severity: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: 'trivy-results.sarif'
```
wenzowski, line 42: would love to have a yaml formatter going to we can limit this type of patch

.github/workflows/tests.yaml
```diff
@@ -5,62 +5,65 @@ env:
...
-
 jobs:
   install-test-env:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: dev env setup
         uses: ./.github/actions/dev-env-setup
+      - run: make install_dev_tools
+      - run: make stop_pg # cleanly stop the database, otherwise it will complain in the subsequent jobs
```
wenzowski, line 22: doesn't `install_dev_tools` depend on `stop_pg`? Is it restarting it again?

.github/workflows/tests.yaml
```diff
@@ -5,62 +5,65 @@ env:
...
+      - run: yarn install --frozen-lockfile --check-files
```
wenzowski, line 23: tyvm for `--check-files`

marcellmueller (approved):

wenzowski (commented):
Questions here releate to tech debt; should not block merge

.github/workflows/tests.yaml
matthieu-foucault: yes, the pgtap install starts pg

.github/actions/dev-env-setup/action.yml
matthieu-foucault: Yes, because if we have a cache hit, GH actions won't save the cache, and we had a cache hit given that the tool-versions wasn't changed

.github/workflows/tests.yaml
wenzowski: should it maybe not?

.github/actions/dev-env-setup/action.yml
wenzowski: maybe we should add a `hashfiles('.github/actions/dev-env-setup/action.yml')` in as well; bumping to cache@v3 should invalidate too

.github/workflows/tests.yaml
matthieu-foucault: it needs to start pg to install the extension and run its tests, but it should probably stop it after...

PR #532, opened 2022-08-29
463 reroute

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/board

Sets the nextjs basepath to `/applicantportal` and a redirect from `/` to `applicantportal` as well as the cypress base path.

### DIFF
--- a/app/next.config.js
+++ b/app/next.config.js
@@ -11,7 +11,7 @@ module.exports = {
         source: '/',
         destination: '/applicantportal',
         basePath: false,
-        permanent: false,
+        permanent: true,
       },
     ];
   },
--- a/app/next.config.js
+++ b/app/next.config.js
@@ -3,6 +3,18 @@
 // module.exports = nextConfig
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


app/next.config.js
```diff
@@ -3,6 +3,18 @@
...
 module.exports = {
+  basePath: '/applicantportal',
+
+  async redirects() {
+    return [
+      {
+        source: '/',
+        destination: '/applicantportal',
+        basePath: false,
+        permanent: false,
```
wenzowski, line 14:
```suggestion
        permanent: true,
```

wenzowski, line 14:
```suggestion
        permanent: true,
```

wenzowski (changes requested):
we want a 301 redirect not a 302 right?

marcellmueller:
> we want a 301 redirect not a 302 right?

Good point thank you.

wenzowski (approved):

PR #534, opened 2022-08-29
refactor: reorganise CI workflow

matthieu-foucault:
- use a `main` workflow to orchestrate the build, test and deploy stages
- separate tests into `test-code` and `test-containers`
- add a scheduled nightly workflow run
- remove unnecessary job steps (e.g. running the app before tests other than e2e)


wenzowski:
looks like this is going to cause all code scanning results to fail since we're changing the match key by changing the job names (and there's no way around that)

marcellmueller (approved):
Looks good

Issue #322, opened 2022-07-20
Helm chart linting in CI


marcellmueller:
We need to add a step to lint and verify our Helm chart in CI before a branch can be merged into main.

Risk of downtime if SSL certificate needs to be re-issued/requested from Entrust

- [ ] SA in our tools namespace that has permission to lint
- [ ] https://github.com/bcgov/cas-cif/blob/develop/.github/workflows/lint-chart.yml

eazeredo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/322?planning-poker) with Zenhub @AntBush @marcellmueller @matthieu-foucault @wenzowski

wenzowski:
https://github.com/bcgov/cas-cif/blob/develop/.github/workflows/lint-chart.yml

elliottSlee:
https://redhat-cop.github.io/ci/linting-testing-helm-charts.html

PR #544, opened 2022-08-29
Cutover to vanity URLs

matthieu-foucault:
- add support for www route in prod
- remove edge-terminated route


helm/app/templates/routes/secureRoute.yaml
```diff
@@ -32,3 +32,31 @@ spec:
     name: {{ template "ccbc.fullname" . }}
     weight: 100
   wildcardPolicy: None
+
+{{- if secureRoute.wwwRoute.enable }}
+---
+# The www route isn't managed by certbot to avoid duplicate certificate requests
```
wenzowski, line 38: tech debt: bcgov certbot should not attempt to fetch any cert matching `/^www\./`, would love to see this risk tracked in backlog and brought to PO. Per new ch12 policy we should be sending fix for this upstream if at all possible.

helm/app/values-prod.yaml
```diff
@@ -1,2 +1,3 @@
-replicaCount: 0 # prod should be offline until approval to go live is given
-
+secureRoute:
+  wwwRoute:
+    enable: true
```
wenzowski, line 3: this will certainly work, and we may find that as a rule we should always be adding the `www.` subdomain to all routes regardless of if the route represents apex or sub

wenzowski (approved):

wenzowski:
well [I'm stumped](https://github.com/bcgov/CONN-CCBC-portal/runs/8084429294?check_suite_focus=true#step:3:108) 🤷
something for the morning...

matthieu-foucault:
This is calling for #322 to be added

PR #546, opened 2022-08-29
fix: remove unnecessary deps

wenzowski:
doesn't look like we're using a bunch of stuff in prod? not sure why in package.

needed to remove dicer to fix CVE-2022-24434


marcellmueller (approved):
Nice, thanks for cleaning up.

PR #547, opened 2022-08-29
Refactor review page to use rjsf

matthieu-foucault:


app/components/Review/Components.tsx
```diff
@@ -0,0 +1,37 @@
+import styled from 'styled-components';
+
+export const StyledTitleRow = styled('th')`
+  padding: 16px !important;
```
marcellmueller, line 4: This is tech debt and was already in the codebase (I think this was me as well) but we should likely put in some css resets in the globals file so we can avoid using important which is an antipattern.

app/components/Review/Components.tsx
```diff
@@ -0,0 +1,37 @@
...
+  border-top: 1px solid rgba(0, 0, 0, 0.16);
+`;
+
+export const StyledH4 = styled('h4')`
+  margin: 0;
+`;
+
+export const StyledColLeft = styled('th')`
+  // Todo: workaround for Jest styled component theme prop error
```
marcellmueller, line 13: We can uncomment these as well since we are wrapping our tests with `<GlobalTheme/>` provider.

app/components/Form/ApplicationForm.tsx
```diff
@@ -191,11 +176,28 @@ const ApplicationForm: React.FC<Props> = ({
   const isWithdrawn = status === 'withdrawn';
   const isDraft = status === 'draft';
 
+  const isSubmitEnabled = useMemo(() => {
+    if (isWithdrawn) return false;
```
marcellmueller, line 180: I don't mind skipping curly brackets for a single return statement but this function feels weird with them all removed.

matthieu-foucault: Would blank lines between the conditions help readability? 

```ts
  const isSubmitEnabled = useMemo(() => {
    if (isWithdrawn) return false;

    if (sectionName === 'review')
      return noErrors || formData.review?.acknowledgeIncomplete;

    if (sectionName === 'acknowledgements')
      return areAllAcknowledgementsChecked;
      
    if (sectionName === 'submission') return areAllSubmissionFieldsSet;

    return false;
  }, [
    sectionName,
    noErrors,
    areAllAcknowledgementsChecked,
    areAllSubmissionFieldsSet,
    isWithdrawn,
    formData,
  ]);
```

marcellmueller: It's fine as is just looks more like ruby than js :p

app/components/Review/Components.tsx
matthieu-foucault: Created #693

VB-at-Bis (approved):
Looks good

PR #549, opened 2022-08-30
ci: depcheck prevents unneeded deps

wenzowski:
`depcheck` does not appear to be parsing correctly. Anyone have suggestions on something that works with typescript?

### DIFF
--- a/.github/workflows/test-code.yaml
+++ b/.github/workflows/test-code.yaml
@@ -67,6 +67,19 @@ jobs:
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v2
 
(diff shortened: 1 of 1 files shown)


wenzowski:
Per depcheck/depcheck#319 we should try configuring parsers/detectors:

```js
  parsers: { // the target parsers
    '**/*.ts': depcheck.parser.typescript
  },
  detectors: [ // the target detectors
    depcheck.detector.requireCallExpression,
    depcheck.detector.importDeclaration,
    depcheck.detector.typescriptImportType,
    depcheck.detector.typescriptImportEqualsDeclaration
  ],
```

PR #551, opened 2022-08-30
fix: secureRoute is a value not a function

wenzowski:
continues #544 to resolve [failed deploy](https://github.com/bcgov/CONN-CCBC-portal/runs/8083900978)

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -41,7 +41,11 @@ RUN sed -i -nr '/node|yarn/p' ${APP_HOME}/.tool-versions && \
   cat ${APP_HOME}/.tool-versions | cut -f 1 -d ' ' | xargs -n 1 asdf plugin-add && \
   asdf plugin-update --all && \
   asdf install && \
-  asdf reshim
+  asdf reshim && \
+  pushd ${APP_HOME}/.asdf/installs/nodejs/$(awk '/^nodejs/ { print $2 }' .tool-versions)/lib && \
+  npm i npm corepack && \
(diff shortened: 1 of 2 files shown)


app/Dockerfile
```diff
@@ -41,7 +41,11 @@ RUN sed -i -nr '/node|yarn/p' ${APP_HOME}/.tool-versions && \
   cat ${APP_HOME}/.tool-versions | cut -f 1 -d ' ' | xargs -n 1 asdf plugin-add && \
   asdf plugin-update --all && \
   asdf install && \
-  asdf reshim
+  asdf reshim && \
+  pushd ${APP_HOME}/.asdf/installs/nodejs/$(awk '/^nodejs/ { print $2 }' .tool-versions)/lib && \
+  npm i npm corepack && \
+  rm -f package.json package-lock.json && \
+  popd
```
matthieu-foucault, line 48: Is this intentionally in this PR?

matthieu-foucault (approved):

app/Dockerfile
wenzowski: yes I rebased it on the main I already merged but haven't pushed yet because I'm waiting for happo to pass (don't judge)

wenzowski: it'll disappear on its own

wenzowski: now it's gone 🪄

matthieu-foucault: magic

PR #553, opened 2022-08-30
feat: skip keycloak login page and redirect to BCeID login

matthieu-foucault:
Currently blocked on the CSS app being configured for BCeID both


matthieu-foucault:
This will need to be implemented with two different login buttons

PR #554, opened 2022-08-30
454 -  Change Evidence of connectivity speeds upload

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/454

This PR also reverts the other supporting documents schema back to an object and not a function as that conditional required isn't needed anymore.

### DIFF
--- a/app/formSchema/pages/supportingDocuments.ts
+++ b/app/formSchema/pages/supportingDocuments.ts
@@ -1,55 +1,49 @@
-const supportingDocuments = (isEvidenceOfConnectivityRequired) => {
-  // This page is a function because evidenceOfConnectivitySpeeds is required depending on if an upload
-  // field from another page was filled. So we check the form data in schema.ts and pass it here if it is filled.
-
-  return {
-    supportingDocuments: {
-      title: 'Supporting documents',
-      type: 'object',
-      description:
-        'Please upload additional attachments. Please avoid using special characters in the file name. The maximum size per file is 100MB',
-      required: [
-        'copiesOfRegistration',
-        'preparedFinancialStatements',
-        'logicalNetworkDiagram',
-        'projectSchedule',
-        'communityRuralDevelopmentBenefits',
-        isEvidenceOfConnectivityRequired ? 'evidenceOfConnectivitySpeeds' : '',
-      ],
-      properties: {
-        copiesOfRegistration: {
-          title:
-            'Copies of registration and other relevant documents related to incorporation, limited partnership, joint venture, not-for-profit status, etc.',
-          type: 'string',
-        },
-        preparedFinancialStatements: {
-          title:
-            'Independently prepared financial statements for the last three (3) years',
-          type: 'string',
-        },
-        logicalNetworkDiagram: {
-          title:
-            'Please refer to Annex 3 of the application guide for the Logical Network Diagram requirements',
-          type: 'string',
-        },
-        projectSchedule: {
-          title:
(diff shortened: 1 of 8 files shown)


app/formSchema/schema.ts
```diff
@@ -22,10 +22,7 @@ import {
   templateUploads,
 } from './pages';
 
-const useSchema = (formData) => {
-  const isEvidenceOfConnectivityRequired =
-    formData?.templateUploads?.supportingConnectivityEvidence;
-
+const useSchema = () => {
```
matthieu-foucault, line 25: can schema be an exported constant? That will avoid recreating a new schema object all the time and triggering unnecessary re-renders

app/formSchema/uiSchema.ts
```diff
@@ -654,10 +653,9 @@ const uiSchema = {
...
     },
   },
-  evidenceOfConnectivitySpeeds: {
+  otherSupportingMaterials: {
     'ui:widget': 'FileWidget',
-    'ui:description':
-      'Evidence of connectivity speeds (e.g, screen captures of Speed Test results)',
+    'ui:description': 'Other supporting materials (if applicable)',
     'ui:options': {
       maxLength: MAX_LONG_INPUT_LENGTH,
```
matthieu-foucault, line 660: do we need an option to hide the "optional" keyword?

app/formSchema/pages/supportingDocuments.ts
```diff
@@ -1,55 +1,49 @@
...
+      },
+      communityRuralDevelopmentBenefits: {
+        title:
+          'Please refer to Annex 3 of the application guide for the community letters of support requirements. If you do not currently have these, you will have a 30-day grace period after the intake closes to submit them.',
+        type: 'string',
+      },
+      otherSupportingMaterials: {
+        title:
+          'Please upload any other files such as evidence of connectivity speeds (such as screen captures of speed test results), a written commitment to facilitating access to Passive Infrastructure, evidence of imminent access to Third Party infrastructure, evidence of other funding sources, network information and/or coverage for ISED, or other documents to support this application.',
+        type: 'string',
```
matthieu-foucault, line 43: should our schemas ask for an array of strings when we allow multiple uploads? How are we storing the uploaded files ids in the json data?

matthieu-foucault (changes requested):

app/formSchema/uiSchema.ts
wenzowski: any chance we could implement as a string label to render for optional? This way we'd be enforcing that all optional fields have some label even if it isn't `(optional)`

marcellmueller: > any chance we could implement as a string label to render for optional? This way we'd be enforcing that all optional fields have some label even if it isn't `(optional)`

That's well out of scope of this ticket as well as there are some fields iirc that are not required but they wanted optional hidden.

marcellmueller: > do we need an option to hide the "optional" keyword?

good catch, I did that and renamed `ignoreOptional` to `hideOptional` to make it a bit more clear

app/formSchema/pages/supportingDocuments.ts
matthieu-foucault: This is out of scope but I'd love to discuss this at some point

matthieu-foucault (approved):

PR #555, opened 2022-08-30
fix: re-add CI env vars that were lost in refactor

matthieu-foucault:
This should fix our deployment


marcellmueller (approved):

wenzowski (approved):
That’ll do it

PR #559, opened 2022-08-30
415 - Switch up some requireds, optionals, & if applicables

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/415

@wenzowski I added `altOptionalText` to `ui:options` so we can override the optional text as you asked. I decided to keep the `hideOptional` flag to hide the label on non required fields but I'm open to other suggestions.

### DIFF
--- a/app/formSchema/pages/authorizedContact.ts
+++ b/app/formSchema/pages/authorizedContact.ts
@@ -9,6 +9,7 @@ const authorizedContact = {
       'authPositionTitle',
       'authEmail',
       'authTelephone',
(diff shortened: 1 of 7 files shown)


wenzowski:
> I decided to keep the hideOptional flag to hide the label on non required fields
Makes sense. Small increments of value. I'd love to track eliminating `hideOptional` as a new tech debt ticket though, if we can?

wenzowski (approved):

PR #563, opened 2022-08-30
494 block new applications

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/494

The custom `create_application` mutation is being handled in #566 now, I've updated the tickets accordingly.

### DIFF
--- a/app/pages/dashboard.tsx
+++ b/app/pages/dashboard.tsx
@@ -4,6 +4,8 @@ import { withRelay, RelayProps } from 'relay-nextjs';
 import defaultRelayOptions from '../lib/relay/withRelayOptions';
 import { usePreloadedQuery } from 'react-relay/hooks';
 import { graphql } from 'react-relay';
+import Alert from '@button-inc/bcgov-theme/Alert';
+import styled from 'styled-components';
 import StyledGovButton from '../components/StyledGovButton';
 import { useCreateApplicationMutation } from '../schema/mutations/application/createApplication';
 import { Layout } from '../components';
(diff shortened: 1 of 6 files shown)


app/pages/dashboard.tsx
```diff
@@ -89,17 +98,25 @@ const Dashboard = ({ preloadedQuery }: RelayProps<{}, dashboardQuery>) => {
...
-          <p>There are no currently open intakes.</p>
+          <StyledAlert size="small" variant="warning">
+            <p>
+              New applications will be accepted after updates to ISED&lsquo;s
+              Eligibility Mapping tool are released.
+            </p>
+            <p>
+              Please check this page after <b>September 15</b> for an update.
+            </p>
+          </StyledAlert>
```
matthieu-foucault, line 109: Wanna put this in a component to reuse in both pages?

app/tests/pages/index.test.tsx
```diff
@@ -39,4 +58,26 @@ describe('The index page', () => {
...
     expect(screen.getByText('Go to dashboard')).toBeInTheDocument();
   });
+
+  it('Does not display alert message when there is an open intake', async () => {
+    pageTestingHelper.loadQuery();
+    pageTestingHelper.renderPage();
+
+    expect(
+      screen.queryByText(
+        `New applications will be accepted after updates to ISED‘s Eligibility Mapping tool are released.`
```
matthieu-foucault, line 68: Same comment as the previous PR regarding having this text in a variable. If we're asked to change the text, this test will not fail (but it should)

matthieu-foucault (changes requested):
Looks good, just a bit of code cleanup

app/tests/pages/index.test.tsx
marcellmueller: Sounds good, moved the message to a constant.

PR #566, opened 2022-08-31
feat: application status type table and new behaviours

AntBush:
Adds application_status_type and modifies application_status to meet the DB changes. 

Adds a create application mutation to add a draft status and prevents creating applications when there are no available intakes.

https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/470


app/schema/mutations/application/withdrawApplication.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { withdrawApplicationMutation } from '__generated__/withdrawApplicationMutation.graphql';
+import useMutationWithErrorMessage from '../useMutationWithErrorMessage';
+
+const mutation = graphql`
+  mutation withdrawApplicationMutation($input: WithdrawApplicationInput!) {
+    withdrawApplication(input: $input) {
+      application {
+        updatedAt
+        ccbcNumber
+        intakeId
```
matthieu-foucault, line 12: you'll want to fetch the status here. Ideally you'd have a fragment here, so that the component using the mutation gets all the data it needs

matthieu-foucault (changes requested):

db/deploy/mutations/submit_application.sql
```diff
@@ -12,7 +12,7 @@ declare
   application_status varchar;
 begin
 
-  select status from ccbc_public.application where id = application_row_id into application_status;
+  select status into application_status from ccbc_public.application_status where application_id = application_row_id ORDER BY id desc;
```
matthieu-foucault, line 15:
```suggestion
  select ccbc_public.application_status(
    select row(application.*)::ccbc_public.application from ccbc_public.application where id = application_row_id
  ) into status;
```

db/deploy/trigger_functions/create_draft_status.sql
```diff
@@ -0,0 +1,15 @@
+-- Deploy ccbc:trigger_functions/create_draft_status to pg
+BEGIN;
+
+create or replace function ccbc_private.create_draft_status()
```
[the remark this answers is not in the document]
AntBush: Removed!

db/deploy/mutations/submit_application.sql
AntBush: Added! Required an additional bracket around the nested select

app/schema/mutations/application/withdrawApplication.ts
AntBush: I've added in status, and you mean to create a fragment for the dashboard and then reuse that fragment here right?

matthieu-foucault: Correct, that would ensure that the mutation return and the fragment that is displayed when triggering the mutation are always in sync

db/test/unit/tables/application_status_type_test.sql
```diff
@@ -0,0 +1,34 @@
+begin;
+
+select * from no_plan();
```
matthieu-foucault, line 3: select plan(3);

db/deploy/computed_columns/application_status.sql
```diff
@@ -0,0 +1,15 @@
...
+
+create or replace function ccbc_public.application_status(application ccbc_public.application) returns text as 
+$$
+select status from ccbc_public.application_status 
+    where application_id= application.id
+    order by id desc limit 1;
+$$ language sql stable;
+
+comment on function ccbc_public.application_status is 'Computed column to return status of an application';
+
```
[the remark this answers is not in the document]
AntBush: Apparently not? But I've gone ahead and added it, I do agree with your other comment that it's odd that these were still executing without explicit grant

matthieu-foucault (changes requested):
one small change

db/deploy/tables/application_status_type.sql
```diff
@@ -0,0 +1,30 @@
...
+end
+$grant$;
+
+comment on table ccbc_public.application_status_type is 'Table containing the different statuses that can be assigned to an application';
+
+comment on column ccbc_public.application_status_type.name is 'Name of and primary key of the status of an application';
+
+comment on column ccbc_public.application_status_type.description is 'Description of the status type';
+
+-- move this out once we have established a way to deploy prod data
```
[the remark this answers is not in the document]
AntBush: I was under the assumption we'd pick up the same deployment data style as cas-cif, storing data under the `data/dev data/test` etc before release, but I'm guessing there's not much time for that?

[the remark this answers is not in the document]
matthieu-foucault: This has to be done today then :smile:

matthieu-foucault (approved):

PR #589, opened 2022-09-02
ci: limit concurrency

wenzowski:
I'll concede that I used too many concurrencies

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,6 +3,10 @@ name: build
 
 on: workflow_call
 
+concurrency:
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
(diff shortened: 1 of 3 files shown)


marcellmueller (approved):
Are you going to test this by running your node branches again? :sweat_smile:

.github/workflows/build.yaml
```diff
@@ -3,6 +3,10 @@ name: build
 
 on: workflow_call
 
+concurrency:
+  group: build-${{ github.ref }}
```
AntBush, line 7: Should we use ` group: ${{ github.workflow }}-${{ github.ref }}` so that way no one goofs in the future and adds a group name of the same name?

wenzowski: unfortunately no dice on this #590

PR #590, opened 2022-09-02
ci: refactor concurrency key to inherit workflow name

wenzowski:
as recommended by @AntBush


AntBush (approved):
👍

wenzowski:
sadly this change looks like it results in the `test-containers` jobs cancelling the `test-code` jobs

PR #591, opened 2022-09-02
refactor: separate uischema into separate pages

marcellmueller:


app/formSchema/uiSchema/uiSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+    ...organizationLocation.uiOrder,
+    ...contactInformation.uiOrder,
+    ...authorizedContact.uiOrder,
+    ...alternateContact.uiOrder,
+    // ...review.uiOrder,
+    ...acknowledgements.uiOrder,
+    ...submission.uiOrder,
+  ],
+  'ui:title': '',
+  ...projectInformation,
```
matthieu-foucault, line 50:
```suggestion
  projectInformation,
```
UI schema should match the shape of the data, instead of flattening the objects.
In `ApplicationForm`, you'll want to always use `uiSchema[sectionName]`, instead of the current behaviour

app/formSchema/uiSchema/uiSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+  projectPlan,
+  submission,
+  supportingDocuments,
+  review,
+  techSolution,
+  templateUploads,
+} from './pages';
+
+const uiSchema = {
+  'ui:order': [
```
matthieu-foucault, line 26: This should just be the order of the pages

matthieu-foucault (changes requested):

app/formSchema/uiSchema/uiSchema.ts
marcellmueller: Okay take another look when you have time. Our testing caught a couple of bugs and the Happo screenshot testing helped me fix the missing `ui:field` in the review page.

matthieu-foucault (approved):
Looks good as far as I can tell. Happo diffs are either negligible (whitespace) or spurious diffs

PR #592, opened 2022-09-02
feat: only deploy test and prod with tag

AntBush:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/582

- [x] update release-it
- [x] separate release workflow

[Action ](https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3055162551) for attempt to deploy with a tag
[Action ](https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3055048718) for attempt to deploy without a tag (deploy step is skipped)


.github/workflows/deploy.yaml
```diff
@@ -22,6 +22,21 @@ env:
...
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Is Tagged Commit
+        id: tagVersion
+        # Check if one of the commits is associated with a tag
+        run: |
+          echo ::set-output name=tagVersion::$( git tag --contains ${{ github.sha  }} | grep v*)
```
matthieu-foucault, line 38: Looks like this only checks the last commit that was pushed. This would require that we fast-forward the release branch. Although this is the process that is documented, I would suggest that we update that process (and the documentation in the readme) to allow users to click the github ui merge button. 
Looking at the [git docs](https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---mergedltcommitgt), and confirmed with local testing, this should work: 
`git tag --merge ${{ github.sha  }} --no-merge ${{ github.event.before }}`

matthieu-foucault (changes requested):

.github/workflows/deploy.yaml
AntBush: In older version of git these two commands can't be combined. I'm sure github runners are using the latest version but it would be funny if it were an old enough version that this was not compatible with the use of this command

AntBush:
Looking at https://github.com/bcgov/CONN-CCBC-portal/pull/660#issuecomment-1248734204 the code present seems to work exactly as expected. @matthieu-foucault let me know if I'm missing anything or if the behaviour is not defined as you'd expect.

matthieu-foucault (approved):
Looks good, thanks for the walkthrough

PR #593, opened 2022-09-02
ci: add ossf scorecards

wenzowski:
something to consider
https://github.com/ossf/scorecard-action


marcellmueller (approved):
No badges? Also should we change readme badge from experimental?

wenzowski:
Yes I figure we've got to at least be [Maturing](https://github.com/bcgov/repomountie/blob/master/doc/lifecycle-badges.md) at this point. Figured we'd assess ossf badging once we see what it finds next week.

PR #594, opened 2022-09-05
Fixed duplicate attachment issue

VB-at-Bis:
- removed duplicate uploadFieldDefinitions which is also defined in app/backend/lib/graphql/index.ts


app/.postgraphilerc.js
```diff
@@ -11,12 +11,6 @@ module.exports = {
       pgArchivedColumnName: 'archived_at',
       pgArchivedColumnImpliesVisible: false,
       pgArchivedRelations: false,
-      uploadFieldDefinitions: [
-        {
-          match: ({ schema, table, column, tags }) =>
-            table === 'attachment' && column === 'file',
-        },
-      ],
```
matthieu-foucault, line 19: It is unclear to me how this change is related to the bug. Removing this also alters our generated graphql schema, which I don't think is desirable.

matthieu-foucault (changes requested):
I am not sure how the changes in this PR are related to the bug. Can we have some tests to ensure that the root cause of the issue was found and fixed?

app/.postgraphilerc.js
VB-at-Bis: @matthieu-foucault  I had to remove  uploadFieldDefinitions here because it is also defined in app/backend/lib/graphql/index.ts, which references `app/.postgraphilerc.js`. It looks like uploadFieldDefinitions plugin was firing twice.
I tried to remove same block in app/backend/lib/graphql/index.ts but it resulted in runtime errors when file was uploaded.

VB-at-Bis:
>  Can we have some tests to ensure that the root cause of the issue was found and fixed?
It was tested locally (uploaded file appears once in the database now). I did not find any tests that verify if uploaded file appears in database, so I guess writing tests for file upload feature could be a separate ticket.

matthieu-foucault:
@VB-at-Bis would you be able to rebase your branch on `main` rather that making a merge commit, as per our [team agreements](https://github.com/bcgov/CONN-CCBC-portal/blob/main/docs/Team_Agreements.md#development-and-version-control). If this workflow doesn't work for you let's discuss it and consider updating the agreement.

VB-at-Bis:
> @VB-at-Bis would you be able to rebase your branch on `main` rather that making a merge commit, as per our [team agreements](https://github.com/bcgov/CONN-CCBC-portal/blob/main/docs/Team_Agreements.md#development-and-version-control). If this workflow doesn't work for you let's discuss it and consider updating the agreement.

Sure I will rebase.. Last commit 88454e5 happened because I noticed GitHub message `This branch is out-of-date with the base branch` and I clicked on `Update branch`.  :) Sorry, I did not think it would cause problems
image

matthieu-foucault:
@VB-at-Bis You do have to rebase from your local environment though, not the local UI, as the github rebase button will not sign your commits

app/.postgraphilerc.js
matthieu-foucault: Although there is a bit of code that is duplicated, I don't think that the plugin is added twice, as the `uploadFieldDefinitions` value is being overridden here: https://github.com/bcgov/CONN-CCBC-portal/blob/c260600bb22a1804c769a5ad645fec8dc3ab238b/app/backend/lib/graphql/index.ts#L42-L51

As for writing tests, it is part of our team's definition of done, so I don't think it should be a separate ticket.

VB-at-Bis: It seems it was not overriden but added second time and that's what was causing duplication. I have no other explanation why issue was happening and disappeared after I removed duplicate in  app/.postgraphilerc.js
I totally agree with writing tests as part of our team's definition of done, but unfortunately, file upload feature was implemented without any tests.. and e2e Cypress tests have comments like `  // // Todo: file uploads` etc.

VB-at-Bis:
> Approving. However, until we have tests we can't be sure that the bug is fixed, and that the behaviour observed locally wasn't a lack of race condition. I was unable to reproduce the bug locally

Agree. We need to have e2e tests for file upload that would check DB.

PR #603, opened 2022-09-06
feat: increase trivy timeout

matthieu-foucault:


marcellmueller (approved):
Looks good I was just confused why `timeout-minutes: 60` on line 56 wasn't doing anything. Should we remove?

matthieu-foucault:
`timeout-minutes` is for the entire job, whereas the 10min timeout is for the trivy scan itself

PR #616, opened 2022-09-08
eslint

marcellmueller:


marcellmueller:
@matthieu-foucault I removed the warning rules and disabled eslint during nextjs build though I'm not sure what the next step is or how to bypass the code scanning failure.

matthieu-foucault:
Since we are introducing new code scanning errors with this PR, the process will be to temporarily remove the check as "required". Let's do this when we're ready to merge so that we don't introduce errors in our other PRs in progress.

marcellmueller:
@matthieu-foucault take a look at this and let me know if there is anything you'd like for me to disable.

matthieu-foucault (approved):

PR #619, opened 2022-09-08
613 mitigate for os upgrade

marcellmueller:


matthieu-foucault (changes requested):
This increase in CPU requests will lead to an extra 2584m used by our chart, not counting extra replicas when we redeploy, and we only have 2290 left in our quota

matthieu-foucault (approved):
nevermind, re-did the math, looks good

PR #620, opened 2022-09-08
450 calculate totals

marcellmueller:
https://app.zenhub.com/workspaces/connectivity-intake-61c0c1204c8bcb001491c5f3/board

### DIFF
--- a/app/cypress/integration/index.spec.js
+++ b/app/cypress/integration/index.spec.js
@@ -176,10 +176,6 @@ context('Homepage', () => {
       '[id="root_otherFundingSourcesArray_0_requestedFundingPartner2627"]'
     ).type(123);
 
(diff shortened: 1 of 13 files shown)


app/cypress/integration/index.spec.js
```diff
@@ -113,16 +113,12 @@ context('Homepage', () => {
     cy.get('[id="root_fundingRequestedCCBC2526"]').type(123);
     cy.get('[id="root_fundingRequestedCCBC2627"]').type(123);
 
-    cy.get('[id="root_totalFundingRequestedCCBC"]').type(12345);
```
matthieu-foucault, line 116: we could assert that the total is the expected value here

matthieu-foucault (approved):

PR #632, opened 2022-09-12
295 automate submission fields

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/295

### DIFF
--- a/app/components/Layout.tsx
+++ b/app/components/Layout.tsx
@@ -17,7 +17,7 @@ const StyledLayout = styled('div')`
   min-height: 100vh;
 `;
 const StyledMain = styled('main')`
-  margin: 50px auto;
+  margin: auto 50px;
   max-width: ${(props) => props.theme.width.pageMaxWidth};
   flex: 1;
   padding: 2em 1.5em;
--- a/app/components/Layout.tsx
+++ b/app/components/Layout.tsx
@@ -17,7 +17,7 @@ const StyledLayout = styled('div')`
   min-height: 100vh;
 `;
 const StyledMain = styled('main')`
-  margin: auto 50px;
+  margin: 50px auto;
   max-width: ${(props) => props.theme.width.pageMaxWidth};
   flex: 1;
   padding: 2em 1.5em;
--- a/app/cypress/integration/index.spec.js
+++ b/app/cypress/integration/index.spec.js
@@ -468,11 +468,6 @@ context('Homepage', () => {
 
     cy.get('h1').contains('Submission');
 
-    cy.get('input[id="root_submissionCompletedFor"]').type('test');
-
-    cy.get('#root_submissionDate').click();
-    cy.get(`.react-datepicker__day--001`).eq(2).click({ force: true });
-
     cy.get('input[id="root_submissionCompletedBy"]').type('test');
     cy.get('input[id="root_submissionTitle"]').type('test');
 
--- a/app/cypress/integration/index.spec.js
+++ b/app/cypress/integration/index.spec.js
@@ -1,4 +1,6 @@
 /* eslint-disable cypress/no-unnecessary-waiting */
+import { dateTimeFormat } from '../../lib/theme/functions/formatDates';
+
 context('Homepage', () => {
   beforeEach(function () {
     cy.sqlFixture('dev/001_intake');
@@ -133,8 +135,6 @@ context('Homepage', () => {
     cy.get('[id="root_infrastructureBankFunding2526"]').type(123);
 
     cy.get('input[id="root_otherFundingSources-0"]').parent().click();
-
-    cy.get('[id="root_otherFundingSourcesArray_0_fundingPartnersName"]');
 
     cy.get('[id="root_otherFundingSourcesArray_0_fundingPartnersName"]').type(
       'test'
@@ -293,7 +293,7 @@ context('Homepage', () => {
 
     cy.get('input[id="root_typeOfOrganization-0"]').parent().click();
 
-    cy.get('input[id="root_organizationName"]').type('test');
+    cy.get('input[id="root_organizationName"]').type('Test org name');
 
     cy.get('input[id="root_isNameLegalName-0"]').parent().click();
 
@@ -468,8 +468,18 @@ context('Homepage', () => {
 
     cy.get('h1').contains('Submission');
 
+    cy.get('[id="root_submissionCompletedFor"]').should(
+      'have.text',
+      'Test org name'
+    );
+
     cy.get('input[id="root_submissionCompletedBy"]').type('test');
     cy.get('input[id="root_submissionTitle"]').type('test');
+
+    cy.get('[id="root_submissionDate"]').should(
+      'have.text',
+      dateTimeFormat(new Date(), 'date_year_first')
+    );
 
     cy.get('body').happoScreenshot();
 
--- a/app/formSchema/pages/submission.ts
+++ b/app/formSchema/pages/submission.ts
@@ -14,16 +14,16 @@ const submission = {
         title: 'Completed for (Legal organization name)',
         type: 'string',
       },
-      submissionDate: {
-        title: 'On this date (YYYY-MM-DD)',
-        type: 'string',
-      },
       submissionCompletedBy: {
         title: 'Completed by',
         type: 'string',
       },
       submissionTitle: {
         title: 'Title',
+        type: 'string',
+      },
+      submissionDate: {
+        title: 'On this date (YYYY-MM-DD)',
         type: 'string',
       },
     },
--- a/app/formSchema/uiSchema/pages/submission.ts
+++ b/app/formSchema/uiSchema/pages/submission.ts
@@ -8,10 +8,7 @@ const submission = {
   'ui:title': '',
   'ui:field': 'SubmissionField',
   submissionCompletedFor: {
-    'ui:widget': 'ReadOnlySubmissionWidget',
-    'ui:options': {
-      'field-name': 'submissionCompletedFor',
-    },
+    'ui:widget': 'SubmissionCompletedForWidget',
   },
   submissionDate: {
     'ui:widget': 'ReadOnlySubmissionWidget',
--- a/app/formSchema/uiSchema/pages/submission.ts
+++ b/app/formSchema/uiSchema/pages/submission.ts
@@ -1,14 +1,19 @@
+import { ReadOnlySubmissionWidget } from '../../../lib/theme/widgets';
+
 const submission = {
   'ui:order': [
     'submissionCompletedFor',
-    'submissionDate',
     'submissionCompletedBy',
     'submissionTitle',
+    'submissionDate',
   ],
   'ui:title': '',
   'ui:field': 'SubmissionField',
+  submissionCompletedFor: {
+    'ui:widget': ReadOnlySubmissionWidget,
+  },
   submissionDate: {
-    'ui:widget': 'DatePickerWidget',
+    'ui:widget': ReadOnlySubmissionWidget,
   },
 };
 
--- a/app/formSchema/uiSchema/pages/submission.ts
+++ b/app/formSchema/uiSchema/pages/submission.ts
@@ -1,5 +1,3 @@
-import { ReadOnlySubmissionWidget } from '../../../lib/theme/widgets';
-
 const submission = {
   'ui:order': [
     'submissionCompletedFor',
@@ -10,10 +8,13 @@ const submission = {
   'ui:title': '',
   'ui:field': 'SubmissionField',
   submissionCompletedFor: {
-    'ui:widget': ReadOnlySubmissionWidget,
+    'ui:widget': 'ReadOnlySubmissionWidget',
+    'ui:options': {
+      'field-name': 'submissionCompletedFor',
+    },
   },
   submissionDate: {
-    'ui:widget': ReadOnlySubmissionWidget,
+    'ui:widget': 'ReadOnlySubmissionWidget',
   },
 };
 
--- a/app/lib/theme/widgets/ReadOnlySubmissionWidget.tsx
+++ b/app/lib/theme/widgets/ReadOnlySubmissionWidget.tsx
@@ -1,0 +1,52 @@
+import { useRouter } from 'next/router';
+import Link from 'next/link';
+import { WidgetProps } from '@rjsf/core';
+import styled from 'styled-components';
+import schema from '../../../formSchema/schema';
+
+const StyledContainer = styled('div')`
+  margin-bottom: 8px;
+`;
+
+const StyledValue = styled('div')`
+  margin-top: 12px;
+  margin-bottom: 4px;
+  padding: 0.6em 0;
+  white-space: nowrap;
+`;
+
+const StyledError = styled('div')`
+  color: #e71f1f;
+  font-weight: 700;
+`;
+
+const StyledLink = styled(Link)`
+  color: #e71f1f;
+`;
+
+const ReadOnlySubmissionWidget: React.FC<WidgetProps> = ({ value }) => {
+  const router = useRouter();
+  const { id } = router.query;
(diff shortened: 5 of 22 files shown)


app/formSchema/pages/submission.ts
```diff
@@ -26,6 +22,10 @@ const submission = {
         title: 'Title',
         type: 'string',
       },
+      submissionDate: {
+        title: 'On this date (YYYY-MM-DD)',
+        type: 'string',
+      },
```
matthieu-foucault, line 28: Is this change needed?

app/formSchema/uiSchema/pages/submission.ts
```diff
@@ -1,14 +1,19 @@
...
+    'submissionDate',
   ],
   'ui:title': '',
   'ui:field': 'SubmissionField',
+  submissionCompletedFor: {
+    'ui:widget': ReadOnlySubmissionWidget,
+  },
   submissionDate: {
-    'ui:widget': 'DatePickerWidget',
+    'ui:widget': ReadOnlySubmissionWidget,
```
matthieu-foucault, line 16:
```suggestion
    'ui:widget': 'ReadOnlySubmissionWidget',
```

Widgets should be strings and added to the theme, otherwise themes won't work

app/lib/theme/widgets/ReadOnlySubmissionWidget.tsx
```diff
@@ -0,0 +1,52 @@
...
+        {value ? (
+          value
+        ) : (
+          <StyledError>
+            No legal organization name was provided. Please return to the{' '}
+            <StyledLink href={`/form/${id}/${pageNumber}`}>
+              Organization Profile
+            </StyledLink>{' '}
+            page and enter one.
+          </StyledError>
```
matthieu-foucault, line 45: Are we missing a test for this?

db/deploy/mutations/submit_application.sql
```diff
@@ -24,8 +28,31 @@ begin
...
+  select form_data -> 'submission' ->> 'submissionCompletedFor',
+    form_data -> 'submission' ->> 'submissionCompletedBy',
+    form_data -> 'submission' ->> 'submissionTitle',
+    form_data -> 'submission' ->> 'submissionDate'
+   from ccbc_public.application where id = application_row_id
+   into submission_completed_for, submission_completed_by, submission_title, submission_date;
+
+  if coalesce(submission_completed_for, '') = '' then
+    raise 'The application cannot be submitted as the submission field submission_completed_for is null or empty';
+end if;
```
matthieu-foucault, line 40:
```suggestion
  end if;
```

db/sqitch.plan
```diff
@@ -32,3 +32,4 @@ mutations/create_application 2022-08-31T16:16:23Z Anthony Bushara <anthony@butto
 @1.0.1 2022-09-02T00:40:17Z Matthieu Foucault <matthieu@button.is> # release v1.0.1
 @1.1.0 2022-09-06T23:59:22Z Matthieu Foucault <matthieu@button.is> # release v1.1.0
 @1.1.1 2022-09-09T20:35:47Z Marcel Mueller,,, <marcel@Thinkpad> # release v1.1.1
+mutations/submit_application [mutations/submit_application@1.0.0] 2022-09-13T19:52:43Z Marcel Mueller,,, <marcel@button.is> # Check all submission fields in form_data are filled before submission
```
matthieu-foucault, line 35:
```suggestion
mutations/submit_application [mutations/submit_application@1.0.0] 2022-09-13T19:52:43Z Marcel Mueller <marcel@button.is> # Check all submission fields in form_data are filled before submission
```

You should configure sqitch so that the correct user name is added in the plan

db/test/unit/mutations/submit_application_test.sql
```diff
@@ -89,6 +98,34 @@ select results_eq(
   'Increases the ccbc number when submitting applications'
 );
 
+select throws_like(
+  $$
+    select ccbc_public.submit_application(5)
+  $$,
+  'The application cannot be submitted as the submission field submission_date is null or empty'
```
matthieu-foucault, line 105:
```suggestion
  'The application cannot be submitted as the submission field submission_date is null or empty',
  'Throws an exception if submission_date is empty'
```
Adding test descriptions will improve logging if the tests fail

db/revert/mutations/submit_application@1.0.0.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ccbc:mutations/submit_application from pg
```
matthieu-foucault, line 1: Where is the revert created by `sqitch rework`? There should be a revert that resets the function to its v1.0.0 behaviour

matthieu-foucault (changes requested):

app/formSchema/pages/submission.ts
marcellmueller: I prefer to keep it in the same order as it appears on the form itself or the `ui:schema`, it's more readable that way.

db/test/unit/mutations/submit_application_test.sql
marcellmueller: I don't see how this is any more helpful or descriptive than what is already there, which already shows in the logs when a test fails.

matthieu-foucault: Fair, I assumed the log was more cryptic without a drescription

app/components/Layout.tsx
```diff
@@ -17,7 +17,7 @@ const StyledLayout = styled('div')`
   min-height: 100vh;
 `;
 const StyledMain = styled('main')`
-  margin: auto 50px;
+  margin: 50px auto;
```
matthieu-foucault, line 20: Why is this changed? This is making pages such as organization location very narrow. If there are overall layout fixes to make, they belong in a separate PR

matthieu-foucault (changes requested):

app/components/Layout.tsx
marcellmueller: oh that was a mistake, I was playing with something because I noticed the layout isn't centered anymore. I guess our happo setup doesn't catch things like that. I'll remove it though.

matthieu-foucault: Happo did catch it, but we have so many spurious diffs that we're blindly approving happo currently

app/lib/theme/widgets/ReadOnlySubmissionWidget.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <StyledValue>
+        {value}
+        {submissionCompletedFor && !value && (
+          <StyledError>
+            No legal organization name was provided. Please return to the{' '}
+            <StyledLink href={`/form/${id}/${pageNumber}`}>
+              Organization Profile
+            </StyledLink>{' '}
+            page and enter one.
+          </StyledError>
```
matthieu-foucault, line 51: This pattern of having field-specific branching behaviour within widgets is at odds with having a declarative uiSchema. If we need specific behaviour for a single field, we should either create a component for it, or find a generic way to have this custom behaviour. 

Can the [custom validation](https://react-jsonschema-form.readthedocs.io/en/docs/usage/validation/#custom-validation-rules) add some JSX as an error?
A generic solution could be a `customError` uiSchema option that would be handled by our field template.

If we do need a custom component, in this case it should be a custom field, not a widget. Widgets should only concern themselves with rendering an input/value.

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -26,6 +26,29 @@ const mockQueryPayload = {
...
+  Query() {
+    return {
+      openIntake: {
+        closeTimestamp: '2022-08-27T12:51:26.69172-04:00',
+      },
+    };
+  },
+};
+
+const submissionPayload = {
```
matthieu-foucault, line 41:
```suggestion
const submissionPayload = {
  ...mockQueryPayload,
```

This way you don't need to copy the `Query()` function every time

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -166,41 +254,25 @@ describe('The application form', () => {
   });
 
   it('submission page submit button is enabled on when all inputs filled', async () => {
-    componentTestingHelper.loadQuery();
+    componentTestingHelper.loadQuery(submissionPayload);
```
matthieu-foucault, line 257: `submissionPayload` doesn't have the `submissionCompletedFor` and `submissionDate`, should they be present for the button to be enabled?

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -166,41 +254,25 @@ describe('The application form', () => {
...
-          screen
-            .getByRole('button', { name: 'Submit' })
-            .hasAttribute('disabled')
-        ).toBeFalse()
-      );
+      expect(
+        screen.getByRole('button', { name: 'Submit' }).hasAttribute('disabled')
+      ).toBeFalse();
     });
   });
```
matthieu-foucault, line 278: Looks like the test below has the wrong title?

matthieu-foucault (changes requested):

app/tests/components/Form/ApplicationForm.test.tsx
marcellmueller: Man you're being way too opinionated. This is getting annoying.

matthieu-foucault: In what way proposing solutions following best practices to reduce code duplication is being opinionated?

marcellmueller: I mean most of this review in general.

matthieu-foucault: Feel free to discuss any feedback on the relevant threads. I'm happy to expand on any of my suggestions. Some of them are probably beyond the scope of this PR and should be tracked in a separate issue

app/formSchema/uiSchema/pages/submission.ts
```diff
@@ -1,14 +1,20 @@
...
     'submissionCompletedBy',
     'submissionTitle',
+    'submissionDate',
   ],
   'ui:title': '',
   'ui:field': 'SubmissionField',
+  submissionCompletedFor: {
+    'ui:widget': 'ReadOnlySubmissionWidget',
+    'ui:options': {
+      'field-name': 'submissionCompletedFor',
```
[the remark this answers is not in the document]
marcellmueller: I made a custom read only widget for submission page because the design has different styling than the other read only widget (no padding, not bold). Rather than make a ton of `ui:options` flags it made sense to make one just for this page as it is the only place in the design that has it.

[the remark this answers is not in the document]
matthieu-foucault: If there's a different behaviour just for this page, then it makes sense to have a custom widget instead of using `readonly` (I'd like to know why our designs aren't consistent across pages though). 
I suppose that rendering the error within the widget makes sense in this case because we're not displaying other errors, just disabling the submit field (which is a separate rabbit hole we won't go into here...).

I still think having behaviour/ui options in the rjsf fields/widgets that is specific to one field is an antipattern and should be avoided. It's unclear when looking at the ui schema that this option will make the `ReadOnlySubmissionWidget` change its behaviour to something that is custom for this specific field. If one field needs a component to display something for that field only, let's make a widget for that field only

app/lib/theme/widgets/ReadOnlySubmissionWidget.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const ReadOnlySubmissionWidget: React.FC<WidgetProps> = ({
+  options,
+  value,
+}) => {
+  const router = useRouter();
+  const { id } = router.query;
+
+  const pageNumber =
+    Object.keys(a.properties).indexOf('organizationProfile') + 1;
```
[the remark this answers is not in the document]
marcellmueller: We should probably have a utility function for this but I am doing it the same way we have been doing it for ages in `ApplicationForm.tsx` here.

app/cypress/integration/index.spec.js
```diff
@@ -468,11 +468,6 @@ context('Homepage', () => {
 
     cy.get('h1').contains('Submission');
 
-    cy.get('input[id="root_submissionCompletedFor"]').type('test');
-
-    cy.get('#root_submissionDate').click();
-    cy.get(`.react-datepicker__day--001`).eq(2).click({ force: true });
-
```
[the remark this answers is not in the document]
marcellmueller: You previously told me that this end to end test was supposed to follow the happy path and didn't need to test every input?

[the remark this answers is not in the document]
matthieu-foucault: I was debating commenting on this one, but I think there's a difference between reducing our test coverage vs adding perfect tests: while changes don't need to have perfect coverage, they shouldn't reduce existing tests coverage

In particular here, since the feature relies on something that happens on another page (inserting the `submissionCompletedFor` and `submissionDate` when the other forms are submitted), I think this is something that would benefit from an end-to-end test.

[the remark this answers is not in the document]
marcellmueller: Okay I’ll add a check for the value tomorrow.

db/sqitch.plan
```diff
@@ -31,5 +31,6 @@ mutations/create_application 2022-08-31T16:16:23Z Anthony Bushara <anthony@butto
 @1.0.0 2022-09-01T23:42:33Z Matthieu Foucault <matthieu@button.is> # release v1.0.0
 @1.0.1 2022-09-02T00:40:17Z Matthieu Foucault <matthieu@button.is> # release v1.0.1
 @1.1.0 2022-09-06T23:59:22Z Matthieu Foucault <matthieu@button.is> # release v1.1.0
-@1.1.1 2022-09-09T20:35:47Z Marcel Mueller,,, <marcel@Thinkpad> # release v1.1.1
+@1.1.1 2022-09-09T20:35:47Z Marcel Mueller <marcel@button.is> # release v1.1.1
```
matthieu-foucault, line 34: This line has been deployed to prod and cannot be updated, otherwise the change will have a different id, leading to the deploy failing.

app/tests/components/Form/ReadOnlySubmissionWidget.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const uiSchema = {
+  submissionDate: {
+    'ui:widget': 'ReadOnlySubmissionWidget',
+  },
+};
+
+const renderStaticLayout = (
+  schema: JSONSchema7,
+  uiSchema: JSONSchema7,
```
matthieu-foucault, line 25:
```suggestion
  uiSchema: any,
```
That, or the uiSchema type exported by rjsf, if there is one. uiSchema isn't a json schema

app/tests/components/Form/ReadOnlySubmissionWidget.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+      onSubmit={() => console.log('test')}
+      schema={schema as JSONSchema7}
+      uiSchema={uiSchema}
+    />
+  );
+};
+
+describe('The ReadOnlySubmissionWidget', () => {
+  it('should render the title', () => {
+    renderStaticLayout(schema as JSONSchema7, uiSchema as JSONSchema7, {});
```
matthieu-foucault, line 40: Instead of a cast here, typing the `schema` variable above would give you better typing. The cast will effectively silence the type checking.

matthieu-foucault (changes requested):
The sqitch change is required, otherwise the deploy will fail.
The test file change are typescript details and quick fixes.

matthieu-foucault (approved):
Great work. Thanks for making the requested changes

PR #638, opened 2022-09-13
ci: use current date for cache

AntBush:
Adds current date for cache in build to avoid trivvy scan pulling in stale artifacts

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -52,11 +52,16 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: current-day
+        uses: josStorer/get-current-time@v2.0.1
+        with:
+          format: YYYY-MM-DD
+          utcOffset: "-7:00"
       - name: Cache Docker layers
         uses: actions/cache@v3
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}
+          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}-${{ steps.current-time.outputs.formattedTime }}
           restore-keys: |
             ${{ runner.os }}-buildx-${{ matrix.name }}
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -52,8 +52,9 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: current-day
+      - name: Get current day
         uses: josStorer/get-current-time@v2.0.1
+        id: current-day
         with:
           format: YYYY-MM-DD
           utcOffset: "-7:00"
@@ -61,7 +62,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}-${{ steps.current-time.outputs.formattedTime }}
+          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}-${{ steps.current-day.outputs.formattedTime }}
           restore-keys: |
             ${{ runner.os }}-buildx-${{ matrix.name }}
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -51,8 +51,8 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Get current day
+      - name: Get current day for cache
+        # To avoid trivvy scans from pulling stale cached images
         uses: josStorer/get-current-time@v2.0.1
         id: current-day
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -52,11 +52,16 @@ jobs:
...
+          format: YYYY-MM-DD
+          utcOffset: "-7:00"
       - name: Cache Docker layers
         uses: actions/cache@v3
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}
+          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}-${{ steps.current-time.outputs.formattedTime }}
           restore-keys: |
             ${{ runner.os }}-buildx-${{ matrix.name }}
```
wenzowski, line 67: Probably need to get this into the restore key too

.github/workflows/build.yaml
```diff
@@ -52,11 +52,16 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: current-day
```
matthieu-foucault:
```suggestion
      - name: current-time
```

.github/workflows/build.yaml
```diff
@@ -52,11 +52,16 @@ jobs:
...
+        uses: josStorer/get-current-time@v2.0.1
+        with:
+          format: YYYY-MM-DD
+          utcOffset: "-7:00"
       - name: Cache Docker layers
         uses: actions/cache@v3
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}
+          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ github.sha }}-${{ steps.current-time.outputs.formattedTime }}
```
wenzowski:
```suggestion
          key: ${{ runner.os }}-buildx-${{ matrix.name }}-${{ steps.current-time.outputs.formattedTime }}-${{ github.sha }}
```
Since we don’t want the sha in the restore key but we do want the date in the restore key we probably need to change the order

wenzowski (changes requested):

.github/workflows/build.yaml
```diff
@@ -52,11 +52,17 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Get current day
```
matthieu-foucault, line 55: It's not obvious why this is needed, can we have comments explaining what's going on?

.github/workflows/build.yaml
```diff
@@ -52,11 +52,17 @@ jobs:
...
+        uses: josStorer/get-current-time@v2.0.1
+        id: current-day
+        with:
+          format: YYYY-MM-DD
```
matthieu-foucault, line 59: do we just want the day, or the time?

wenzowski: We just want the day–we want to bust the cache automatically every day

wenzowski: An alternate approach we were discussing this morning would be to include the base branch’s sha (head for the branch) in the key so we don’t end up with stale docker layers in long lived PRs and every merge to main busts the cache for every long-lived PR.

.github/workflows/build.yaml
wenzowski: Without the date in the restore key it will continue to match as it has been doing and we will not get the effect of a cache-bust every day

.github/workflows/build.yaml
matthieu-foucault: What do you mean by "bust"? Do you want to rebuild without cache daily, or write to the cache daily?

matthieu-foucault: As an alternate alternate approach, we could delete all of the repo's cache on a nightly basis. This is now possible via the API

wenzowski: We want to ignore the cache if the cache was created before today. A nightly and/or manually triggered workflow to actually delete the cache would be a better option. Stale cache currently blocking another PR so I'd lean towards getting both a quick simple fix in now and a plan to get a well thought out fix going in parallel.

wenzowski: ^ @VB-at-Bis

VB-at-Bis: hmm.. https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key looks promising.. let me see if I can find key etc

AntBush: So do we want to continue with this PR or close it in favour of the deletes? I'm leaning to close

VB-at-Bis:
After discussion with the team we found a manual way to clear the cache:
https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key 
Steps to use it:
- find cached image key in GitHub action logs (i.e. see https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3046256745/jobs/4910884869 , select `build/app` job, expand `Cache docker layers` and find line like follows `Cache restored from key: Linux-buildx-app-5bb8f4306193b3abfe09da9d2d94cbf33b7a613f`;
- use personal FitHub token and image key to prepare curl statement like follows: 
`curl -X DELETE -H "Accept: application/vnd.github+json" -H "Authorization: Bearer XXXXXX" https://api.github.com/repos/bcgov/CONN-CCBC-portal/actions/caches?key=Linux-buildx-app-5bb8f4306193b3abfe09da9d2d94cbf33b7a613f`
- run curl statement (if successful, should produce something like below:
`{
  "total_count": 1,
  "actions_caches": [
    {
      "id": 2740,
      "ref": "refs/pull/633/merge",
      "key": "Linux-buildx-app-5bb8f4306193b3abfe09da9d2d94cbf33b7a613f",
      "version": "241a22ad23e37799b851ed553e697ea8915dfceac047c09bcb4c846ed1d09e11",
      "last_accessed_at": "2022-09-13T18:00:40.003333300Z",
      "created_at": "2022-09-13T15:19:42.040000000Z",
      "size_in_bytes": 481268869
    }
  ]
}`

.github/workflows/build.yaml
VB-at-Bis: I added comment to the ticket with steps how to use GitHub API to clear the cached image.

AntBush:
I'm going to close this PR since we have a way to manually delete the cache now if this problem persists

PR #652, opened 2022-09-14
ci: add a manual workflow to clear github actions cache

wenzowski:
looks like #638 got closed without a devops resolution; this pr adds the manual fix documented in the closed pr as a manually triggered pipeline workflow

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,6 +11,10 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
+  clear-cache:
+    if: github.event_name == 'schedule'
+    uses: ./.github/workflows/cache.yaml
+
   build-and-push-image:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,10 +11,6 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
-  clear-cache:
-    if: github.event_name == 'schedule'
-    uses: ./.github/workflows/cache.yaml
-
   build-and-push-image:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/cache.yaml
+++ b/.github/workflows/cache.yaml
@@ -1,0 +1,32 @@
+name: cache
+on:
+  workflow_call:
+  workflow_dispatch:
+concurrency:
+  group: cache-${{ github.ref }}
+  cancel-in-progress: true
+permissions:
+  actions: write
+jobs:
+  clear-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clear cache
+        uses: actions/github-script@v6
+        with:
+          script: |
+            console.log("Preparing to clear cache...")
+            const caches = await github.rest.actions.getActionsCacheList({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+            })
+            for (const cache of caches.data.actions_caches) {
+              console.log("Clearing cache_id", cache.id)
+              await github.rest.actions.deleteActionsCacheById({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                cache_id: cache.id,
+              })
+              console.log("Cleared cache_id", cache.id)
+            }
+            console.log("Cleared cache.")
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -2,8 +2,7 @@
 name: main
 
 on:
-  schedule:
-    - cron: '0 2 * * *'
+  workflow_call:
   push:
     branches: [main]
   pull_request:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -2,7 +2,7 @@
 name: main
 
 on:
-  workflow_call:
+  workflow_call: # be sure to use 'secrets: inherit' in the caller
   push:
     branches: [main]
   pull_request:
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,11 @@
+# Schedule to trigger Main workflow with cleared cache
+name: nightly
+on:
+  schedule:
+    - cron: '0 2 * * *'
+jobs:
+  clear-cache:
+    uses: ./.github/workflows/cache.yaml
+  main:
+    needs: [clear-cache]
+    uses: ./.github/workflows/main.yaml
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -2,10 +2,11 @@
 name: nightly
 on:
   schedule:
-    - cron: '0 2 * * *'
+    - cron: '0 5 * * *'
 jobs:
   clear-cache:
     uses: ./.github/workflows/cache.yaml
   main:
     needs: [clear-cache]
     uses: ./.github/workflows/main.yaml
+    secrets: inherit


.github/workflows/cache.yaml
```diff
@@ -0,0 +1,30 @@
+name: cache
+on: workflow_call
+concurrency:
+  group: cache-${{ github.ref }}
+  cancel-in-progress: true
```
wenzowski, line 5: @matthieu-foucault even in the context of moving the concurrency group up to `main.yaml` this workflow will always be manually called so I figured we'd leave this here?

wenzowski: ...or maybe not since I'm proposing we trigger this on schedule with nightlies. If we land the concurrency refactor first I'll address?

.github/workflows/build.yaml
```diff
@@ -11,6 +11,10 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
+  clear-cache:
+    if: github.event_name == 'schedule'
+    uses: ./.github/workflows/cache.yaml
+
```
wenzowski, line 17: I guess as a separate job this is going to run in parallel with the build job. Perhaps it should be a step that always runs before build if it should be triggered so the build is clean when it should be.

matthieu-foucault: The cache needs to be cleared before building the asdf plugins for tests as well. Maybe another `nightly.yaml` workflow that is clearing cache and then triggering the `main.yaml` workflow would be better. With a separate workflow you wouldn't need the `if: github.event_name == 'schedule'`

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,11 @@
...
+name: nightly
+on:
+  schedule:
+    - cron: '0 2 * * *'
+jobs:
+  clear-cache:
+    uses: ./.github/workflows/cache.yaml
+  main:
+    needs: [clear-cache]
+    uses: ./.github/workflows/main.yaml
```
wenzowski, line 11: @matthieu-foucault something like this?

.github/workflows/main.yaml
```diff
@@ -2,8 +2,6 @@
 name: main
 
 on:
```
matthieu-foucault, line 4:
```suggestion
on:
  workflow_call:
    secrets:
      SONAR_TOKEN: { required: true }
      HAPPO_API_KEY: { required: true }
      HAPPO_API_SECRET: { required: true }
      OPENSHIFT_SERVER: { required: true }
      OPENSHIFT_TOKEN: { required: true }
      OPENSHIFT_APP_NAMESPACE: { required: true }
      OPENSHIFT_METABASE_NAMESPACE: { required: true }
      NEXT_PUBLIC_GROWTHBOOK_API_KEY: { required: true }
      CLIENT_SECRET: { required: true }
      OPENSHIFT_SECURE_ROUTE: { required: true }
      AWS_S3_BUCKET: { required: true }
      AWS_S3_REGION: { required: true }
      AWS_S3_KEY: { required: true }
      AWS_S3_SECRET_KEY: { required: true }
      AWS_ROLE_ARN: { required: true }
      CERTBOT_EMAIL: { required: true }
      CERTBOT_SERVER: { required: true }
```

You need this to able to call the workflow from `nightly`, and there you need to explicitly pass secrets, like we're doing below

.github/workflows/cache.yaml
matthieu-foucault: Not sure we should worry about this workflow's concurrency. We can probably leave this out?

.github/workflows/nightly.yaml
matthieu-foucault: something like this, with secrets (see my other comment)

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,11 @@
+# Schedule to trigger Main workflow with cleared cache
+name: nightly
+on:
+  schedule:
+    - cron: '0 2 * * *'
```
matthieu-foucault, line 5:
```suggestion
    - cron: '0 5 * * *'
```
Maybe a bit later? 2am UTC is 6pm pacific in winter, which could be annoying if someone works late. We don't want to add more jobs than necessary to our limited concurrency pool during work hours

matthieu-foucault (changes requested):

.github/workflows/cache.yaml
wenzowski: Conceivably could take a while to iterate through the loop if it was a busy day with a large cache and I didn't want a race condition where new job is also iterating in parallel

.github/workflows/nightly.yaml
wenzowski: agreed

.github/workflows/build.yaml
wenzowski: resolved by splitting `nightly.yaml`

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,29 @@
...
+  schedule:
+    - cron: '0 5 * * *'
+jobs:
+  clear-cache:
+    uses: ./.github/workflows/cache.yaml
+  main:
+    needs: [clear-cache]
+    uses: ./.github/workflows/main.yaml
+    secrets:
+      SONAR_TOKEN: { required: true }
```
matthieu-foucault, line 13:
```suggestion
      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
... and so on. This is annoyingly verbose but reusable workflow aren't necessarily in the same repo, so it make sense that we have to be explicit

AntBush: Actually we can use 
`secrets: inherit` since they're in the same repo now/relatively recently:

https://github.blog/changelog/2022-05-03-github-actions-simplify-using-secrets-with-reusable-workflows/

AntBush: I think inheritance is allowed so long as they're under the same org?

matthieu-foucault: Great find, let's do this! That's a lot of yaml we can delete then

matthieu-foucault (approved):

.github/workflows/cache.yaml
```diff
@@ -0,0 +1,32 @@
...
+            for (const cache of caches.data.actions_caches) {
+              console.log("Clearing cache_id", cache.id)
+              await github.rest.actions.deleteActionsCacheById({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                cache_id: cache.id,
+              })
+              console.log("Cleared cache_id", cache.id)
+            }
+            console.log("Cleared cache.")
```
matthieu-foucault, line 32: This looks good, but how can be we confident that this works as expected before merging?

wenzowski: Unfortunately, workflows can’t be manually triggered until they’re in the main branch. Perhaps a new testing draft pr based on this one where I commit a push trigger to a workflow…

matthieu-foucault: That's what I was thinking and suggested to @AntBush  for #592

wenzowski: Runs successfully and deletes all cache layers
https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3062028759/jobs/4942521697

matthieu-foucault: beautiful. We should note this as a way to test GH workflows

matthieu-foucault (approved):

PR #654, opened 2022-09-14
fix: update date for banner

AntBush:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/648


matthieu-foucault (approved):
Looking good

PR #656, opened 2022-09-14
feat: add save as draft button to submission page

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/598

### DIFF
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -2,16 +2,15 @@ import { useMemo, useState } from 'react';
 import { useRouter } from 'next/router';
 import { IChangeEvent, ISubmitEvent } from '@rjsf/core';
 import { graphql, useFragment } from 'react-relay';
-import Button from '@button-inc/bcgov-theme/Button';
 import type { JSONSchema7 } from 'json-schema';
-import { CalculationForm, FormBase } from '.';
+import styled from 'styled-components';
+import { CalculationForm, FormBase, SubmitBtns } from '.';
 import uiSchema from '../../formSchema/uiSchema/uiSchema';
 import schema from '../../formSchema/schema';
 import { schemaToSubschemasArray } from '../../utils/schemaUtils';
 import { Review } from '../Review';
 import { acknowledgements } from '../../formSchema/pages';
 import ApplicationFormStatus from './ApplicationFormStatus';
-import styled from 'styled-components';
 
 // https://github.com/rjsf-team/react-jsonschema-form/issues/2131
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
@@ -181,6 +180,8 @@ const ApplicationForm: React.FC<Props> = ({
 
   const [reviewConfirm, setReviewConfirm] = useState(false);
   const [savingError, setSavingError] = useState(null);
+  const [saveAsDraft, setSaveAsDraft] = useState(true);
+
   const [areAllAcknowledgementsChecked, setAreAllacknowledgementsChecked] =
     useState(verifyAllAcknowledgementsChecked(formData['acknowledgements']));
   const [areAllSubmissionFieldsSet, setAreAllSubmissionFieldsSet] = useState(
@@ -319,6 +320,7 @@ const ApplicationForm: React.FC<Props> = ({
   };
 
   const handleChange = (e: IChangeEvent<any>) => {
+    setSaveAsDraft(true);
     saveForm(e.formData);
   };
 
@@ -376,25 +378,16 @@ const ApplicationForm: React.FC<Props> = ({
   const isCustomPage = customPages.includes(sectionName);
   const isSubmitPage = pageNumber >= subschemaArray.length;
 
-  const formatSubmitBtn = () => {
-    if (isWithdrawn) {
-      return 'Continue';
-    }
-    if (!isSubmitPage) {
-      return 'Save and continue';
-    }
-    return 'Submit';
-  };
-
   const submitBtns = (
-    <>
-      <Button
-        variant="primary"
-        disabled={handleDisabled(sectionName, noErrors) || isSubmitting}
-      >
-        {formatSubmitBtn()}
-      </Button>
-    </>
+    <SubmitBtns
+      isWithdrawn={isWithdrawn}
+      isSubmitPage={isSubmitPage}
+      formData={formData}
+      saveAsDraft={saveAsDraft}
+      saveForm={saveForm}
+      setSaveAsDraft={() => setSaveAsDraft(false)}
+      disabled={handleDisabled(sectionName, noErrors) || isSubmitting}
+    />
   );
 
   const isDisabled = () => {
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -180,7 +180,7 @@ const ApplicationForm: React.FC<Props> = ({
 
   const [reviewConfirm, setReviewConfirm] = useState(false);
   const [savingError, setSavingError] = useState(null);
-  const [saveAsDraft, setSaveAsDraft] = useState(true);
+  const [savedAsDraft, setSavedAsDraft] = useState(false);
 
   const [areAllAcknowledgementsChecked, setAreAllacknowledgementsChecked] =
     useState(verifyAllAcknowledgementsChecked(formData['acknowledgements']));
@@ -294,7 +294,7 @@ const ApplicationForm: React.FC<Props> = ({
       },
       onCompleted: () => {
         if (isSaveAsDraftBtn) {
-          setSaveAsDraft(false);
+          setSavedAsDraft(true);
         }
       },
       ...mutationConfig,
@@ -326,7 +326,7 @@ const ApplicationForm: React.FC<Props> = ({
   };
 
   const handleChange = (e: IChangeEvent<any>) => {
-    setSaveAsDraft(true);
+    setSavedAsDraft(false);
     saveForm(e.formData);
   };
 
@@ -390,7 +390,7 @@ const ApplicationForm: React.FC<Props> = ({
       isUpdating={isUpdating}
       isSubmitPage={isSubmitPage}
       formData={formData}
-      saveAsDraft={saveAsDraft}
+      savedAsDraft={savedAsDraft}
       saveForm={saveForm}
       status={status}
     />
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -211,7 +211,8 @@ const ApplicationForm: React.FC<Props> = ({
     mutationConfig?: Partial<
       UseDebouncedMutationConfig<updateApplicationMutation>
     >,
-    isRedirectingToNextPage = false
+    isRedirectingToNextPage = false,
+    isSaveAsDraftBtn = false
   ) => {
     if (isWithdrawn) {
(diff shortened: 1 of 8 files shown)


app/components/Form/SubmitBtns.tsx
```diff
@@ -0,0 +1,108 @@
...
+  text-align: center;
+
+  & a {
+    color: #ffffff;
+  }
+`;
+
+type Props = {
+  disabled: boolean;
+  formData: any;
```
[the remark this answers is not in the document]
AntBush: Can probably dismiss this alert since (afaik) there's no easy way to get the type of the formData from the schema. Would be nice though.

app/components/Form/SubmitBtns.tsx
```diff
@@ -0,0 +1,108 @@
...
+  isSubmitPage: boolean;
+  isUpdating: boolean;
+  saveAsDraft: boolean;
+  saveForm: any;
```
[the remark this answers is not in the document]
AntBush: Should be typed as a Function, or better yet with a function type expression

app/components/Form/SubmitBtns.tsx
```diff
@@ -0,0 +1,108 @@
...
+  status: string;
+};
+
+const SubmitBtns = ({
```
AntBush, line 49: To remain consistent let's change this to `SubmitButtons` rather than the shorthand

app/components/Form/SubmitBtns.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <StyledFlex>
+      <Button variant="primary" disabled={disabled}>
+        {formatSubmitBtn()}
+      </Button>
+      {isSubmitPage && isDraft && (
+        <>
+          <StyledButton
+            variant="secondary"
+            onClick={(e: React.MouseEvent<HTMLInputElement>) => {
```
AntBush, line 80: Thoughts on passing a `handleSubmit` function rather than just the saveForm function?

app/components/Form/SubmitBtns.tsx
```diff
@@ -0,0 +1,108 @@
...
+  saveForm: any;
+  status: string;
+};
+
+const SubmitBtns = ({
+  disabled,
+  formData,
+  isSubmitPage,
+  isUpdating,
+  saveAsDraft,
```
AntBush, line 54: I read this as "should we save as draft", but from what I'm reading it's closer to "has the save as draft completed". Thoughts on changing the name?

app/components/Form/SubmitBtns.tsx
```diff
@@ -0,0 +1,108 @@
...
+            {isUpdating ? (
+              <LoadingSpinner />
+            ) : (
+              <>{saveAsDraft ? 'Save as draft' : 'Saved'}</>
+            )}
+          </StyledButton>
+          <StyledToast
+            style={{
+              visibility: saveAsDraft ? 'hidden' : 'visible',
+              opacity: saveAsDraft ? 0 : 1,
```
AntBush, line 96: I'd still consider myself a css n00b, why both visibility and opacity for this? Also any reason we are going through CSS route vs a conditional render?

app/styles/GlobalTheme.tsx
```diff
@@ -22,6 +22,7 @@ const theme = {
     success: '#2E8540',
     text: '#313132',
     descriptionGrey: 'rgba(49, 49, 50, 0.7)',
+    successGreen: '#2e8540',
```
AntBush, line 25: Much appreciated!

app/tests/components/Form/SubmitBtns.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+      ...defaultProps,
+      isSubmitPage: true,
+    };
+
+    const { rerender } = renderStaticLayout(props);
+
+    await userEvent.click(
+      screen.getByRole('button', { name: 'Save as draft' })
+    );
+
```
AntBush, line 170: Great tests! 🧪

app/components/Form/SubmitBtns.tsx
```diff
@@ -0,0 +1,108 @@
...
+              e.preventDefault();
+              saveForm(formData, {}, false, true);
+            }}
+            disabled={isUpdating || !saveAsDraft}
+            style={{ padding: isUpdating ? '4px 24px' : '12px 24px' }}
+          >
+            {isUpdating ? (
+              <LoadingSpinner />
+            ) : (
+              <>{saveAsDraft ? 'Save as draft' : 'Saved'}</>
```
AntBush, line 90: This is a nitpick, but I don't think we need the fragment here. It does look cleaner this way though vs when it's run through prettier so I don't really care about it.

I'm surprised sonarcloud didn't complain about this though, usually it doesn't like nested ternary operators. Wouldn't be surprised if this is circumventing it's check

AntBush (changes requested):
Great stuff! Just a couple of small changes

app/components/Form/SubmitBtns.tsx
marcellmueller: Yeah I considered that, I can definitely change that.

app/components/Form/SubmitBtns.tsx
marcellmueller: I believe there is a few reasons for this.  We are using ``e.preventDefault()` because RJSF will use any child button as a submit button if not. The `handleSubmit` function also will change the form page once completed.   
  
Also you can see I added an argument `isSaveAsDraftBtn` to the `saveForm` function which was needed to set the state for the save button once the mutation had completed.

app/components/Form/SubmitBtns.tsx
marcellmueller: yeah this one I was stuck on the naming for a bit. Happy to change it but `hasTheSaveAsDraftCompleted` seems a bit verbose. I was just looking at it as saveAsDraft is true, if not then we use saved....not ideal but I'm happy to change it if you have a good name.

app/components/Form/SubmitBtns.tsx
marcellmueller: Without the visibility as hidden if you mouse over the component with 0 opacity the cursor will still change to the select/highlight cursor and allow you to select the text.  

An alternative could be to use `display` instead. I tried it with conditional rendering but then the transition wouldn't work.

app/components/Form/SubmitBtns.tsx
marcellmueller: The fragment is required..React yells at me if I don't have a node in a conditional. Also I don't believe this is a nested ternary in that sense, it's completely separate from the parent one.

AntBush: Weird didn't expect it to yell, especially since a string is technically a valid React Node. I'm fine with this though since it looks cleaner this way.

app/components/Form/ApplicationForm.tsx
```diff
@@ -176,6 +176,8 @@ const ApplicationForm: React.FC<Props> = ({
 
   const [reviewConfirm, setReviewConfirm] = useState(false);
   const [savingError, setSavingError] = useState(null);
+  const [saveAsDraft, setSaveAsDraft] = useState(true);
```
matthieu-foucault, line 179: It took me a few passes to understand what this state does.
`save` is an action, so one could expect `saveAsDraft` to be a function, not a boolean.
Should this be `canSaveAsDraft` ? Or, alternatively, setting it to `false` by default and rename it to `savedAsDraft`?
A comment indicating that this is only for the submission page could be useful, as we're saving drafts on edit.

app/components/Form/SubmitBtns.tsx
AntBush: Ah sorry I meant moreso to send a function that fully handles the `onClick` in this case, but this is more a matter of my own style.

app/components/Form/SubmitButtons.tsx
```diff
@@ -0,0 +1,108 @@
...
+  saveForm: any;
+  status: string;
+};
+
+const SubmitButtons = ({
+  disabled,
+  formData,
+  isSubmitPage,
+  isUpdating,
+  saveAsDraft,
```
matthieu-foucault, line 54: Same comment as for the `ApplicationForm` state. Here the name is even more problematic, because of the `saveForm` below, which is a function.

app/components/Form/SubmitButtons.tsx
```diff
@@ -0,0 +1,108 @@
...
+  saveForm,
```
matthieu-foucault, line 55: Thoughts on calling this `onSubmit`? This would be in line with idiomatic variable naming conventions.

app/components/Form/SubmitBtns.tsx
AntBush: I'm good with a long name if you are, so long as it's clear! Could also add a js-doc comment around the prop/variable where it's declared but that's not the style of the rest of the app.

app/components/Form/ApplicationForm.tsx
```diff
@@ -204,7 +206,8 @@ const ApplicationForm: React.FC<Props> = ({
     mutationConfig?: Partial<
       UseDebouncedMutationConfig<updateApplicationMutation>
     >,
-    isRedirectingToNextPage = false
+    isRedirectingToNextPage = false,
+    isSaveAsDraftBtn = false
```
matthieu-foucault, line 210: When we have several optional parameters like this, its worth considering having a single `options` parameter instead. This will make the calling code easier to read.

matthieu-foucault (commented):
Mostly naming convention comments, the actual behaviour look good.

app/components/Form/SubmitButtons.tsx
matthieu-foucault: `onSave`, or `onClick` would be valid options too. Props that are functions called on an action tend to be called `onSomething`.

If `onSubmit` is confusing, is the name of the component also confusing? Should it be `SaveButtons`?

Going further, if we are to adopt idiomatic naming conventions, the `saveForm` function would be better defined as `handleSave`, or `handleSaveForm` (but maybe that's a change for another PR)

app/components/Form/SubmitButtons.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+`;
+
+type Props = {
+  disabled: boolean;
+  formData: any;
+  isSubmitPage: boolean;
+  isUpdating: boolean;
+  saveAsDraft: boolean;
+  saveForm: any;
```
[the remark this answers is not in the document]
marcellmueller: I thought Mattieu said using any is fine?

app/components/Form/SubmitButtons.tsx
marcellmueller: `SubmitButtons` is fine since the primary use for this is to submit the form. This feature is on one page and makes sense to have in the same component. `onSave` or `onClick` is confusing as well because it is not clear that what it is doing, remember that we have the main submit button for the entire form in this component which is controlled by the `handleSubmit` in the parent.

app/components/Form/SubmitButtons.tsx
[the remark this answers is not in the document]
marcellmueller: I committed the suggested change though personally I don't care to type things like this.

AntBush (approved):
Looks good!

Issue #661, opened 2022-09-13
Enable time travel in dev and test namespaces


VB-at-Bis:
### Description

The time travel feature was added, but is only enabled if the appropriate env variable is set (`ENABLE_MOCK_TIME`).

The `config.js` file has some mechanism to enable it, but it incorrectly relies on the `NODE_ENV` env variable. `NODE_ENV` is always `production` when running the docker container.

The three `json` config files are at in the `app` folder and would be more visible if moved to a `config` folder

### AC

- [x] create an `app/config` folder, containing
  - `index.js`, moved+renamed from `app/config.js`
  - `local.json`, moved+renamed from `app/development.json`
  - `dev.json`, `test.json` and `prod.json`, containing env variables for the three openshift environments, respectively
- [x] update the `app/config/index.js` behaviour to:
  - load `local.json` if `!NODE_ENV || NODE_ENV === 'development'`
  - load the appropriate json file otherwise, based on `OPENSHIFT_APP_NAMESPACE` (defaulting to prod)

```
const tokens = config.get("OPENSHIFT_APP_NAMESPACE").split('-');
const env = tokens[tokens.length - 1];
```

- [x] The `dev.json`, `test.json` and `prod.json` should only contain the necessary env variables. Anything defined in the helm chart (https://github.com/bcgov/CONN-CCBC-portal/blob/49ea4c04486abc4bb6905bb98d0d009cc95f18ad/helm/app/templates/deployment.yaml#L148-L188) will be overridden anyway, so it shouldn't be in those json files. `prod.json` might be an empty json object as a result, and that's okay.
- [x] The `dev.json`, `test.json` and `prod.json` should not attempt to set `NODE_ENV`

VB-at-Bis:
@matthieu-foucault  I saw it in Makefile: https://github.com/bcgov/CONN-CCBC-portal/blob/main/Makefile#L118 and assumed that's enough...
- you mean add `sqitch --chdir mocks_schema deploy` in  helm\app\templates\deployment.yaml after line 142?

VB-at-Bis:
PR submitted: https://github.com/bcgov/CONN-CCBC-portal/pull/678 @matthieu-foucault

matthieu-foucault:
The problem is that our deployment script isn't using the Makefile, that should probably be refactored at some point

PR #665, opened 2022-09-16
661 config changes

VB-at-Bis:
### Description: 

Changed logic how environment-specific configuration file is loaded.

References #661 

### Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist for DoD:
- [x] Tests written and passing (e2e, db, and/or unit where applicable)
- [x] All acceptance criteria met
- [ ] Peer Reviewed


MarsRomer:
do we still need to do #661  to enable use of timetravel in test?

VB-at-Bis:
Yes, we need it

app/backend/create-server.ts
```diff
@@ -1,7 +1,7 @@
 import Debug from 'debug';
 import { createServer as _createServer } from 'http';
 import dotenv from 'dotenv';
-import config from '../config.js';
+import config from '../config/index.js';
```
matthieu-foucault, line 4:
```suggestion
import config from '../config;
```
Does this work?

app/config/dev.json
```diff
@@ -8,18 +7,9 @@
...
   "SESSION_SECRET": "superSecretSession",
-  "NEXT_PUBLIC_GROWTHBOOK_API_KEY": "",
-  "AWS_S3_BUCKET": "",
-  "AWS_S3_REGION": "us-west-2",
-  "AWS_S3_KEY": "",
-  "AWS_S3_SECRET_KEY": "",
-  "AWS_ROLE_ARN": "",
   "SITEMINDER_LOGOUT_URL": "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi",
   "ENABLE_MOCK_TIME" : true,
   "ENABLE_MOCK_COOKIES" : true
```
matthieu-foucault, line 15: Do we need anything else than those three variables? The rest is set by the environment variables in the helm chart

app/config/index.js
```diff
@@ -159,10 +159,14 @@ const config = convict({
...
+const env = config.get('NODE_ENV') !== 'development' 
+  ? chunks[chunks.length-1] || 'local' 
+  : 'local';
 
 try {
-  config.loadFile('./' + env + '.json');
+  config.loadFile('./config/' + env + '.json');
 }
 catch(e){
   console.log(e);
```
matthieu-foucault, line 172: Noticing that this will start the app anyway if we don't find the json file. Is this what we want?

app/config/production.json
```diff
@@ -0,0 +1,4 @@
+{
+  "OPENSHIFT_APP_NAMESPACE": "ff61fb-prod",
+  "SITEMINDER_LOGOUT_URL": "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi"
```
matthieu-foucault, line 3: Should there be a different logout url for prod?

app/config/production.json
```diff
@@ -0,0 +1,4 @@
+{
+  "OPENSHIFT_APP_NAMESPACE": "ff61fb-prod",
```
matthieu-foucault, line 2: Don't need to define this as it will be overridden by an env variable

app/config/test.json
```diff
@@ -8,16 +7,7 @@
...
-  "NEXT_PUBLIC_GROWTHBOOK_API_KEY": "",
-  "AWS_S3_BUCKET": "",
-  "AWS_S3_REGION": "us-west-2",
-  "AWS_S3_KEY": "",
-  "AWS_S3_SECRET_KEY": "",
-  "AWS_ROLE_ARN": "",
+  "OPENSHIFT_APP_NAMESPACE": "ff61fb-test",
   "SITEMINDER_LOGOUT_URL": "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi",
   "ENABLE_MOCK_TIME" : true,
   "ENABLE_MOCK_COOKIES" : true
```
matthieu-foucault, line 13: same comment as for `dev.json`, not sure we need anything else than those three.

app/components/SessionExpiryHandler.tsx
```diff
@@ -3,7 +3,7 @@ import { useRouter } from 'next/router';
 import type { SessionExpiryHandlerQuery } from '../__generated__/SessionExpiryHandlerQuery.graphql';
 import { graphql, fetchQuery, useRelayEnvironment } from 'react-relay';
 import { SessionTimeoutHandler } from '@bcgov-cas/sso-react';
-import config from '../config.js';
+import config from '../config/index.js';
```
matthieu-foucault, line 6: Looks like we have a fix that was undone at some point. This front-end file should not load the `config`. It should use next.js' [`publicRuntimeConfig`](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) instead

app/config/local.json
```diff
@@ -0,0 +1,16 @@
...
+  "PGPASSWORD": "",
+  "PGHOST": "localhost",
+  "PGPORT": 5432,
+  "PGDATABASE": "ccbc",
+  "PGSCHEMA": "ccbc_public",
+  "HOST": "localhost",
+  "OPENSHIFT_APP_NAMESPACE": "ff61fb-dev",
+  "SESSION_SECRET": "superSecretSession",
+  "SITEMINDER_LOGOUT_URL": "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi",
+  "ENABLE_MOCK_TIME" : true,
```
matthieu-foucault, line 14: Is this what is leading to happo displaying the time travel components? Should we set it to `false`, so that happo displays screenshots that match the production site?
`ENABLE_MOCK_COOKIES` below should remain `true` to allow cypress to mock the current date

matthieu-foucault (changes requested):

app/backend/create-server.ts
VB-at-Bis: yes, when it's folder then index.js is default
https://nodejs.org/api/modules.html : 
```require('./some-library') would attempt to load:

./some-library/index.js
./some-library/index.node```

app/config/index.js
VB-at-Bis: Yes, that was original code

app/config/production.json
VB-at-Bis: Good point. Is it https://logon.gov.bc.ca/clp-cgi/logoff.cgi ?

app/config/local.json
VB-at-Bis: Done

app/config/test.json
VB-at-Bis: Done

app/config/production.json
VB-at-Bis: Removed

app/config/local.json
VB-at-Bis: Now happo screenshots do not have Time Travel

matthieu-foucault (approved):
Created #676 to fix the SessionExpiryHandler config behaviour

PR #666, opened 2022-09-16
feat/477-return-to-dashboard

AntBush:
Handles the return to dashboard AC using the change in SubmitBtns from Marcel's PR


app/components/Form/SubmitButtons.tsx
```diff
@@ -82,7 +92,7 @@ const SubmitButtons = ({
       <Button variant="primary" disabled={disabled}>
         {formatSubmitBtn()}
       </Button>
-      {isSubmitPage && isDraft && (
+      {isDraftAndSubmitPage && (
         <>
```
marcellmueller, line 96: I think we can get rid of these redundant fragments surrounding this button other than that this is looks good to me!

marcellmueller (approved):
Just one tiny thing, other than that this can get merged.

Edit: n/m this is good as is.

PR #672, opened 2022-09-19
Add typescript types for JSON schemas

matthieu-foucault:
Taking some changes out of #547

This PR only adds types to all of our json schemas. For some forms, we have to cast the object as we use `enumNames`, which are an `rjsf` feature and not part of the json-schema spec.


app/package.json
```diff
@@ -35,6 +35,7 @@
     "express-session": "^1.17.2",
     "graphql-upload": "^15.0.2",
     "js-cookie": "^3.0.1",
+    "json-schema": "^0.4.0",
```
AntBush, line 38: Should we be using from the [types](https://www.npmjs.com/package/@types/json-schema) package or does it not really matter?

AntBush (commented):
Other than the one comment I don't see anything that requires changes.

app/package.json
AntBush: Ah gotcha, cool! I did check that the `@types /json-schema` supports draft-07 since I did notice that the summary randomly cuts off before listing anything after 6 and it is in the files for the package.

AntBush (approved):
👍

PR #673, opened 2022-09-19
UISchemas support themes + simplify customTitles

matthieu-foucault:
Another set of changes taken out of #547 

This PR:
- changes the UI schemas to use strings for widgets, allowing to change the theme without needing to update the UISchema widgets
- adds supports for multiple themes to the Form component
- refactors custom titles to be a JSX element instead of a function, since they are static elements. This removes some typescript workarounds


marcellmueller (approved):

app/components/Form/FormBase.tsx
```diff
@@ -1,23 +1,25 @@
...
-  formData: any;
+  theme?: any;
 }
 
-const Form = withTheme(formTheme);
 const FormBase: React.ForwardRefRenderFunction<
   object,
   FormPropsWithTheme<object>
 > = (props, ref) => {
+  const Form = useMemo(
```
marcellmueller, line 18: Should we update `CalculationForm` with this?

matthieu-foucault: `CalculationForm` is defined as a class component, so it can't use hooks. The `Form` there is defined in the constructor, which is fine.
I'd like it for this `FormBase` component to support calculations one day, in a declarative way that doesn't require branching in the `ApplicationForm` component

Issue #471, opened 2022-08-22
json schema should be stored in the database and linked to the form data


matthieu-foucault:
Required for the RFI form enablement.

Benefit:  Good back end validation so that data is coherent.
Risk:  acknowledgements could be bypassed

![image.png](https://images.zenhubusercontent.com/575f1a78bf93871ccd9c8bcd/7ebafa5a-e9c2-4979-bc7f-ff07a5194348)

- [ ]  Update the application JSON schema to be a JSON object
- [x] Create a `form` table that contains a single record, with an `intake1schema` slug. The record must be inserted after deploying the sqitch changes, in an idempotent way, allowing for non-breaking changes in prod (see [ciip](https://github.com/bcgov/cas-ciip-portal/blob/develop/schema/data/prod/form_json.sql)]. If we want to keep our json schema in multiple files, we can consider having a server-side function writing it to the DB (if needed) at boot time.
- [x] The `ApplicationForm` component retrieves the json schema from the db
- [ ] The `submit_application` mutation takes the total number of acknowledgements from the json schema rather than the constant
- [x] The `application_form_data` computed column filters `form_data` records based on the `form_id`

eazeredo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/471?planning-poker) with Zenhub @AntBush @marcellmueller @matthieu-foucault @wenzowski

matthieu-foucault:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/471?planning-poker) with Zenhub @VB-at-Bis @wenzowski

marcellmueller:
@MarsRomer I seem to have issues with it in test though it's working great on my local machine. Sorry about this.

PR #675, opened 2022-09-20
feat: backend validation for acknowledgements

AntBush:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/477

Adds the backend enforcement for acknowledgements

### DIFF
--- a/db/deploy/mutations/submit_application.sql
+++ b/db/deploy/mutations/submit_application.sql
@@ -14,6 +14,7 @@ declare
   submission_completed_by varchar;
   submission_title varchar;
   submission_date varchar;
+  acknowledgements_array_length integer;
 begin
 
   select ccbc_public.application_status(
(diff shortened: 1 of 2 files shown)


db/deploy/mutations/submit_application.sql
```diff
@@ -51,6 +53,10 @@ begin
     raise 'The application cannot be submitted as the submission field submission_date is null or empty';
   end if;
 
+  if acknowledgements_array_length <> 17 or acknowledgements_array_length is null then
```
matthieu-foucault, line 56: magic number `17` :magic_wand: 

Let's add an acceptance criteria to #471 to refactor this so that we get the number of acknowledgements from the schema instead of hardcoding it here

marcellmueller:
I tested this and it's working great. Maybe we should open a ticket about some front end error message on the submission page if this isn't filled? Right now skipping the acknowledgements page is somewhat of an edge case but if the stepper gets built it could happen a lot more easily.

AntBush:
> I tested this and it's working great. Maybe we should open a ticket about some front end error message on the submission page if this isn't filled? Right now skipping the acknowledgements page is somewhat of an edge case but if the stepper gets built it could happen a lot more easily.

Agreed, it silently failing will definitely only be confusing for the user. We will need to consult with Zoe/Elliot about how we wan to handle these errors (display them/redirect to page with errors/should have disabled submit button while there are validation errors/etc)

matthieu-foucault:
> > I tested this and it's working great. Maybe we should open a ticket about some front end error message on the submission page if this isn't filled? Right now skipping the acknowledgements page is somewhat of an edge case but if the stepper gets built it could happen a lot more easily.
> 
> Agreed, it silently failing will definitely only be confusing for the user. We will need to consult with Zoe/Elliot about how we wan to handle these errors (display them/redirect to page with errors/should have disabled submit button while there are validation errors/etc)

There is a feature in CIF and some commented-out code in `useMutationWithErrorMessage` to handle this via a Context, allowing any failing mutation to set a page-level error and display an alert banner.

marcellmueller (approved):
Works great, nice job :rocket:

PR #678, opened 2022-09-20
fix: deploy mock_schema

VB-at-Bis:
### Description

The time travel feature was added, but mock_schema was not deployed, so the time travel does not work in dev and test environments.

This PR fixes the issue.

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -52,6 +52,7 @@ RUN chmod +x /usr/local/bin/dumb-init
 ENTRYPOINT ["dumb-init", "--", "/usr/bin/env", "bash", "-c"]
 
 COPY app/ ${APP_HOME}/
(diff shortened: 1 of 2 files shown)


matthieu-foucault (approved):
Looks good, let's see if it deploys properly

app/Dockerfile
```diff
@@ -52,6 +52,7 @@ RUN chmod +x /usr/local/bin/dumb-init
 ENTRYPOINT ["dumb-init", "--", "/usr/bin/env", "bash", "-c"]
 
 COPY app/ ${APP_HOME}/
+COPY mocks_schema/ ${HOME}/mocks_schema
```
matthieu-foucault, line 55: @VB-at-Bis I didn't pay attention... This is in the app Dockerfile, it should be in the db Dockerfile

PR #684, opened 2022-09-21
fix: in dev environment, drop mocks schema before redeploying

matthieu-foucault:
This is making our dev deploy fail unless you manually drop the schema before redeploying


marcellmueller (approved):

VB-at-Bis (approved):
LGTM

PR #690, opened 2022-09-22
fix: submit button is disabled if acknowledgements not checked

AntBush:
Fix for submit button being disabled if all acknowledgements aren't checked

### DIFF
--- a/app/tests/components/Form/ApplicationForm.test.tsx
+++ b/app/tests/components/Form/ApplicationForm.test.tsx
@@ -68,6 +68,11 @@ const submissionPayload = {
           submissionCompletedBy: 'test',
           submissionTitle: 'test',
         },
+        acknowledgements: {
+          acknowledgementsList: [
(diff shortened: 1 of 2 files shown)


app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -68,6 +68,11 @@ const submissionPayload = {
           submissionCompletedBy: 'test',
           submissionTitle: 'test',
         },
+        acknowledgements: {
+          acknowledgementsList: [
+            1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17,
+          ],
```
matthieu-foucault, line 74: `acknowledgementsList` is an array of strings. You could refactor this to use `acknowledgementsEnum` once #709 is merged

matthieu-foucault (approved):

PR #692, opened 2022-09-22
fix(ci): set gitleaks to static version

AntBush:
Fixes gitleaks to a specific version

### DIFF
--- a/.bin/install-gitleaks-linux-x64.sh
+++ b/.bin/install-gitleaks-linux-x64.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -xeuo pipefail;
 
-version="8.12.0";
+version="8.13.0";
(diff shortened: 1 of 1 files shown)


.bin/install-gitleaks-linux-x64.sh
```diff
@@ -2,7 +2,7 @@
 set -xeuo pipefail;
 
 version="8.12.0";
-releases_api="https://api.github.com/repos/zricethezav/gitleaks/releases/latest";
```
wenzowski, line 5: This is failing because `latest` is no longer `8.12.0` but is now `8.13.0`; why not upgrade instead?

AntBush: No reason, just thought it would've been more appropriate vs upgrading. Looking at the changelog though there isn't anything breaking so I'll go ahead and update

matthieu-foucault (approved):

Issue #462, opened 2022-08-19
HTTP Header fix


MarsRomer:

eazeredo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/462?planning-poker) with Zenhub @AntBush @marcellmueller @matthieu-foucault @wenzowski

wenzowski:
Start with Helmet? https://github.com/bcgov/cas-cif/blob/develop/app/server/middleware/headers.ts

matthieu-foucault:
If any of our zap header rules cannot be fixed on CI, we should add a comment indicating why, and whether it's fixed in production, e.g. because the openshift route is annotated accordingly

matthieu-foucault:
If any of our zap header rules cannot be fixed on CI, we should add a comment indicating why, and whether it's fixed in production, e.g. because the openshift route is annotated accordingly

PR #706, opened 2022-09-26
Set response headers

VB-at-Bis:
### Description

Project uses OWASP Zap scanning to automatically find security vulnerabilities. 
Currently some rules are ignored (https://github.com/bcgov/CONN-CCBC-portal/blob/main/.zap/rules.tsv)

This PR aims to remove all response headers-related rules from the "Ignore" list and thus improve application security

References #462


app/backend/lib/headers.ts
```diff
@@ -0,0 +1,36 @@
...
+    res.append("X-Content-Type-Options","nosniff");
+
+    // https://www.zaproxy.org/docs/alerts/10038/
+    res.append("Content-Security-Policy","default-src 'self'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; frame-ancestors 'self'; form-action 'self'");            
+
+    // https://www.zaproxy.org/docs/alerts/10063-1/
+    res.append("Permissions-Policy","display-capture 'none'");
+
+    // https://www.zaproxy.org/docs/alerts/10037/
+    res.removeHeader("X-Powered-By");
```
AntBush, line 30: Do we need this and the `poweredByHeader: false` in the config or is one or the other sufficient?

VB-at-Bis: As far as I understand, `poweredByHeader: false` in the config affects next.js which adds own 'powered-by', and express middleware works for everything including graphql requests for example. So I'd leave both :)

matthieu-foucault, line 30: Those are (or can be) set by `helmet`, no?

matthieu-foucault:
```
    contentSecurityPolicy: {
      useDefaults: false,
      directives: {
        "default-src": helmet.contentSecurityPolicy.dangerouslyDisableDefaultSrc,
        "script-src": ["'self'"],
      },
    },
```

mmmh, this doesn't look more readable than what you wrote. 

I'd say if they can be enabled with helmet with a simple boolean flag (e.g. `hidePoweredBy: true`), or if they are in the helmet defaults (nosniff and X-Frame-Options are enabled by default), then let's use helmet, otherwise `res.append` is better

VB-at-Bis: Ok.. let me remove res.append entries that match default Helmet settings and test

matthieu-foucault (approved):
:shipit:

app/backend/lib/headers.ts
VB-at-Bis: done

Issue #705, opened 2022-09-26
Funding total errors when all values are empty


matthieu-foucault:
1. the total funding for CIB should not be showing as red - it's optional (and autocalculated)
![image.png](https://images.zenhubusercontent.com/62686bda7385b9a511da26e2/6f6853fc-269c-4385-944e-b800f8bcfac2)
2. The total funding requested from partner is showing an old total despite deleting all the fiscal boxes:
![image.png](https://images.zenhubusercontent.com/62686bda7385b9a511da26e2/76a85905-8403-4e9e-bfc5-1c08e4b29c7a)
![image.png](https://images.zenhubusercontent.com/62686bda7385b9a511da26e2/a75d9743-a915-4a8a-af08-76e6f9569444)
If I type something in another box then it clears it from the review page

_Originally posted by @MarsRomer in https://github.com/bcgov/CONN-CCBC-portal/issues/414#issuecomment-1256673617_

Issue #682, opened 2022-09-21
Flaky test: ApplicationForm test 'acknowledgement page continue is enabled once all checkboxes have been clicked'


AntBush:
The test 'acknowledgement page continue is enabled once all checkboxes have been clicked' seems to fail periodically. I believe this is from how the test was written and needs to be updated.

matthieu-foucault:
Running into this while fixing #705

PR #709, opened 2022-09-26
Run calculations in `handleChange` instead of custom class component

matthieu-foucault:
Fixes the bugs in #705 by:
 - allowing a null value for `totalInfrastructureBankFunding`: easy change, set its type to `["number", "null"]` instead of `"number"`
 - running calculations in the `onChange` handler, rather than in the `CalculationForm`. This is a bigger change, but is needed as the `CalculationForm` is relying on some `setTimeout` which is problematic, as indicated in the comment below.

This PR also fixes the flaky test described in #682, as it was relying on a `waitFor`, maybe because of the timeout?


app/components/Form/CalculationForm.tsx
```diff
@@ -1,141 +0,0 @@
...
-    this.calculating = true;
-
-    // Pass through the new form data to re-render with the updated value
-    this.setState({ formData: e.formData });
-
-    // Get the updated field
-    const updatedField = this.getUpdatedField(this.state.formData, e.formData);
-
-    // Use setTimeout to call onCalculate, to let all the events fire to get everything synched up
-    setTimeout(() => this.onCalculate(e.formData, updatedField));
```
matthieu-foucault, line 68: @marcellmueller @AntBush @VB-at-Bis  When you're looking for reusable code for a problem or reviewing a PR, this kind of code should be considered a red flag. Using `setTimeout` or `useEffect` to work around issues where the state isn't up to date will almost always result in race conditions and bugs. In this case, some form updates where being skipped (if you typed fast enough characters were skipped) and the calculations were sometimes done using the previous form values, resulting in incorrect totals on the review page.

app/components/Form/ApplicationForm.tsx
```diff
@@ -23,27 +22,8 @@ import { useUpdateApplicationMutation } from 'schema/mutations/application/updat
...
-  'acknowledgements',
-  'estimatedProjectEmployment',
-  'projectFunding',
-  'otherFundingSources',
-  'submission',
-];
-
-const CUSTOM_SUBMISSION_FIELD = {
-  SubmissionField: SubmissionDescriptionField,
-};
```
matthieu-foucault, line 46: Not sure why this was in a constant here. I moved it to the default theme.

app/formSchema/pages/acknowledgements.ts
```diff
@@ -11,26 +31,8 @@ const acknowledgements: Record<string, JSONSchema7> = {
...
-            `The Applicant acknowledges that all activities required for this proposed Project must comply with all applicable federal, provincial, and territorial laws, regulations, municipal and other local by-laws.`,
-            `The Applicant acknowledges that knowingly making any false statements or misrepresentations, including by omission, in an application may affect its eligibility and may result in the revocation of funding approval.`,
-            `The Applicant acknowledges that information submitted in an application is subject to the Access to Information Act (Canada) or the Freedom of Information and Protection of Privacy Act (BC), as applicable.`,
-            `The Applicant confirms that, to the best of its knowledge, the information submitted in this application is true and correct as of the date of submission.`,
-          ],
+          type: 'string',
+          enum: acknowledgementsEnum,
```
matthieu-foucault, line 35: this schema could be further improved with #711

app/components/Form/ApplicationForm.tsx
```diff
@@ -325,27 +301,29 @@ const ApplicationForm: React.FC<Props> = ({
   };
 
   const handleChange = (e: IChangeEvent<any>) => {
+    if (isAcknowledgementPage) updateAreAllAcknowledgementFieldsSet(e.formData);
+    if (isSubmitPage) updateAreAllSubmissionFieldsSet(e.formData);
```
matthieu-foucault, line 305: Same as for the acknowledgement page, there is probably a way to enforce that all submission fields are set via the JSON schema spec.

app/components/Form/ApplicationForm.tsx
```diff
@@ -364,112 +342,13 @@ const ApplicationForm: React.FC<Props> = ({
...
-  const isCustomPage = customPages.includes(sectionName);
-  const isSubmitPage = pageNumber >= subschemaArray.length;
-  const isAcknowledgementPage = sectionName === 'acknowledgements';
-
-  const submitBtns = (
-    <SubmitButtons
-      disabled={!isSubmitEnabled || isSubmitting}
-      isUpdating={isUpdating}
-      isSubmitPage={isSubmitPage}
-      formData={formData}
```
matthieu-foucault, line 376: note that `formData={formData}` has been updated below to `formData={formData[sectionName]}`

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -4,11 +4,11 @@ import ComponentTestingHelper from '../../utils/componentTestingHelper';
 import compiledQuery, {
   ApplicationFormTestQuery,
 } from '__generated__/ApplicationFormTestQuery.graphql';
-import { fireEvent, screen, waitFor, within } from '@testing-library/react';
```
matthieu-foucault, line 7: This file is getting big. Created #712 to fix that

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -560,15 +573,97 @@ describe('The application form', () => {
...
     );
   });
-  describe('the review page', () => {
-    beforeAll(() => {
-      // Some rjsf features require window.crypto, which isn't provided by jsdom
-      Object.defineProperty(global.self, 'crypto', {
-        value: {
-          getRandomValues: (arr) => crypto.randomBytes(arr.length),
+
+  it('should set the correct calculated value on the employment page', async () => {
```
matthieu-foucault, line 577: moved from `CalculationForm.test.tsx`

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -560,15 +573,97 @@ describe('The application form', () => {
...
+            lastEditedPage: 'estimatedProjectEmployment',
+          },
         },
-      });
-    });
+      }
+    );
+  });
+
+  it('should set the correct calculated value on the project page', async () => {
```
matthieu-foucault, line 620: moved from `CalculationForm.test.tsx`

app/tests/components/Form/ApplicationForm.test.tsx
```diff
@@ -215,38 +215,45 @@ describe('The application form', () => {
...
-    componentTestingHelper.loadQuery();
+    componentTestingHelper.loadQuery({
+      ...mockQueryPayload,
+      Application() {
+        return {
+          id: 'TestApplicationID',
+          formData: {
+            ...mockFormData,
+            acknowledgements: {
+              acknowledgementsList: acknowledgementsEnum.slice(0, -1),
```
AntBush, line 234: This almost feels like cheating but it does serve the same purpose. Do we have any care about testing for clicking each box or since we have it covered in our e2e tests it's not a concern?

matthieu-foucault: reverted the changes

app/components/Form/ApplicationForm.tsx
matthieu-foucault: This could be done with a `oneOf`: either fields are required, or the review page checkbox is checked. Doing it this way would also mean that the only thing we're ignoring is the "required" attribute, and are still validating data types

AntBush (dismissed):
Looks good to me! Thanks for cleaning up the ApplicationForm component as well, and I appreciate you moving the acknowledgementsEnum out of the schema makes accessing it much easier

app/components/Form/ApplicationForm.tsx
```diff
@@ -226,19 +202,14 @@ const ApplicationForm: React.FC<Props> = ({
...
-    } else if (formData[sectionName]) {
-      newFormData = { ...formData };
-      newFormData[sectionName] = {
-        ...formData[sectionName],
-        ...newFormSectionData,
-      };
-    } else {
-      newFormData = { ...formData };
-      newFormData[sectionName] = { ...newFormSectionData };
-    }
```
matthieu-foucault, line 241: This was greatly simplified below

AntBush (approved):
:shipit:

Issue #655, opened 2022-09-14
FileWidget test is flaky


AntBush:
The FileWidget test produces flaky results. As seen in this action ([attempt 1 ](https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3055725858/attempts/1)/ [succeeded attempt](https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3055725858)) 

This implies that a race condition is present, and the purpose of this ticket is to:

- [ ] Determine whether this race condition is present in the test or the widget to determine severity
- [ ] Resolve this race condition

AntBush:
[Logs Archive](https://app.zenhub.com/files/476885245/ea4005a2-34b1-42a9-8ab3-da9fbce2d56d/download)

[Relevant Section](https://app.zenhub.com/files/476885245/e52e579b-842e-4d5c-8408-5a9a349be3d8/download)

matthieu-foucault:
Nothing for PO to review here, only code

PR #713, opened 2022-09-27
fix: no longer infinite loop on upload pages

AntBush:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/708

Removes the internal state and useEffects to update the form

### DIFF
--- a/app/cypress/integration/index.spec.js
+++ b/app/cypress/integration/index.spec.js
@@ -287,9 +287,8 @@ context('Homepage', () => {
 
     cy.get('h1').contains('Template uploads');
 
-    // // Todo: file uploads
-    // TODO: Remove this once test is no longer flaky (infinite loop)
-    // cy.get('header > div').contains('Last saved:');
+    // // Todo: file upload
+    cy.get('header > div').contains('Last saved:');
 
     cy.get('body').happoScreenshot({ component: 'Template Uploads Page' });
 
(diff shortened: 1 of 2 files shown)


matthieu-foucault (commented):
Should #655 be fixed in this PR too. Now that we don't have `useEffect`s , all of these `waitFor` in the test are probably not needed

AntBush:
> Should #655 be fixed in this PR too. Now that we don't have `useEffect`s , all of these `waitFor` in the test are probably not needed

I hope so? It's hard to tell if the flakiness is fixed or if we just didn't hit the error

matthieu-foucault:
> > Should #655 be fixed in this PR too. Now that we don't have `useEffect`s , all of these `waitFor` in the test are probably not needed
> 
> I hope so? It's hard to tell if the flakiness is fixed or if we just didn't hit the error

I meant, can you update to test to remove the `waitFor`s ? If the test works without `waitFor`s, it's less likely to be flaky

AntBush:
> > > Should #655 be fixed in this PR too. Now that we don't have `useEffect`s , all of these `waitFor` in the test are probably not needed
> > 
> > 
> > I hope so? It's hard to tell if the flakiness is fixed or if we just didn't hit the error
> 
> I meant, can you update to test to remove the `waitFor`s ? If the test works without `waitFor`s, it's less likely to be flaky

Ah yeah that makes much more sense 😅

AntBush:
https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3137446447/jobs/5095649787#step:5:4162

Looks like it doesn't solve the flaky issue? The test is failing without the waitFors (unless I'm missing some piece of additional work that needs to be done)

matthieu-foucault (approved):

PR #718, opened 2022-09-27
feat: add form_data and form_data_status_type tables

AntBush:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/508


db/deploy/computed_columns/application_form_data.sql
```diff
@@ -0,0 +1,16 @@
+-- Deploy ccbc:computed_columns/application_form_data to pg
+
+begin;
+
+create or replace function ccbc_public.application_form_data(application ccbc_public.application) returns jsonb as
```
matthieu-foucault, line 5:
```suggestion
create or replace function ccbc_public.application_form_data(application ccbc_public.application) returns ccbc_public.form_data as
```

having a computed column that returns the entire record instead of only the jsonb column would be preferable. You wouldn't need `application_form_data_id`, or other functions for the fields that get added later on

db/deploy/tables/form_data_status_type.sql
```diff
@@ -0,0 +1,28 @@
+-- Deploy ccbc:tables/form_data_status_type to pg
+
+begin;
+
+create table ccbc_public.form_data_status_type(
+  name varchar(1000) primary key,
+  description varchar(1000)
+);
+
+insert into ccbc_public.form_Data_status_type values
```
matthieu-foucault, line 10:
```suggestion
insert into ccbc_public.form_data_status_type values
```

db/deploy/tables/form_data_status_type.sql
```diff
@@ -0,0 +1,28 @@
...
+ ('draft', 'Draft'), ('submitted', 'Submitted');
```
matthieu-foucault, line 11: Should we rename those to `pending` and `committed`, to avoid confusion with the application statuses?

AntBush: I think so, will also alleviate confusion of having a `withdrawn` application but a `submitted` form

PR #719, opened 2022-09-27
feat: number of households allow decimals

matthieu-foucault:
Other than the change above, this PR contains the following refactor:
 - `NumberWidget` is used by default for all numbers
 - Added a `NumericStringWidget`, taking code out of `NumberWidget`


marcellmueller (approved):
Nice refactor, I think this is a lot better now.

PR #723, opened 2022-09-28
467 mvp application stepper

marcellmueller:
https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/467

Settled on a fixed width for the stepper so that we wouldn't get a jumpy effect when changing to pages with longer names. This will probably be a good one to pull down and test locally when reviewing.

### DIFF
--- a/app/components/Layout.tsx
+++ b/app/components/Layout.tsx
@@ -20,7 +20,6 @@ const StyledLayout = styled('div')`
 `;
 
 const StyledMain = styled('main')`
-  margin: auto 50px;
   display: flex;
   width: 100%;
   max-width: ${(props) => props.theme.width.pageMaxWidth};
--- a/app/components/Layout.tsx
+++ b/app/components/Layout.tsx
@@ -6,6 +6,7 @@ import getConfig from 'next/config';
 import styled from 'styled-components';
 
 const runtimeConfig = getConfig()?.publicRuntimeConfig ?? {};
+
 const StyledFooter = styled(Footer)`
   width: 100%;
 `;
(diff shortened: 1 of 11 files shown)


app/components/Layout.tsx
```diff
@@ -14,17 +15,17 @@ const StyledLayout = styled('div')`
   display: flex;
   flex-direction: column;
   justify-content: space-between;
+  align-items: center;
   min-height: 100vh;
 `;
+
 const StyledMain = styled('main')`
   margin: auto 50px;
```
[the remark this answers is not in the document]
marcellmueller: I've removed the entire line.

app/components/Stepper.tsx
```diff
@@ -0,0 +1,86 @@
...
+  padding: 16px 40px;
+`;
+
+const Stepper = () => {
+  const router = useRouter();
+  const rowId = router.query.id;
+
+  const formPageList = uiSchema['ui:order'].filter((page) => {
+    return page !== 'submission';
+  });
```
matthieu-foucault, line 52: this never changes, so you can move it above the component to avoid recalculating it on every render

app/components/Stepper.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const formPageSchema = schema.properties;
+
+  return (
+    <StyledNav>
+      {formPageList.map((page) => {
+        const isCurrentPage =
+          formPageList[Number(router.query.page) - 1] === page;
+
+        const pageNumber = getFormPage(page);
```
matthieu-foucault, line 62: `page` is the sectionName, but `router.query.page` is the page number. This is potentially confusing. Renaming `page` to `formName` or `formKey` would potentially make it clearer. 

You should also compare indices directly instead of searching for the page name in an array to retrieve its page number.

```suggestion
      {formPageList.map((formName, index) => {

        const pageNumber = index + 1
        const isCurrentPage = Number(router.query.page) === pageNumber
```

app/tests/components/Stepper.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+    expect(acknowledgements).toHaveAttribute('href', '/form/1/20');
+  });
+
+  it('should have correct background colour when not active active', () => {
+    renderStaticLayout();
+    const link = screen.getByText('Project area').parentElement;
+
+    const style = window.getComputedStyle(link);
+
+    expect(style.backgroundColor).toBe(theme.color.stepperGrey);
```
matthieu-foucault, line 114:
```suggestion
    expect(link).toHaveStyle({backgroundColor: theme.color.stepperGrey});
```
This works too, see https://github.com/testing-library/jest-dom#tohavestyle

matthieu-foucault (approved):
Looks great! All my comments are somewhat minor refactor and marginal performance improvements

PR #725, opened 2022-09-28
revert: disable CSP header

matthieu-foucault:
temporary regression while we get #741 to work properly


VB-at-Bis (approved):
Looks good for now

PR #735, opened 2022-10-03
fix: invert title and description in FileWidget

matthieu-foucault:
- use help instead of description where appropriate
- see Happo screenshots for outcomes of PR: the only visual changes are on the review page


marcellmueller (approved):
Looks good!

PR #738, opened 2022-10-04
chore: simplify PR template

matthieu-foucault:


.github/pull_request_template.md
```diff
@@ -1,49 +1,11 @@
 ---
 name: Pull Request Template
 about: Default pull request template
-title: ''
+title: '<type>[optional scope]: <description>'
```
VB-at-Bis, line 4: Shall we add list of types like it was before ( bug fix, new feature, breaking change, documentation only).
What is scope here? FE, BE, test, devops or something else?

matthieu-foucault: I can add some support documentation, following the example of Angular: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

VB-at-Bis (approved):
Looks good. Thanks for team agreement update!

matthieu-foucault:
Looks like the PR template didn't work as expected... opening a new PR shortly

PR #741, opened 2022-10-04
fix/eslint

marcellmueller:


matthieu-foucault:
@marcellmueller looks like the autofix actually broke some things

matthieu-foucault (approved):
Getting there! Feel free to momentarily turn off the eslint check and merge this.

Issue #695, opened 2022-09-23
Change eligibility mapping tool stats upload


elliottSlee:
### User Story
As an applicant
I want the eligibility mapping tool stats upload to be restricted to xml files
so that I upload the correct file.

### Acceptance Criteria:
- [ ] Update wording of input description
"Please upload the XML file that was attached to the email you received upon completion of the project coverage."
- [ ] file type is restricted to XML

![Screen Shot 2022-09-23 at 11.00.34 AM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/3cbe35c0-7e32-45dc-a57e-5dde0100ff22)

PR #748, opened 2022-10-06
feat: change eligibility mapping accepted file type and description

marcellmueller:
Implements #695


VB-at-Bis (approved):
Looks safe to merge

PR #750, opened 2022-10-07
chore(db): remove sqitch verify scripts

matthieu-foucault:
Quick PR to simplify our db folder


VB-at-Bis (approved):
Looks safe - we did not call `sqitch verify` anywhere

matthieu-foucault:
> Looks safe - we did not call `sqitch verify` anywhere

@VB-at-Bis it was called after every deploy, but I don't think it's needed, given that we have transactions and many automated tests

AntBush (approved):

Issue #121, opened 2022-05-10
Make application form uneditable when intake closes


elliottSlee:
### User Story
As an analyst
I want submitted applications to lock on the date that the intake closes
so that changes won't be made after I begin reviewing.

### Description
Network BC & Innovation, Science, and Economic Development Canada would like applications to be editable until the intake closes then be locked at a particular closing date so they can begin looking at applications without them changing.

### Acceptance Criteria
- [x] Dashboard is visible in between intakes (after the first intake and subsequent intakes)
- [x] **No edit button**
  Given that an application is status **Received** (or submitted and intake is closed)
  when the applicant is on the dashboard
  then the edit button will be removed
  and a view button appears

#
Added by Tony, reflected in PR 770 and 751
- [x] Given **any** intake is open
      when an application with status *draft* is open (in this case it is also represented as a form with status pending)
      then the form is editable
- [x] Given an application has been submitted
    when it's **respective** intake is open
    then the form is editable

- [x] In all other cases, it is uneditable
#
#### Technical details
This behaviour is currently implemented by comparing the client-side date with the intake close timestamp. The client's date should not be relied upon. This should be implemented with an `isEditable` [computed column](https://www.graphile.org/postgraphile/computed-columns/) on the `form_data` table:
`form_data_is_editable()` returns `true` if `form_data.status` is `draft` and there is an open intake or `form_data.status` is `submitted` and application's intake is open

#
- [x] **Input fields disabled**
  Given that an applications status is either received or submitted and its intake is closed (past intake close date in database; application's intake id != open intake id)
  when an applicant views the application
  then they will be directed to the review page
  and when they navigate to other pages
  then all input fields will be disabled
  and when they navigate to the submit page
  then the submit button will be disabled
  and a "Return to dashboard" button will be visible beside the disabled submit button.
#

![Screen Shot 2022-09-27 at 11.53.02 AM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/abf21e11-37c6-4e44-bad9-a2d0e519edb5)

eazeredo:
@AntBush , please, clarify the Acceptance Criteria regarding the rule changing the IsEditable to something like
`If Intake is "open" And Status in ("Draft," "Submitted").`

AntBush:
There would be two conditions: 
1. Given **any** intake is open
    when an application with status *draft* is open (in this case it is also represented as a form with status pending)
    then the form is editable
2. Given an application has been submitted
[...]

PR #751, opened 2022-10-07
feat: application form uneditable when intake closes

AntBush:
Implements #121

Adds a computed column `form_data_is_editable` and uses that to check wheher the form should be editable rather than doing it on the frontend.


VB-at-Bis (approved):
Looks good; changes as discussed

Issue #538, opened 2022-08-29
Status changes to Received (automated)


MarsRomer:
### User Story
As an applicant
I want my submitted applications to be received by NetworkBC
so that my application will be assessed.

### Acceptance Criteria
- [ ] **Status changed to Received**
  Given that an application status is _Submitted_
  When the current date & time are after intake closing date & time (or current DateTime-TimeTravel feature)
  Then the date & time are recorded in the database
  And the status is changed to _Received_.

Note: Will need to handle scenarios here where the intake close date changes. Updates to the job will be required if the closing date changes.

### Dev notes
- [x] add `receive_applications()` function that adds a 'Received' status for every application that is submitted and where the intake is closed.
- [x] create `ccbc_job_executor` role in the DB
- [x] `ccbc_job_executor` can see all applications, intake, application statuses, etc...
- [x] `ccbc_job_executor` can insert `application_status`
- [x] configure cronjob in helm chart to call `psql -d ccbc -c 'set role ccbc_job_executor; select ccbc_public.receive_applications();'` twice a day at 9am/pm PST
- [x] cronjob should be run with `ccbc_app` user and `ccbc_job_executor` role
- [x] add received status type for applications

MarsRomer:
Are the newest changes in test?

matthieu-foucault:
They are, yes. To run this in the test environment, we need to:
- manually update the intake close date
- manually trigger the cron job

MarsRomer:
thanks was just making sure :)

PR #754, opened 2022-10-11
538 status changes to received

marcellmueller:
This is for issue #538. 

Definitely open to more ideas for test cases. I know @matthieu-foucault said we should ideally use the `create_application` function in tests though since I needed to add specific intakes to them I stuck with `insert`, let me know if you have other thoughts thanks.


db/deploy/create_roles.sql
```diff
@@ -10,7 +10,7 @@ begin
     select true
     from pg_catalog.pg_roles
     where rolname = 'ccbc_guest') then
-    
+
```
AntBush, line 13: It looks like we did a rework and then didn't change the file other than some spacing fixes (I'm guessing be prettier). I think we should remove the rework for readability

db/deploy/functions/receive_applications.sql
```diff
@@ -0,0 +1,49 @@
...
+    open applications for select id, intake_id
+		    from ccbc_public.application
+		    where intake_id in (
+                select id from ccbc_public.intake where now() >= close_timestamp);
+
+    loop
+        fetch applications into current_app;
+        exit when not found;
+
+        select s.status into last_status from ccbc_public.application_status s inner join
```
AntBush, line 27: We can probably use the `application_status` computed column function here to simplify this, we should add execute permission for that function. There's also the possibility that how we store an application status changes and then we'd have two places to have to check/change.

db/revert/ccbc_job_executor_permissions.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ccbc:ccbc_job_executor_permissions from pg
+
+begin;
+
+-- XXX Add DDLs here.
```
AntBush, line 5: I'm guessing we want a revert in here if we want to remove this role at one point? Looking at the old revert for `create_roles` it implies we only want to do it manually so I'm not sure, maybe @matthieu-foucault can shed some light on this

db/test/unit/functions/receive_applications_test.sql
```diff
@@ -0,0 +1,66 @@
...
+delete from ccbc_public.application;
+delete from ccbc_public.application_status;
+
+insert into
+  ccbc_public.intake(id, open_timestamp, close_timestamp, ccbc_intake_number)
+overriding system value
+values
+  (1, '2022-08-19 09:00:00 America/Vancouver','2022-11-06 09:00:00 America/Vancouver', 1),
+  (2, '2021-08-19 09:00:00 America/Vancouver','2021-11-06 09:00:00 America/Vancouver', 2);
+
```
AntBush, line 22: We should add the `mocks.set_mocked_time_in_transaction` function here as the tests will start to fail after Nov 6

db/test/unit/roles/create_roles_test.sql
```diff
@@ -14,6 +14,11 @@ select isnt_superuser(
     'ccbc_guest should not be a super user'
 );
 
+select has_role( 'ccbc_guest', 'role ccbc_job_executor exists' );
```
AntBush, line 17:
```suggestion
select has_role( 'ccbc_job_executor', 'role ccbc_job_executor exists' );
```

db/verify/functions/receive_applications.sql
```diff
@@ -0,0 +1,7 @@
+-- Verify ccbc:functions/receive_applications on pg
+
+BEGIN;
+
```
AntBush, line 4: We can just delete verify files as we aren't using them anymore, same applies to the rest of the `verify` files

AntBush (changes requested):
Appreciate the work! It's good stuff, just a couple minor tweaks

db/deploy/create_roles.sql
marcellmueller: I'm confused as to what you mean, we did the sqitch rework to add the `ccbc_job_executor` in `create_roles@1.5.0.sql`. This white space change is from the pre commit hook.

AntBush: Ah I misunderstood, the `.sql` file without the tag is the latest one while the tagged version is the old version, so you have them swapped in this case. Just need to change the file names in this case then. When this is deployed in prod it won't create the latest roles

db/deploy/functions/receive_applications.sql
marcellmueller: @VB-at-Bis thoughts on this?

app/cypress/integration/upload.spec.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
+/// <reference types="Cypress" />
+
+context('Homepage', () => {
```
AntBush, line 4: I think this got added in by accident

db/revert/ccbc_job_executor_permissions.sql
marcellmueller: @AntBush okay I took care of everything except for this, take a look when you get time. 

@matthieu-foucault did you have any thoughts on this comment?

matthieu-foucault: > thoughts

The comment in the old revert from `create_roles` don't necessarily apply to this project.

marcellmueller: I'm not sure I get the point of this, maybe if we dropped the other roles on revert but given how `create_roles` revert currently is I'm not going to bother.

AntBush (approved):

Issue #699, opened 2022-09-26
Landing page updates when intake 1 closes


MarsRomer:
### User Story 
As an applicant when the the first intake ends and the second intake is not open yet
I want clear description of what is happening
so I know how to manage my applications

### Acceptance Criteria
- [ ] **Scenario: Intake closed, no upcoming intake known**
Given that the current date in the database is after an intake close date
and there is no upcoming intake in the database
when a user is on the landing page
then they will see a callout that states,
"Applications are not currently being accepted"
"Intake [Previous intake number] has ended ([previous intake start date] - [previous intake end date])"
#
- [ ] **Scenario: intake closed, upcoming intake in database**
Given that the current date in the database is after an intake close date
and there is an upcoming intake in the database
when a user is on the landing page
then they will see a callout that states,
"Applications are not currently being accepted."
"The next application intake begins [intake opening date]."
#
- [ ] **Scenario: intake is open**
Given that the current date in the database is between an intake open & close date
when a user is on the landing page
then they will see a callout that states,
"Applications are accepted until [intake close date]."
"Review of applications will not begin until this date. Draft and submitted applications can be edited until then."

### Design
https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=6344%3A28591
https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=6344%3A27752

elliottSlee:
@zoeyli-46 & I opted to have explanations of actions the applicant can and cannot take not on the landing page because users largely didn't read the landing page. Also, these actions have been intuitive for users in usability testing thus far. If these explanations are necessary, it would be best to have them where the actions are possible (i.e., dashboard).

matthieu-foucault:
This looks straightforward and can be estimated if that good looks to you @MarsRomer

matthieu-foucault:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/699?planning-poker) with Zenhub @AntBush @marcellmueller @VB-at-Bis

matthieu-foucault:
blocked pending clarifications from @elliottSlee :
- do we still want to show the current alert re. ISED's eligibility mapping tool?
[...]

PR #765, opened 2022-10-12
feat: add callout displaying intake state on landing page

matthieu-foucault:
Implements #699 + adds second intake in dev and test data

### DIFF
--- a/app/pages/index.tsx
+++ b/app/pages/index.tsx
@@ -36,8 +36,7 @@ const BoldText = styled('strong')`
 `;
 
 const StyledCallout = styled(Callout)`
-  margin-top: 16px;
-  margin-bottom: 16px;
+  margin: ${(props) => props.theme.spacing.medium} 0;
 `;
 
 const getPagesQuery = graphql`
--- a/app/pages/index.tsx
+++ b/app/pages/index.tsx
@@ -3,6 +3,9 @@ import { withRelay, RelayProps } from 'relay-nextjs';
 import { graphql } from 'react-relay';
 import Link from '@button-inc/bcgov-theme/Link';
 import styled from 'styled-components';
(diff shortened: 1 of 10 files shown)


app/pages/index.tsx
```diff
@@ -25,13 +28,26 @@ const StyledBtnContainer = styled('div')`
...
+// while a `strong` element may be displayed as bold,
+// that should not be relied upon and font-weight should be used instead.
+// Using a `strong` element is still useful on a semantic level.
+const BoldText = styled('strong')`
+  font-weight: bold;
+`;
+
+const StyledCallout = styled(Callout)`
+  margintop: 16px;
+  marginbottom: 16px;
```
marcellmueller, line 40:
```suggestion
  margin-top: 16px;
  margin-bottom: 16px;
```

Or just `margin: 16px 0` would be better. Though I don't believe this is will actually do anything anyways as it is already the default element style:

margin

marcellmueller (changes requested):
Looks great other than that one issue I commented on.

app/pages/index.tsx
matthieu-foucault: Good catch! This is a half-baked refactor :sweat_smile:

marcellmueller: I'd say so, it is cleaner and ensures the value of margin on both sides is identical. Though if we were using a global margin variable to ensure they were coming from the same constant I could see why you would think it's easier to read.

matthieu-foucault: I was hesitating between multiple options. Now there's both the shorthand property, and a `spacing` section in our theme.

marcellmueller: Love it, nice job.

marcellmueller (approved):

Issue #223, opened 2022-06-14
Multiple Login Buttons for BCeID providers


MarsRomer:
### User Story
As an applicant 
I want the login button to lead directly to BCeID
So that I don't try using options not relevant to me

### Acceptance Criteria
- [x] Update landing page to include buttons for login to Business BCeID, Basic BCeID, and Register for Basic BCeID
- [x] **Business BCeID Login button** bypasses login options and leads directly to the Business BCeID login page
- [x] **Basic BCeID Login button** bypasses login options and leads directly to the Basic BCeID login page
- [x] **Register for Basic BCeID button** leads directly to the Basic BCeID registration page

### Technical Details
- Each login button should call `/login` with a `kc_idp_hint` parameter
- The `authorizationUrlParams` option of the  `sso-express` need to be defined as a function, forwarding the `kc_idp_hint` value (not without ensuring that it's one of the accepted values first, as blindly forwarding request parameters is a security issue)

![Screen Shot 2022-10-05 at 2.27.59 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/2fe0391e-aa92-4845-a585-bb8f1e1acde1)

elliottSlee:
@MarsRomer before we develop this can we discuss it some more? Applicants would be able to see this option on every page, which isn't very good UX. Instead of the link in footer, what do you think of having two login buttons on the landing page:
![Screen Shot 2022-08-23 at 2.26.30 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/d29b97be-5190-4207-bf6c-11bcf640272f)

MarsRomer:
@matthieu-foucault we have decided that we don't need an IDIR login for the prod env (at least not for now).  I just want to confirm that if we don't implement this ticket, then the login will not have the IDIR option, or will it?  Perhaps we need to reword this ticket for Removing IDIR only?

matthieu-foucault:
We need the first acceptance criteria of this ticket, "Login button bypasses login options and leads directly to the BCeID login page"

Github login isn't an option anymore, so we can update the AC to have IDIR login as the conditional login for dev/test only

matthieu-foucault:
Blocked on SSO config issue, see https://chat.developer.gov.bc.ca/channel/sso?msg=AsTDGZXxWhkR9E49m

elliottSlee:
[...]

PR #768, opened 2022-10-13
feat: landing page shows multiple login buttons

matthieu-foucault:
Implements #223


marcellmueller (changes requested):
Works great. Only minor design issue, in the ticket all the buttons are the same width where in this implementation they are as wide as the text allows. I think ideally they would be all the same just as in the example in the [issue](https://github.com/bcgov/CONN-CCBC-portal/issues/223).

Screenshot at 2022-10-17 09-55-00

marcellmueller (approved):

PR #769, opened 2022-10-14
feat: use renovate for docker dependencies

AntBush:
Implements # 762

This branch will not create PRs since the config is not yet in the main branch

Introduces renovate to our CI chain, which will trigger whenever trivy finds a vulnerability to update our docker file.

Since our dockerfiles run a `yum update` then the first action if this is to occur is to run the `cache.yaml` workflow from our actions to get an updated package.


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: 'docker' # See documentation for possible values
+    directory: '/db' # Location of package manifests
+    schedule:
+      interval: 'daily'
```
matthieu-foucault, line 11:
```suggestion
      interval: 'daily'
      time: "00:00"
      timezone: "America/Vancouver"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+  - package-ecosystem: 'docker'
+    directory: '/app'
+    schedule:
+      interval: 'daily'
```
matthieu-foucault, line 15:
```suggestion
      interval: 'daily'
      time: "00:00"
      timezone: "America/Vancouver"
```

matthieu-foucault (changes requested):
Let's run our dependabot overnight so that we don't get randomly swarmed with PRs blocking actions in the middle of the day.

Is there no way to only trigger dependabot if there is a security update with the docker image? This configuration may result in a lot of unnecessary PRs

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: 'docker' # See documentation for possible values
+    directory: '/db' # Location of package manifests
```
VB-at-Bis, line 9: only /db ? what about /app ?

AntBush: It's just on the next one since directory won't take an array, although if you check the discussion on teams we won't be using dependabot since we can't get the function we want out of it

.github/workflows/test-containers.yaml
```diff
@@ -94,3 +103,22 @@ jobs:
...
           sarif_file: 'trivy-results.sarif'
+      - name: Get Results Length From Sarif
+        id: scan-results-length
+        run: echo "results-length=$(cat trivy-results.sarif | jq '.runs[0].results | length')" >> $GITHUB_OUTPUT
+
+  renovate:
+    needs: ['trivy-scan-db', 'trivy-scan-app']
+    if: needs.trivy-scan-db.outputs.results-length > 0 || needs.trivy-scan-app.outputs.results-length > 0
+    env:
+      LOG_LEVEL: 'debug'
```
matthieu-foucault, line 114: We probably don't want to run CI tools in debug mode?
```suggestion
```

matthieu-foucault (changes requested):

matthieu-foucault (approved):

Issue #605, opened 2022-09-06
upload: e2e test


matthieu-foucault:

eazeredo:
Rough estimated in 5!

matthieu-foucault:
5! = 120 :scream:

PR #772, opened 2022-10-17
test: e2e upload file tests

VB-at-Bis:
Implements #605


app/cypress/integration/upload.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    cy.get('body').happoScreenshot({ component: 'Dashboard Page' });
+
+    cy.get('a').contains('Edit').click();
+    cy.wait(2000);
+    cy.get('a').contains('Supporting documents').click();
+    cy.wait(2000);
+    
+    // cy.visit('http://localhost:3000/applicantportal/form/1/12');
+    cy.get('#root_copiesOfRegistration-btn').click(); 
+    cy.get('[data-testid=file-test]').first().selectFile('cypress/fixtures/doc.txt', { force: true });
```
matthieu-foucault, line 41: Can we use https://testing-library.com/docs/cypress-testing-library/intro/ to select elements in cypress tests? Using ids doesn't match the [testing-library guiding principles](https://testing-library.com/docs/guiding-principles), which we're trying to follow

app/cypress/integration/upload.spec.js
```diff
@@ -0,0 +1,47 @@
...
+
+    // Dashboard page
+    cy.get('h1').contains('Dashboard');
+
+    cy.get('body').happoScreenshot({ component: 'Dashboard Page' });
+
+    cy.get('a').contains('Edit').click();
+    cy.wait(2000);
+    cy.get('a').contains('Supporting documents').click();
+    cy.wait(2000);
```
matthieu-foucault, line 37: Instead of clicking its way through the application, our test could create the application via an sql fixture (we could create a `db/data/testFixtures/applications.sql` file) and do a `cy.visit('/form/1/12')`.

We should also have a `reset_db.sql` file in our fixtures, and this file should be run before every test, to avoid side effects. Looking at the happo diff, this test has a side effect (it creates an application, which shows up in the next test).

app/cypress/integration/upload.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    
+    // cy.visit('http://localhost:3000/applicantportal/form/1/12');
+    cy.get('#root_copiesOfRegistration-btn').click(); 
+    cy.get('[data-testid=file-test]').first().selectFile('cypress/fixtures/doc.txt', { force: true });
+    cy.wait(2000);
+    cy.get('a').contains('doc.txt');
```
matthieu-foucault, line 43: If we trust the UI to check that the file was indeed uploaded, do we need a test somewhere else to ensure that a request is sent to our S3 storage? Maybe we want a backend test suite for `resolveFileUpload` ?

app/cypress/integration/upload.spec.js
VB-at-Bis: hmm .. our main Cypress test is full of ids: https://github.com/bcgov/CONN-CCBC-portal/blob/main/app/cypress/integration/index.spec.js#L59
Besides, all out upload controls on the page look exactly same: 
```<input data-testid="file-test" type="file" required="" style="display: none;">```

Comment (VB-at-Bis) [reply app/cypress/integration/upload.spec.js:41]: I will change `cy.get('#root_copiesOfRegistration-btn').click(); ` to `cy.get('input[id="root_copiesOfRegistration-btn"]).click(); ` though..

File: app/cypress/integration/upload.spec.js (hunk 2 above)
Comment (VB-at-Bis) [reply app/cypress/integration/upload.spec.js]: Done

File: app/cypress/fixtures/doc.txt [hunk 4]
```diff
@@ -0,0 +1 @@
+placholder
```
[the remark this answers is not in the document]
VB-at-Bis: Done

app/cypress/integration/upload.spec.js
VB-at-Bis: Done

app/cypress/integration/upload.spec.js
VB-at-Bis: Posted a question for discussion in team's channel

marcellmueller (approved):
Great work :)

PR #775, opened 2022-10-17
feat: submit button is disabled if form is uneditable

AntBush:
Implements missing detail  for disabled submit in #121


marcellmueller (changes requested):
No tests?

AntBush:
> No tests?

Added some, now just need to wait...

marcellmueller (approved):
Thanks for adding some tests, I appreciate that. 

I was testing this and it's working great. It's a nitpick but I noticed the return to dashboard doesn't have a left margin on it like the save as draft button.

Mattieu said we aren't doing pixel perfect design so I'll approve this and leave it up to you if you want to change that.

Screenshot at 2022-10-18 12-18-34
Screenshot at 2022-10-18 12-18-18

Issue #26, opened 2022-04-05
Log in - internal users (IDIR)


eazeredo:
### User Story
As a user, 
I want to login with IDIR 
so that I can securely access the received applications.

### Preconditions
Developers have a list of IDIRs that are granted Analyst access to all features in the Portal, and IDIRs that are granted Administrator additional access to configuring intake dates

### Authentication (User identity verification)
- [x] Given the SSO is enabled 
When a User accesses the login page
And the user is previously authenticated 
Then the user is redirected to the analyst dashboard, 
#
- [x] Given the SSO is enabled 
When a User accesses the login page
And the user is NOT previously authenticated 
Then the user is automatically redirected to the Landing page url: www.connectingcommunitiesbc.ca/analyst
AND "Login with IDIR" button leads directly to IDIR login page.
#
- [x] If the IDIR credentials are correct, but the user with the IDIR does not have access to the analyst portal, then the user is redirected to the error page 

### Authorization (Permittion verification) 
- [X] Create `Analyst` and `Admin` roles in Single Sign On app
#
- [x] Enable IDIR IDP in CSS app
#
- [ ] Assign list of IDIRs to the `Analyst` and `Admin` roles 
#
- [x] Validate Permission (if a role has permission to perform an action)
#
- [x] When the credentials do not match the list of IDIRS with access, 
Then they will receive an error alert saying "Your IDIR has not been granted access"
#
- [x] Train PO to manage roles

##  Design
**Landing Page**
![Screen Shot 2022-10-05 at 2.29.48 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/1d158a01-c8ee-4517-85fb-1a7e0a49289a)
#
**Error**
![image.png](https://images.zenhubusercontent.com/61e8875b1de5eff8f072cd04/cf5ef635-7912-4871-a801-76c9aa960f2d)

AntBush:
> If there is SSO, Analyst is automatically authenticated and redirected to the analyst dashboard,

Can we remove this? Matthieu has identified that this is not possible

MarsRomer:
Blocked by:  need to determine if we will use siteminder or Azure for login

MarsRomer:
needs design for if log in fails

matthieu-foucault:
@MarsRomer @marcellmueller did the meeting with SSO team and security happen, and do we have recommendations for the identity provider choice?

MarsRomer:
[...]

PR #784, opened 2022-10-19
26-update-base-url

marcellmueller:
Implements #26


marcellmueller:
@matthieu-foucault these e2e tests were weird to debug, they worked fine locally but were timing out in CI. I changed the baseurl to `/applicantportal` and increased the wait in this commit which felt a little wrong but it is passing now:

https://github.com/bcgov/CONN-CCBC-portal/pull/784/commits/0c6cff164cd8ddf2e46560a26104d72508651689

AntBush (approved):
Looks good, something to consider for the future is right now we have a couple references to `/` that auto-redirects to `/applicantportal` that will end up be confusing to analysts in the future

PR #786, opened 2022-10-19
fix: receive-applications job should run with ccbc_app user

matthieu-foucault:
re. #538


marcellmueller (approved):
Thank you :bow:

marcellmueller:
> re. #538

Oh actually just a minor yaml issue it looks like

PR #788, opened 2022-10-20
chore(ci): use old digest to test renovate

AntBush:
Test to see if renovate will behave as expected


AntBush:
Works just as expected!

PR #790, opened 2022-10-20
26 idir landing

marcellmueller:
Implements #26


docs/analyst_permissions.md
```diff
@@ -0,0 +1,16 @@
+### Overview of analyst permissions
```
marcellmueller, line 1: @matthieu-foucault I started this but was unsure of what you all wanted me to add here.

app/lib/relay/withRelayOptions.tsx
```diff
@@ -15,17 +16,32 @@ const withRelayOptions: WiredOptions<any> = {
...
+        },
+      };
+    }
+
+    // Redirect them to landing route responding to the page they are trying to access
     return {
       redirect: {
-        destination: '/',
+        destination: request.url.startsWith('/analystportal')
+          ? '/analystportal'
```
matthieu-foucault, line 43: remember to replace all the `/analystportal` to `/analyst`, as per the issue's acceptance criteria

marcellmueller:
@matthieu-foucault this is the commit I added redirects for authenticated users included in the acceptance criteria:
 "Given the SSO is enabled
When a User accesses the login page
And the user is previously authenticated
Then the user is redirected to the analyst dashboard"

https://github.com/bcgov/CONN-CCBC-portal/pull/790/commits/bbc8baf0b8bcf14b4ca404689e277ce9f8b602bf

app/pages/analyst/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+          <p>
+            NetworkBC members with access to Connecting Communities BC (CCBC)
+            applications may log in with their IDIR.
+          </p>
+
+          {session?.sub ? (
+            <StyledBtnContainer>
+              <ButtonLink href="/analyst/dashboard">Go to dashboard</ButtonLink>
+            </StyledBtnContainer>
+          ) : (
```
matthieu-foucault, line 48: Will this ever show now that this page automatically redirects to the dashboard?

app/pages/analyst/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+      authRole === 'ccbc_admin' || authRole === 'ccbc_analyst';
+
+    // They're logged in and have a ccbc_admin or ccbc_analyst role
+    if (isAuthenticatedAnalyst) {
+      return {
+        redirect: {
+          destination: `/analyst/dashboard`,
+        },
+      };
+    }
```
matthieu-foucault, line 77:
```suggestion
    const authRole = getAuthRole(request);
    const isAuthenticatedAnalyst =
      authRole?.pgRole === 'ccbc_admin' || authRole?.pgRole === 'ccbc_analyst';

    // They're logged in and have a ccbc_admin or ccbc_analyst role
    if (isAuthenticatedAnalyst) {
      return {
        redirect: {
          destination: authRole.landingRoute,
        },
      };
    }
```

app/pages/analyst/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+    // They're logged in with IDIR but don't have an authorized role
+    if (isIdirUser && !isAuthenticatedAnalyst) {
+      return {
+        redirect: {
+          destination: `/analyst/request-access`,
+        },
+      };
+    }
+
+    return {};
```
matthieu-foucault, line 88:
```suggestion
    return defaultRelayOptions.serverSideProps(ctx);
```

There's some redirects that are handled there that you probably want to reuse (e.g redirect applicants to their dashboard)?

docs/analyst_permissions.md
matthieu-foucault: This looks great. Maybe some technical details can be added, regarding pages access:
 - which pages can be accessed by which user
 - what is the page redirection behaviour
 - a note that client-side routing will not go through the page redirection logic. The other day, I was suggesting that the logo link should use the back-end routing instead of a Nextjs Link. I remember now that this is the reason why

app/pages/analyst/index.tsx
marcellmueller: No, I was unsure if I should remove it this morning but I've gone ahead and removed it.

docs/analyst_permissions.md
marcellmueller: Sounds good, I updated the docs.

docs/analyst_permissions.md
```diff
@@ -0,0 +1,53 @@
...
+Landing route: `/analyst/dashboard`
+
+###### Applicant portal BCeID user:
+
+Role given: `ccbc_auth_user`
+Landing route: `/applicantportal/dashboard`
+
+#### Client side routing
+
+Client side routing will not go through the page redirection login so keep that in mind when using `next/link` as it could cause security issues and bugs. In many cases it will be preferrable to trigger a page reload.
```
matthieu-foucault, line 53:
```suggestion
Client side routing will not go through the page redirection login so keep that in mind when using `next/link` as it could cause bugs. In many cases it will be preferrable to trigger a page reload.
```

If front-end bugs can cause security issues, we're in trouble. The security (i.e. authorization layer), is on the DB side, so what would happen is either an exception being thrown, or the page not showing any data.

matthieu-foucault (approved):

docs/analyst_permissions.md
marcellmueller: That makes sense thank you.

PR #795, opened 2022-10-20
feat: add roles ccbc_admin and ccbc_analyst

marcellmueller:
Implements # \<issue number\>


AntBush (approved):
LGTM! I know we're adding permissions for roles later so not much else to do here

PR #798, opened 2022-10-21
fix: ccbc_job_executor can execute receive_applications()

matthieu-foucault:


marcellmueller (approved):
Thanks again for doing this.

Issue #764, opened 2022-10-12
Configure prod Metabase to show test data


MarsRomer:
## Description

NWBC team members need to prepare questions and dashboards to get quick insights on the intake data. 

Metabase does not provide an easy way to transfer questions and dashboards between instances, so users should use the prod instance, not test.

To enable them to prepare questions and dashboards prior to the intake closing, a solution is to:
1. create a database connection in the prod metabase that is pointing to our test database,  and
2. after the first intake closes, update the connection details to point to the prod database

## Acceptance Criteria
- [x] Given I'm logged in to Prod Metabase as a NWBC analyst or an Admin
When I visualize the data in the CCBC database
Then I see the data of the Test environment
#
- [x] Given I'm logged in to Prod Metabase as an Admin
When I visualize the data in the "CCBC (Prod)" database
Then I see the data of the Prod environment
#
- [ ] Document the step-by-step to change the CCBC database connection properties in Metabase

## Technical details

To allow metabase to connect to a different namespace's database, a `NetworkPolicy` object must be added to the helm chart. We already have a `metabase-ingress` template, a second network policy can be added to that file

- [x] A second network policy should be conditionally added when `.Values.metabase.prodIngress.enable` is `true`
- [x] Set  `.Values.metabase.prodIngress.enable` to `true` in `dev` and `test`
- [x] Set up a connection to `dev` (for initial testing, so that we don't need to cut a release to ensure that the network policy works) and `test` data in production metabase.

- [x] Production database should only be accessible by metabase admins

MarsRomer:
@matthieu-foucault do you know if it's possible for myself and the dev team to still be able to see the actual prod data in metabase if we do this ticket?  or is this a global change?  I will need to be able to see the prod data along the way

MarsRomer:
is this the one you updated?  Given I'm logged in to Prod Metabase as an Admin
When I visualize the data in the "CCBC (Prod)" database
Then I see the data of the Test environment
can we then modify it to say
'Then I see the data of the Test and Prod Environment'

matthieu-foucault:
I just edited again, there was a copy-paste typo

matthieu-foucault:
@marcellmueller an acceptance criteria is to document the steps to update the metabase connection settings. We can do this with a Loom video recording updated to our Scrum Teams channel, so that whoever is taking this on when the first intake closes can do it easily

PR #802, opened 2022-10-24
feat: add metabase prod ingress network policy

marcellmueller:
Implements #764


helm/app/values.yaml
```diff
@@ -74,6 +74,9 @@ autoscaling:
 metabase:
   instanceName: ccbc-metabase
   namespace: '' # The value must be passed in via the deploy script
+  prodNamespace: b40eb8-prod
```
matthieu-foucault, line 77: Should be an env variable

helm/app/templates/networkPolicies/metabase-ingress.yaml
```diff
@@ -16,3 +16,24 @@ spec:
...
         matchLabels:
           app.kubernetes.io/instance: {{ .Values.metabase.instanceName }}
+
+{{- if .Values.metabase.prodIngress.enable }}
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: {{ .Release.Name }}-metabase-prod-ingress
+  labels:
+{{ include "ccbc.labels" . | nindent 4 }}
```
matthieu-foucault, line 26:
```suggestion
  labels: {{ include "ccbc.labels" . | nindent 4 }}
```
The point of `nindent` is that you can define it on the previous line. This way you don't need to break the indentation in your template, and it's more readable

.github/actions/app/action.yaml
```diff
@@ -14,6 +14,9 @@ inputs:
   openshift_metabase_namespace:
     description: 'The metabase namespace being deployed to'
     required: true
+  openshift_metabase_prod_namespace:
+    description: 'The metabase prod namespace allowing us to connect dev and test databases'
```
matthieu-foucault, line 18:
```suggestion
    description: 'The metabase prod namespace that is allowed to connect dev and test databases'
```

matthieu-foucault (approved):
Looks good! Just some nitpicking items

PR #809, opened 2022-10-24
697 - Returning to submit page at a later date changes the date to the previous date

marcellmueller:
This bug was from when we moved the form_data to it's own table. Just had to use the `updated_at` column from `form_data` instead of the `application` table one.

### DIFF
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -104,10 +104,10 @@ const ApplicationForm: React.FC<Props> = ({
         formData {
           jsonData
           isEditable
+          updatedAt
           id
         }
         status
-        updatedAt
         intakeByIntakeId {
           closeTimestamp
         }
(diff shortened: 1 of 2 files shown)


app/components/Form/ApplicationForm.tsx
```diff
@@ -275,6 +274,8 @@ const ApplicationForm: React.FC<Props> = ({
       newFormData[sectionName] = { ...calculatedSectionData };
     }
 
+    console.log(updatedAt, new Date(updatedAt));
+
```
matthieu-foucault, line 278:
```suggestion
```

matthieu-foucault (commented):
Do we want a test with this bugfix, to make sure the bug doesn't come back next time there's a refactor?

matthieu-foucault:
@marcellmueller , actually looks like this was tested, and the test data needs updating

marcellmueller:
> Do we want a test with this bugfix, to make sure the bug doesn't come back next time there's a refactor?

Yeah I thought about that, thanks for pointing out the current tests. I just had to update all the payloads and move the updatedAt fields to the formData objects.

matthieu-foucault (approved):

Issue #697, opened 2022-09-23
Returning to submit page at a later date changes the date to the previous date


elliottSlee:
![Screen Shot 2022-09-23 at 12.24.17 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/5a10d8bd-19a1-4d2c-b7a4-3388b0b4dd25)

marcellmueller:
@MarsRomer @elliottSlee @matthieu-foucault not sure if this is actually a bug. Looking at the code I remember that I was asked to use the `updated_at` field to auto fill the submission fields. 

If a user returns to the submission page and nothing has changed in the form at all then the date will be the last `update_at` date. If they change their name or title on that page or update anything anywhere else will trigger an autosave and the date will update to the current date.

If we want another behaviour then let me know but this current behaviour was intentional.

MarsRomer:
You are correct.  But it's not behaving that way.  When I change the name or title, the submitted date stays the same

marcellmueller:
Are you doing this with a draft or submitted application? Looking at the code we only update it for draft applications. It might be a holdover from before we decided to allow editing submitted applications.

marcellmueller:
I guess the way to fix this is to use the new `is_editable` field we have to update the date.

MarsRomer:
draft

MarsRomer:
if it's submitted application then that date should not change no matter what

marcellmueller:
For sure, thanks for your help. I actually think that our `update_at` field isn't updating properly in the database.

marcellmueller:
Fixed and going into review, this was a small bug from when we moved form data into it's own table. We needed to use that `updated_at` column instead of the application tables.

PR #818, opened 2022-10-25
fix: setting the submission page fields is done by a custom mutation

matthieu-foucault:
final :crossed_fingers:  fix for #697


VB-at-Bis (approved):
Looks good

Issue #819, opened 2022-10-26
Remove email uniqueness constraint


MarsRomer:
## Description

If a user logs in with two different identity providers, the user creation will fail on the second login, leading to missing created_by/updated_by foreign keys.

Impact:
 - missing data: created_by/updated_by FKs are not set
 - after log in, the user gets redirected to `applicantportal` instead of `applicantportal/dashboard`

## Acceptance criteria

- [x] Users can log in with two different IdPs, with the same email address configured, and no error will be thrown on the backend

MarsRomer:
Database is not recording the ID for the business bceid

MarsRomer:
Is the best way for me to test this by creating two applications in prod?  each using a unique basic bceid that is tied to the same email?

matthieu-foucault:
> creating two applications in prod

You mean in "test", right?

Just logging in should be enough to see the fix. You should be able to see multiple users with the same email in the `ccbc_user` table

MarsRomer:
yes

MarsRomer:
metabase wont load the Form Data table in Test just now

MarsRomer:
looks like something is broken as Test Metabase is still not loading any data

MarsRomer:
I don't see my applications that all belong to the same email in my dashboard as expected
![image.png](https://images.zenhubusercontent.com/62686bda7385b9a511da26e2/45f513fd-5cc6-467f-9fef-393640bf95fc)
![image.png](https://images.zenhubusercontent.com/62686bda7385b9a511da26e2/f6744147-5655-48bc-a205-7477e4518c9a)

matthieu-foucault:
> as expected

Was seeing application for both accounts expected? 

BCeID identifies user with a User Id, not email.

Our applications do not belong to an email, but to a BCeID account, if you're expecting a different behaviour, then that's a different issue.

MarsRomer:
I must have misunderstood how to test this then.  I am to check in metabase then to see if the bug is fixed? I thought I was told to check the dashboard

matthieu-foucault:
As mentioned above

> Just logging in should be enough to see the fix. You should be able to see multiple users with the same email in the ccbc_user table

Looking at the test data, I can indeed see multiple users with your email

PR #820, opened 2022-10-26
fix: ccbc_user.email should not be unique

matthieu-foucault:
Implements #819

### DIFF
--- a/db/revert/tables/ccbc_user_001_no_unique_email.sql
+++ b/db/revert/tables/ccbc_user_001_no_unique_email.sql
@@ -1,0 +1,7 @@
+-- Revert ccbc:tables/ccbc_user_001_no_unique_email from pg
+
+begin;
+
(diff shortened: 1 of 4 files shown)


db/revert/tables/ccbc_user_001_no_unique_email.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ccbc:tables/ccbc_user_001_no_unique_email from pg
+
+begin;
+
+create unique index ccbc_user_email_address on ccbc_public.ccbc_user(email_address);
```
AntBush, line 5: Wouldn't this, on revert, fail if someone has two accounts with the same email?

matthieu-foucault: Possibly, but reverts are for local development only

AntBush (approved):
Only concern has been addressed, assuming tests pass all good

Issue #32, opened 2022-04-05
Application - Details


eazeredo:

elliottSlee:
@eazeredo what is this card referring to?

eazeredo:
This is the page that both applicant and analyst will see the application when they cannot edit it.

--
Eduardo Azeredo

Sent via mobile device
________________________________

elliottSlee:
#116 is about the review page that applicants can see. Since this one is for analysts, should we make it for release 2?

eazeredo:
@elliottSlee Moved to Release 2.

marcellmueller:
@MarsRomer I just made sure this is working in test. Looks like there are 3 received applications with ids being: 5, 10 and 13.

So they can be viewed by logging in as analyst and going to routes:

/analyst/application/5
/analyst/application/10
/analyst/application/13

elliottSlee:
@marcellmueller Is it a much much work to have the final part of the URL be the CCBC ID so it's more intuitive for users?

PR #821, opened 2022-10-26
32 analyst application view

marcellmueller:
Implements #32

Requires an application with received status in the database which can then be viewed by going to route `/analyst/application/[applicationId]`


app/components/Review/fields/ArrayBooleanField.tsx
```diff
@@ -0,0 +1,38 @@
...
+              </StyledColRight>
+            )}
+          </tr>
+        );
+      })}
+    </>
+  );
+};
+
+export default ArrayBooleanField;
```
marcellmueller, line 38: @matthieu-foucault I was unsure if this was the correct way to go about this. For multiple reasons the acknowledgements page wasn't rendering correctly so I made it's own field type here. Also the formData returns a list of enums so even if it was rendering correctly the questions and answers would have been the same.

app/components/Review/fields/ArrayBooleanField.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+// Field to display a list of yes/no values instead of the enum for array type fields
+const ArrayBooleanField: React.FC<FieldProps> = ({
+  id,
+  formData,
+  rawErrors,
+  schema,
+}) => {
+  const fieldSchema = schema as any;
+  const enumValues = fieldSchema.items.enum;
```
matthieu-foucault, line 13:
```suggestion
import { JSONSchema7 } from 'json-schema';
import React from 'react';
import { StyledColLeft, StyledColRight, StyledColError } from '../index';

// Field to display a list of yes/no values instead of the enum for array type fields
const ArrayBooleanField: React.FC<FieldProps> = ({
  id,
  formData,
  rawErrors,
  schema,
}) => {
  const enumValues = (schema.items as JSONSchema7).enum as string[];
```

if we want to avoid an `any`

app/components/Review/fields/ReviewCheckboxField.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { FieldProps } from '@rjsf/core';
+import { StyledColLeft, StyledColRight, StyledColError } from '../index';
+
+// Field to display a list of yes/no values instead of the enum for array type fields
```
matthieu-foucault, line 5: Copy-pasted comment?

app/components/Review/fields/ArrayBooleanField.tsx
```diff
@@ -0,0 +1,38 @@
+import { FieldProps } from '@rjsf/core';
+import React from 'react';
+import { StyledColLeft, StyledColRight, StyledColError } from '../index';
+
+// Field to display a list of yes/no values instead of the enum for array type fields
```
matthieu-foucault, line 5:
```suggestion
/**
Field to display a list of yes/no values instead of the enum for array type fields
*/
```
consider using jsdoc so that VSCode displays the comment when hovering on the name

app/pages/analyst/application/[applicationId].tsx
```diff
@@ -0,0 +1,69 @@
...
+import { withRelay, RelayProps } from 'relay-nextjs';
+import { graphql } from 'react-relay';
+import { usePreloadedQuery } from 'react-relay/hooks';
+import FormBase from 'components/Form/FormBase';
+import { schema as fullSchema, analystUiSchema, validate } from 'formSchema';
+import defaultRelayOptions from '../../../lib/relay/withRelayOptions';
+import FormDiv from '../../../components/FormDiv';
+import Layout from '../../../components/Layout';
+import { ApplicationIdQuery } from '../../../__generated__/ApplicationIdQuery.graphql';
+import ReviewTheme from '../../../components/Review/ReviewTheme';
```
Comment (matthieu-foucault) [inline app/pages/analyst/application/[applicationId].tsx:11]:
```suggestion
import defaultRelayOptions from 'lib/relay/withRelayOptions';
import FormDiv from 'components/FormDiv';
import Layout from 'components/Layout';
import { ApplicationIdQuery } from '__generated__/ApplicationIdQuery.graphql';
import ReviewTheme from 'components/Review/ReviewTheme';
```

app/pages/analyst/application/[applicationId].tsx
```diff
@@ -0,0 +1,69 @@
...
+        <FormBase
+          theme={ReviewTheme}
+          schema={fullSchema}
+          uiSchema={analystUiSchema as any}
+          formContext={{
+            // validate errors and pass through formContext for review checkbox section
+            errors: formErrorSchema,
+          }}
+          formData={jsonData}
+          liveValidate
```
Comment (matthieu-foucault) [inline app/pages/analyst/application/[applicationId].tsx:51]:
```suggestion
```
I don't think this has any effect since the data never changes

app/tests/pages/analyst/application/[applicationId].test.tsx
```diff
@@ -0,0 +1,697 @@
...
+        .closest('tr')
+        .getElementsByTagName('td')[0]
+    ).toHaveTextContent('Yes');
+
+    expect(
+      section
+        .getAllByText(acknowledgementsEnum[16])[0]
+        .closest('tr')
+        .getElementsByTagName('td')[0]
+    ).toHaveTextContent('Yes');
```
Comment (matthieu-foucault) [inline app/tests/pages/analyst/application/[applicationId].test.tsx:661]: this can be a loop

app/tests/pages/analyst/application/[applicationId].test.tsx
```diff
@@ -0,0 +1,697 @@
...
+  it('displays the correct title', () => {
+    pageTestingHelper.loadQuery();
+    pageTestingHelper.renderPage();
+
+    expect(
+      screen.getByRole('heading', { name: 'Application' })
+    ).toBeInTheDocument();
+  });
+
+  it('should have correct heading styles', () => {
```
Comment (matthieu-foucault) [inline app/tests/pages/analyst/application/[applicationId].test.tsx:73]: Looks like there's a lot of duplication (about 450 lines of test code?) with the test on the applicant's side. Any way we can refactor these tests so that we avoid the duplication and are consistent in what we test for both pages?

app/tests/utils/mockFormData.ts
```diff
@@ -211,6 +213,33 @@ const mockFormData = {
...
+      acknowledgementsEnum[8],
+      acknowledgementsEnum[9],
+      acknowledgementsEnum[10],
+      acknowledgementsEnum[11],
+      acknowledgementsEnum[12],
+      acknowledgementsEnum[13],
+      acknowledgementsEnum[14],
+      acknowledgementsEnum[15],
+      acknowledgementsEnum[16],
+    ],
```
matthieu-foucault, line 235:
```suggestion
    acknowledgementsList: acknowledgementsEnum,
```
why not simply using the whole array?

matthieu-foucault (changes requested):
Very minor stylistic comments on the production code side, looks great otherwise.
We probably want to do some refactoring in our tests here, rather than creating another gigantic test file.

I would love to see some happo screenshot so that we can be confident that this works and looks as expected, without reverting to manual testing

app/components/Review/fields/ReviewCheckboxField.tsx
marcellmueller, line 5:
```suggestion
```

app/tests/pages/analyst/application/[applicationId].test.tsx
Comment (marcellmueller) [reply app/tests/pages/analyst/application/[applicationId].test.tsx]: Good call on this one.

app/tests/pages/analyst/application/[applicationId].test.tsx
Comment (marcellmueller) [reply app/tests/pages/analyst/application/[applicationId].test.tsx:73]: Okay I've made all of the requested changes except for this (I could also add a happo screenshot once that PR gets merged). It might require a bit of work as this test uses `pageTestingHelper` while `ApplicationForm` tests use `componentTestingHelper`.

Happy to discuss this after standup about how to go about reusing some of the tests.

Comment (marcellmueller) [reply app/tests/pages/analyst/application/[applicationId].test.tsx]: @matthieu-foucault I moved shared tests to `/tests/components/ReviewTheme.ts` which is an exported function. Let me know what you think of it.

app/tests/components/Review/ReviewTheme.ts
```diff
@@ -0,0 +1,430 @@
...
+
+  it('should have the correct field and value in the Organization profile section', () => {
+    renderTest(mockQueryPayloadWithFormData);
+
+    expect(document.getElementById('bandNumber-error')).toBeNull();
+  });
+};
+
+// eslint-disable-next-line jest/no-export
+export default sharedReviewThemeTests;
```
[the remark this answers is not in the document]
marcellmueller: @matthieu-foucault Not sure what to do about this. Not that exporting from tests is something we will do regularly but I'm not sure I see the benefit to this rule.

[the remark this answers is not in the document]
matthieu-foucault: Figuring out why our eslint check ignores the `// eslint-disable-next-line` comments would be great.
The rule can be useful to avoid running tests multiple times, but it's a false positive here.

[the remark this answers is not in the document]
marcellmueller: You want me to debug that as part of this ticket?

app/.sentryclirc
```diff
@@ -0,0 +1,2 @@
+[auth]
+token=b3814657fe20451680d5193f67bfa28ab3dd8281882e42c2b2e76ff35c3b1ec6
```
matthieu-foucault, line 2: This needs to be removed (and an issue opened to update the detect-secrets configuration?)

marcellmueller: Done thanks.

app/tests/components/Review/ReviewTheme.ts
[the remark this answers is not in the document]
matthieu-foucault: A separate PR would be better, to avoid adding more noise to this one

matthieu-foucault (approved):
Happy to merge the PR. The issue should stay in progress until cypress tests and happo snapshots are written though

marcellmueller:
> Happy to merge the PR. The issue should stay in progress until cypress tests and happo snapshots are written though

Mocking auth login for new roles and Cypress testing should have been it's own issue/ticket as you are spending a significant amount of time on it. So either this is finished or it shouldn't have been started yet given it's blocked. In this case I'd argue in this case that happo screenshot testing should be it's own ticket given the issues you've run into.

matthieu-foucault:
Making cypress tests into their own ticket leads to the work never being prioritized, as we've seen with the file uploads. This should be part of our Definition of Done. I'm not confident this page is and will remain fully functional, and looks as expected, unless there's a cypress test.

PR #825, opened 2022-10-26
fix: FileWidget cannot call two mutations concurrently

### LINKED ISSUE #807: Duplicate file uploads
Attachment upload of file was duplicated

matthieu-foucault:
Potential fix for #807.

I managed to confirm that the mutation was sent twice, resulting in double uploads, but I am unable to manually reproduce the bug.

This ensures that `handleChange` will not trigger another mutation if the previous one is still in flight.


AntBush (approved):
Looks good to me, I'm wondering if there's a way we can test this using cypress to try and trigger the file upload twice and see if the request happens twice? 

Also sonarcloud is complaining about the lack of tests on the error state

Issue #771, opened 2022-10-14
Write view for Metabase use


MarsRomer:
As an analyst 
I want to be able to see all the form data in one table (view) 
so that it is easy to query and I don't need to update my queries when there are changes to the database

### Acceptance Criteria

-  [x] All form data is available in metabase, with each form field parsed out in separated columns
 - [x] View should return the name of uploaded fields
-  [x] view should show current application status
- [x] only show applications that do not have 'Draft' status (i.e. has `submitted` in its `application_status` records)
- [x] view is tested

Note:  Ali has already written scripts to parse the data into separate columns so please connect with him if you need to use his work.

matthieu-foucault:
> only show applications that have 'Received' status

@MarsRomer , I thought we had discussed using metabase for filtering instead. Did you forget to update the acceptance criteria?

VB-at-Bis:
Talked to Ali. Example of the report - https://ccbc-metabase-test.apps.silver.devops.gov.bc.ca/question/77-parsedformdata-oct13

MarsRomer:
hey @VB-at-Bis what is the name of the view?  Im looking for it in Metabase

VB-at-Bis:
> 

summary_view :)

VB-at-Bis:
Was it deployed to test? I created question in Metabase https://ccbc-metabase-test.apps.silver.devops.gov.bc.ca/question/81-summary-view but it fails to load ..

VB-at-Bis:
View did not work as expected due to Postgresql rule that `Access to tables referenced in the view is determined by permissions of the view owner`. View is created under `ccbc` user. Changing the owner of the view or changing permissions of the `ccbc` user to bypass RLS would mean opening security hole and effectively disabling row-level security.

Thus, alternative approach was implemented (using Metabase Model) - see https://ccbc-metabase.apps.silver.devops.gov.bc.ca/model/36-all-application-data 
Using Model provides same user experience as using Table or View (foreign key relationships, filters etc.)

P.S. ticket https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/828 was created to remove unneeded `summary_view` code.

MarsRomer:
hey @VB-at-Bis  thank you for this, is it possible to add the column that holds the 'state' of the application, i.e. Submitted, Draft etc?  I don't think I see it in this model

[...]

PR #815, opened 2022-10-25
feat: analyst dashboard backend

AntBush:
Implements # 639 backend

PR #829, opened 2022-10-27
fix: grant public to computed column function

VB-at-Bis:
Implements #771


marcellmueller (commented):
This is failing the `check_immutable_sqitch_files` test. I believe you need to do either a `sqitch add` or `sqitch rework`.

db/deploy/views/summary_view.sql
```diff
@@ -138,6 +138,7 @@ FROM ccbc_public.application a
     comment on column ccbc_public.summary_view.archived_at is 'archived at timestamp';
     comment on column ccbc_public.summary_view.intake_id is 'Application Intake Number, used as the prefix for CCBC reference number';
 
+    grant execute on function ccbc_public.application_status to ccbc_auth_user;
```
matthieu-foucault, line 141:
```suggestion
    grant execute on function ccbc_public.application_status to public;
```
Should execute be granted to `public`?  Metabase doesn't use `ccbc_auth_user`

Additionally, can you add a pgTap test, using https://pgtap.org/documentation.html#function_privs_are?

db/deploy/computed_columns/application_status@1.11.0.sql
```diff
@@ -0,0 +1,17 @@
...
+select status from ccbc_public.application_status
+    where application_id= application.id
+    order by id desc limit 1;
+$$ language sql stable;
+
+grant execute on function ccbc_public.application_status to public;
+
+comment on function ccbc_public.application_status is 'Computed column to return status of an application';
+
+commit;
```
marcellmueller, line 17: When you do a `sqitch rework` this file tagged `@1.11.0` is supposed to be the same as the original file and any changes happen in the original `computed_columns/application_status` file. 

Yes it's confusing and I did the same thing before. I think the tag is to let us know which release our app was on as a history for the file.

You can see this by looking at other @ tagged files in the project, they are older versions of the original.

VB-at-Bis: hmm.. I understand it as `changes were made to the application after it was released with version 1.11.0 ` .
So I believe name is correct

VB-at-Bis:
```Commit: 9f6bbbc8c058e721e2b5de6a529baff564d034b9 [9f6bbbc]
Parents: 6026d29853
Author: marcellmueller <marcel@button.is>
Date: October 26, 2022 12:55:07 PM
Committer: marcellmueller
chore: release v1.11.0

Comment (marcellmueller) [reply db/deploy/computed_columns/application_status@1.11.0.sql:17]: The issue isn't the name/release version, it's that the changes have to happen in the original `application_status` file as you can see by looking at other tagged release files (their contents are older).

**Review by matthieu-foucault — CHANGES_REQUESTED** [review]
I'd recommend squashing the commits and undoing every change other than the following two files:
- db/deploy/computed_columns/application_status.sql
- db/test/unit/computed_columns/application_status_test.sql

**Review by marcellmueller — APPROVED** [review]

File: db/sqitch.plan [hunk 3]
```diff
@@ -73,8 +73,8 @@ tables/ccbc_user_001_no_unique_email 2022-10-26T19:00:26Z Matthieu Foucault <mat
 tables/application_001_add_analyst_permission 2022-10-24T14:56:56Z Anthony Bushara <anthony@button.is> # Adds the analyst permissions and RLS for the application table
 tables/application_status_001_give_read_and_insert_to_ccbc_analyst 2022-10-24T17:29:23Z Anthony Bushara <anthony@button.is> # Gives the select and insert permission to ccbc_analyst
 computed_columns/application_project_name [computed_columns/application_project_name@1.9.0] 2022-10-24T19:27:49Z Anthony Bushara <anthony@button.is> # Add ccbc_analyst to application_project_name computed column
-computed_columns/application_status [computed_columns/application_status@1.9.0] 2022-10-24T19:30:15Z Anthony Bushara <anthony@button.is> # Add ccbc_analyst to application_status computed column
```
matthieu-foucault, line 76: Why was the sqitch plan changed here and a rework done? This change is already reworked in this release, it doesn't need to be reworked again

marcellmueller: I pointed out that it was failing the `check_immutable_sqitch_files` and assumed it needed a sqitch add or rework.

matthieu-foucault: Looks like this change was added in #815, which was merged after this PR was initially created.

matthieu-foucault: So we have a proper conflict here, where two PRs made similar changes: reworking `computed_columns/application_status`. This PR can now modify `computed_columns/application_status.sql` directly without altering the sqitch plan, or creating the `@1.11.0.sql` files

VB-at-Bis: hmm.. #815 added `computed_columns/application_status [computed_columns/application_status@1.9.0]` so it's ok for current PR to add `computed_columns/application_status [computed_columns/application_status@1.11.0]` files. Otherwise we'll get  check_immutable_sqitch_files  error

matthieu-foucault: `computed_columns/application_status` isn't immutable right now since it's been reworked in #815 and the reworked change is unreleased. You shouldn't get the  `check_immutable_sqitch_files` error.

matthieu-foucault (approved):
Thanks for making the changes

PR #832, opened 2022-10-27
Fix: OWASP Zap CI timeout

marcellmueller:
I'm not sure why this wasn't failing when we merged in the PR with the new redirects though I assume that Github might just be running a bit slow today. Seems that this is caused by the target being set to `http://localhost:3000/` and the redirect to `/applicantportal` causing it to time out most of the time.

Passing twice now: 
https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3339970295/jobs/5529555075
https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3340019510/jobs/5529662044


VB-at-Bis (approved):
Looks good

PR #833, opened 2022-10-27
Fix/metabase network policy

marcellmueller:
Fix for the new metabase network policy breaking connections to dev and test databases through their respective metabase instance.


marcellmueller:
@matthieu-foucault I've confirmed that the dev metabase is working again as well as tested connecting it to the prod metabase.

matthieu-foucault (approved):

matthieu-foucault:
@marcellmueller just needs a rebase

Issue #639, opened 2022-09-13
Table of applications (Analyst Dashboard)


zoeyli-46:
## User story
As an analyst, I want to see all applications in a dashboard so that I can do my job of managing applications.

## Acceptance criteria
- [ ] Given that an analyst is logged in
When they are on the dashboard page
Then they will see a table with the columns CCBC ID, Status, Project title, Organization, Lead, Package
And the rows will be applications
#
- [ ] <s>feature flag - instead of displaying dashboard, have text 'Coming soon'</s>
#
- [ ] Analysts can see any application that is or has been "Received"
#
- [ ] Given that an analyst is on the dashboard page
and the project table is visible
when received applications exist
then the list is ordered by CCBC ID by default, ascending alphabetically

## Not included 
- Lead & Package columns will be empty for now

## Technical details

- CIF has a table that could be reused, see https://github.com/bcgov/cas-cif/tree/develop/app/components/Table
- An RLS policy should be added: analysts can select applications (and its data), as long as the application has "received" in its statuses.
- Consider creating some test applications in the dev data folder, so that developers can easily test the feature.

## Design
![image.png](https://images.zenhubusercontent.com/61e8875b1de5eff8f072cd04/d2732dfb-272f-49ad-8c56-570f06008f6d)
#### Component states/other behaviours 
https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=7039%3A31260

AntBush:
So, to be clear, for this card there is only the infinite scroll for the table and the display of the applications. The search/sort features are not part of this card?

zoeyli-46:
Search sort filter are in a different card! #647 #641 #640 #642 are all tickets related to dashboard and filters. is there anything we should be putting into further separate cards or merging?

PR #837, opened 2022-10-27
test: server and cypress support mock logins with different roles

matthieu-foucault:
Required for e2e tests in #32 and #639


marcellmueller (approved):
Thanks for doing this, looks great. 

Do you think a small section in the readme explaining how this works would be helpful?

marcellmueller (changes requested):
Oh it looks like the test is failing CI

docs/auth.md
```diff
@@ -1,3 +1,23 @@
+# Authentication and Authorization
```
marcellmueller, line 1: Thanks for this :heart:

marcellmueller (approved):
Nice job and I appreciate the readme and eliminating a lot of the `cy.wait` throughout the tests.

I noticed a few in some tests still such as `upload.spec.js`, I assume those are still needed? `cy.contains` doesn't help in every scenario?

matthieu-foucault:
> I noticed a few in some tests still such as upload.spec.js, I assume those are still needed? cy.contains doesn't help in every scenario?

They may not be needed anymore. I did a search and replace of the `cy.get().contains(` pattern, so I didn't look into the `upload.spec.js` too much

Some `wait`s on the initial page load are still needed, as we have "detached DOM" issues otherwise, caused by the may renders that occur on page load

PR #838, opened 2022-10-28
406 implement sentry logging

marcellmueller:
Implements #406


marcellmueller:
@matthieu-foucault do you have any good ideas about how I can test `graphql/index.ts` so I can pass the Sonarcloud scan? Thanks.

matthieu-foucault:
A test asserting that the `postgraphile` function is called with the expected `postgraphileOptions` would be good. This would be somewhat similar to the test I just wrote for `authenticationPgSettings`, where you mock the config with different values (`NODE_ENV` production vs development, `SENTRY_ENVIRONMENT` is defined or not)

app/sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=bcgov-ccbc
+defaults.project=ccbc
+cli.executable=../../../.npm/_npx/a8388072043b4cbc/node_modules/@sentry/cli/bin/sentry-cli
```
matthieu-foucault, line 4: This last config item looks specific to your local setup, we probably want to remove that

app/tests/utils/isRouteAuthorized.test.ts
```diff
@@ -23,6 +23,13 @@ describe('The isRouteAuthorized function', () => {
     expect(isRouteAuthorized('/analyst/request-access', '')).toBe(true);
   });
 
+  it('allows unauthenticated users to access the /error-500 route', () => {
+    expect(isRouteAuthorized('/analyst/request-access', 'ccbc_guest')).toBe(
+      true
+    );
+    expect(isRouteAuthorized('/analyst/request-access', '')).toBe(true);
```
matthieu-foucault, line 30:
```suggestion
    expect(isRouteAuthorized('/error-500', 'ccbc_guest')).toBe(
      true
    );
    expect(isRouteAuthorized('/error-500', '')).toBe(true);
```

app/pages/error-500.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { error500Query } from '__generated__/error500Query.graphql';
+
+const getError500Query = graphql`
+  query error500Query {
+    session {
+      sub
+    }
+  }
+`;
+const Error500Page = ({
```
matthieu-foucault, line 17: You may want to consider making this into a static `pages/500.tsx` page that doesn't use a relay query (and doesn't pass the `session` to `Layout`).
It would be rendered for every 500 error and replace the generic nextjs error page.

matthieu-foucault (changes requested):
The only required change is the copy-paste error in `app/tests/utils/isRouteAuthorized.test.ts`

app/sentry.properties
marcellmueller: > A test asserting that the `postgraphile` function is called with the expected `postgraphileOptions` would be good. This would be somewhat similar to the test I just wrote for `authenticationPgSettings`, where you mock the config with different values (`NODE_ENV` production vs development, `SENTRY_ENVIRONMENT` is defined or not)

I think I might need some help with that if you have time at some point.

app/pages/error-500.tsx
marcellmueller: I considered that, I will have to make a new static layout that doesn't have a login link if there is no session as it will be confusing if there is an error for a logged in user.

app/tests/utils/isRouteAuthorized.test.ts
```diff
@@ -23,6 +23,11 @@ describe('The isRouteAuthorized function', () => {
     expect(isRouteAuthorized('/analyst/request-access', '')).toBe(true);
   });
 
+  it('allows unauthenticated users to access the /error-500 route', () => {
+    expect(isRouteAuthorized('/error-500', 'ccbc_guest')).toBe(true);
+    expect(isRouteAuthorized('/error-500', '')).toBe(true);
+  });
```
matthieu-foucault, line 29:
```suggestion
  it('allows unauthenticated users to access the /500 route', () => {
    expect(isRouteAuthorized('/500', 'ccbc_guest')).toBe(true);
    expect(isRouteAuthorized('/500', '')).toBe(true);
  });
```

app/utils/isRouteAuthorized.ts
```diff
@@ -7,6 +7,7 @@ const pagesAuthorization = [
       '/analyst',
       '/applicantportal',
       '/analyst/request-access',
+      '/error-500',
```
matthieu-foucault, line 10:
```suggestion
      '/500',
```

matthieu-foucault: I don't think this is needed actually, given that our 500 page doesn't use `withRelayOptions` anymore

marcellmueller: Nice catch, I've removed it altogether here:

https://github.com/bcgov/CONN-CCBC-portal/pull/838/commits/a00f745608bf1b6bb66c3c69540c8cfa9d4693af

matthieu-foucault (approved):
:shipit:

Issue #828, opened 2022-10-27
Remove summary_view 


matthieu-foucault:
`summary_view`  was added as a part of  the ticket [#771](https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/bcgov/conn-ccbc-portal/771)

However, view did not work as expected due to Postgresql rule that `Access to tables referenced in the view is determined by permissions of the view owner.` View is created under `ccbc` user. Changing the owner of the view or changing permissions of the `ccbc` user to bypass RLS would mean opening security hole and effectively disabling row-level security.

Thus, alternative approach was implemented (using Metabase Model) and thus, code that adds `summary_view` in Postgres DB should be removed.

MarsRomer:
I can still see them in metabase: ![image.png](https://images.zenhubusercontent.com/62686bda7385b9a511da26e2/352353f4-df57-4a4a-8ac0-20e305d67681)

VB-at-Bis:
Ok those views are no longer listed in Metabase

PR #839, opened 2022-10-28
feat: remove summary_view and tests

VB-at-Bis:
Implements #828


db/sqitch.plan
```diff
@@ -66,7 +66,6 @@ ccbc_job_executor_permissions [ccbc_job_executor_permissions@1.9.0] 2022-10-21T1
 tables/ccbc_user_001_admin_analyst_permissions 2022-10-21T19:25:33Z Marcel Mueller <marcel@button.is> # grant permissions on table ccbc_user to ccbc_admin and ccbc_analyst
 mutations/create_user_from_session [mutations/create_user_from_session@1.9.0] 2022-10-21T19:27:58Z Marcel Mueller <marcel@button.is> # grant execute on function to ccbc_admin and ccbc_analyst
 @1.10.0 2022-10-24T23:29:33Z Matthieu Foucault <matthieu@button.is> # release v1.10.0
-views/summary_view 2022-10-24T20:16:50Z ,,, Vladimir <vladimir@button.is> # feat: view for metabase users
```
matthieu-foucault, line 69: you can't remove things from the sqitch plan that have been deployed. Unfortunately this code has to stay in our repo now.
You have to rework the view, and in the reworked version, run `drop view if exists ccbc_public.summary_view`

VB-at-Bis: Thanks, I already realized that :)

db/sqitch.plan
```diff
@@ -78,3 +78,6 @@ tables/application_form_data_001_add_rls_for_analyst 2022-10-24T19:51:34Z Anthon
 tables/form_data_002_add_permisions_for_analyst 2022-10-24T20:30:23Z Anthony Bushara <anthony@button.is> # Add permissions for analyst
 computed_columns/application_organization_name 2022-10-25T15:05:18Z Anthony Bushara <anthony@button.is> # computed column to display the organization name for analysts
 computed_columns/application_form_data [computed_columns/application_form_data@1.9.0] 2022-10-25T18:34:11Z Anthony Bushara <anthony@button.is> # Add ccbc_analyst permission
+@1.12.0 2022-10-28T18:19:06Z ,,, Vladimir <vladimir@button.is> # rollback summary_view changes
```
matthieu-foucault, line 81: why is there a release in here?

For this PR you can update `computed_columns/application_status.sql` directly since it's still unreleased, and you can do a rework of `views/summary_view`. `views/summary_view.sql` is where you'd drop the view. You should not edit the reworked deploy files (i.e. any file ending with `@1.*.*.sql`)

VB-at-Bis: Ok, I will try to update computed_columns/application_status.sql directly. I believe I already tried and got errors from sqitch .. but maybe those were caused by some other my mistake. :)

matthieu-foucault (approved):

PR #847, opened 2022-11-01
760 navigation sidebar

marcellmueller:
Implements # 760

A few other things I did in this PR:
- added basic pages for `/analyst/application/[applicationId]/assessments` and `/analyst/application/[applicationId]/history` so that the navigation URLs will work.
- added basic tests for those pages (Sonarcloud seems happy)
- added a reusable `AnalystLayout` for `/analyst/application` subpages as they all use the same layout.


app/components/Analyst/NavItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+`;
+
+const NavItem: React.FC<Props> = ({ currentPath, href, icon, label }) => {
+  const selectedColour = currentPath === href ? '#F1F2F3' : 'inherit';
+
+  return (
+    <StyledNavItem selected={selectedColour}>
+      <StyledIconContainer>
+        <FontAwesomeIcon icon={icon} fixedWidth aria-hidden="true" />
```
AntBush, line 53: When the link is hidden, if you click on an icon nothing happens. If you add another <Link> around here that should remedy the issue

marcellmueller: Fixed. Nice catch, really appreciate that :bow:

AntBush (approved):
LGTM!

PR #861, opened 2022-11-03
fix: deploy mocks schema before dev/test data

matthieu-foucault:


marcellmueller (approved):

AntBush:
I think this is missing an additional fix to the 001_received_applications.sql. I'm going to push a commit

PR #879, opened 2022-11-04
Fix permissions in dev/test deployment scripts

matthieu-foucault:
- run pre-upgrade job with postgres command
- set role to table owner before inserting dev/test intakes


marcellmueller (approved):

VB-at-Bis (approved):
LGTM

Issue #328, opened 2022-07-21
Export attachments to ISED - ZIP file


MarsRomer:
### User Story
As an Analyst 
I want to download all the attachments for applications,
so that they are in one place to send to ISED

**The process**
Analyst downloads attachments from the Dashboard to their local drive.
Analyst uploads the zip file into SharePoint.  Renaming of files is handled by this API.

1st Alternate solution:  Build a secure API to download all attachments ~~(filtered as needed)~~.  API provides date that attachment was uploaded so that ISED can see when it was uploaded.  Re-naming is handled by this API.

<s>2nd Alternate: Analyst or admin clicks a 'download for ISED' button and the attachments automatically download into a Sharepoint site below

NWBC and ISED  have a Teams channel where files can be stored on the Sharepoint.  Max storage in this space is 200GB of which 199.95GB are still available.  
https://bcgov.sharepoint.com/teams/053702/Shared%20Documents/Forms/AllItems.aspx?FolderCTID=0x012000CA34C709521FD04A89C9C95DF25451BE&id=%2Fteams%2F053702%2FShared%20Documents%2FGeneral%2FCCBC%5FApplications&viewid=f1e5a84d%2D914a%2D4255%2Db7b4%2Db0a5b94b7be3
</s>
### Acceptance Criteria
- [x] **Export attachments button begins download**
Given that an analyst is on the dashboard 
when they click the Export Attachments button
Then a download will begin of all the attachments of all of the applications as a Zip file with the name "CCBC attachments - [currentdate].zip"
Example: `CCBC attachments - 20230115.zip`
#
 - [x] **Rename files**
Given that an analyst is on the dashboard 
when they click the Export Attachments button
then the filenames will have their CCBC ID as a prefix
Example: `Template 1.xlsx` -> `CCBC-010001 - Template 1.xlsx`
#
- [x] Folder Structure:
![Screen Shot 2022-11-01 at 3.35.58 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/42a9b33a-b1f1-4f82-8e74-303514595613)
#
- [x] **Cancel download**
Given that an analyst has clicked "Export attachments"
and the download has begun
when they cancel it through their browser
then the download will stop

### Wireframe
https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=7010%3A35991

[...]

PR #882, opened 2022-11-07
328 - download attachments

marcellmueller:
Implements #328


app/data/archivePaths.ts
```diff
@@ -0,0 +1,113 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
+const COVERAGE = 'Step 1 - Eligibility Mapping & Statistics';
+const TEMPLATE_UPLOADS = 'Step 2 - Templates';
+const SUPPORTING_DOCUMENTS = 'Step 4 - Supporting Documents';
+
+const archivePaths = {
```
matthieu-foucault, line 7: There's a risk here for future maintenance, which is that two pages might have fields that have the same name.

Rather than `archivePaths` being an object, you may want to consider making it an array, containing the form path for each element.
Additionally, you could use a simple string and use a `replace` to insert the ccbcId and fileName, rather than copying the same function signature for every template. 

It would look like this, which I think would make this file more concise and easier to read:

```ts
const archivePaths = [
  { formPath: 'templateUploads.eligibilityAndImpactsCalculator', archivePath: `/$id/${TEMPLATE_UPLOADS}/$id - Template 1 - $filename`},
  { formPath: 'templateUploads.detailedBudget', archivePath: `/$id/${TEMPLATE_UPLOADS}/$id - Template2 - $filename`},
]
``` 
You would need a little bit of documentation at the top indicating which variables are supported (only `$id` and `$filename` for now?).

You can use [`lodash.get`](https://www.npmjs.com/package/lodash.get) to easily retrieve the formData for any `formPath`

matthieu-foucault: Additionally, the first `/$id` could be omitted given that it's on every file

matthieu-foucault: I wasn't aware this was an intentional choice. If the application behaviour relies on this constraint on the form schema, it should be documented somewhere and potentially enforced with a static or runtime check. I'd suggest creating a technical debt card to document this restriction and the reasons behind it.

app/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "NODE_OPTIONS='--inspect' ts-node --project tsconfig.server.json server.ts",
+    "dev": "nodemon",
```
VB-at-Bis, line 6: is it intentional? I recall we discussed using `nodemon` ...

marcellmueller: Yes we snuck nodemon into this ticket since it should have been set up at the beginning of the project and I needed to restart my server a lot for this PR. This dev command got moved to `nodemon.json`.

Issue #275, opened 2022-07-04
Upload - Implement anti-virus check


eazeredo:
### Description

A ClamAV process should run an analysis of all of the attachments after they are uploaded.

The identified solution is to use ClamAV in AWS (using https://github.com/awslabs/cdk-serverless-clamscan)

Clamscan will tag files with the result of the scan. Any file that doesn't have the "CLEAN" tag will be considered quarantined (see #757)

The antivirus scanning process should not affect the user when uploading the file.

### Acceptance criteria

Given that I am uploading an attachment
When the upload completes
Then the file is analyzed with ClamAV

Given that a file was uploaded
When the ClamAV scan finishes
Then the S3 object is tagged with the result of the scan

### Learning materials

https://aws.amazon.com/blogs/devops/building-apps-with-aws-cdk/
https://aws.amazon.com/blogs/developer/cdk-pipelines-continuous-delivery-for-aws-cdk-applications/

AntBush:
Antivirus candidate: https://github.com/bcgov/clamav

wenzowski:
@MarsRomer can we check in with the common services showcase team to see if there's an instance of clamav already deployed we can share access to? Otherwise we'll have to duplicate the devops effort and deploy our own copy of the clamav appliance in order to have virus scanning

matthieu-foucault:
Looking at the deployment config, this will require 2 cores added to our requests quota if we want it to be highly available https://github.com/bcgov/clamav/blob/master/openshift/templates/clamav-dc.conf

MarsRomer:
Blocked: check with Nick if there is a clamAV instance we can use

eazeredo:
@matthieu-foucault , what information is missing or needs clarification? Can you add a comment here?

matthieu-foucault:
When do we run ClamAV?
Where do we store the results?
What do we do with infected files?
Do we need to use either https://github.com/bcgov/clamav-mirror or https://github.com/bcgov/clamav, or are neither of them needed and we should just follow the [ClamAV docs](https://docs.clamav.net/manual/Installing/Docker.html)?
What are the benefits of running ClamAV on AWS vs OpenShift?

MarsRomer:
Blocked: Find out if pathfinder team will allow this

matthieu-foucault:
Got the go-ahead from Julian: https://chat.developer.gov.bc.ca/channel/aws-how-to?msg=iGBierqHSFyanYxtK

[...]

PR #887, opened 2022-11-09
feat: added terraform for lambda deployment

VB-at-Bis:
Implements #275 

Deployment to AWS using terraform and archived lambda. Lambda archive should be generated by running `./build.sh` in https://github.com/bcgov/CONN-ClamAV-scan project


marcellmueller (approved):
Looks great! :partying_face: Nice work.

Issue #826, opened 2022-10-27
View Application Header


eazeredo:
## User Story
As an Analyst
I want to see the application information on all pages of an application (application, assessments, history)
So that I can see at any moment which application I'm working on.

## Acceptance Criteria:
- [x] **Visible on application page**
Given I'm on the dashboard
When I open an application
Then I'm redirected to the application page
And I see the Application ID, Project name, and Organization name information 

## Not included
- Application Status - #669
- Lead - #758
- Package - #824

## Technical Details
- [x] Create as it's own component with a fragment on Application (like with Application Status for Applicant page)
- [x] Use computed columns for Org Name and Proj Name

## Design
- [x] Width is fill container (100%)
- [x] 40px from the page header above
- [x] 40 px from the content below
- [x] Stroke is outside, #1A5A96, 4px
- [x] CCBC-ID & Organization name font is bold 16px
- [x] Project name font is bold 24px
- [x] 8px between elements
- [x] 12px between text elements and left edge
- [ ] if you're unsure about anything, talk to Elliott or Zoey about it

## Wireframe
![Screen Shot 2022-11-01 at 11.29.47 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/c662edc2-6ce9-4b29-ac45-b469799c37f7)

eazeredo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/826?planning-poker) with Zenhub @AntBush @marcellmueller @matthieu-foucault @VB-at-Bis

PR #890, opened 2022-11-10
826 application header

marcellmueller:
Implements #826


app/components/Analyst/ApplicationHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { ccbcNumber, organizationName, projectName } = useFragment(
+    graphql`
+      fragment ApplicationHeader_application on Application {
+        projectName
+        organizationName
+        ccbcNumber
+      }
+    `,
+    application
+  );
```
matthieu-foucault, line 36: Looks like the right way to do this 👍

VB-at-Bis (approved):
Looks great

Issue #758, opened 2022-10-11
Assign lead


elliottSlee:
### User Story
As an Analyst 
I want to assign a lead to each application 
So that we have a person to drive the application through the assessment process.

### Acceptance Criteria
- [x] **Add the `Lead` column to the dashboard**
Given that an analyst is logged in
when they are on the dashboard
then they will see a column with the header, "Lead"
#
- [x] **Assign lead in dashboard**
Given an application does not have a `Lead` assigned
When an analyst opens the dashboard
Then they will see the "Assign analyst" dropdown in the `Lead` column
When they click on "Assign analyst"
Then a dropdown will appear with the List of Analysts (#759)
and when a name is selected
Then the value is saved 
#
- [x] **Lead name shows in dashboard**
Given an application has a `Lead` assigned
When an analyst opens the dashboard
Then they will see the name of the person assigned
and an arrow indicating that it is still a dropdown
#
- [x] **Text overflow uses ellipses** 
Given that the lead has been assigned
and its width in the dashboard column is > 114px,
then text overflow will use ellipses
(so that the lead column is < 140px including the dropdown arrow and spacing)
#
- [x] **Lead in Application header**
Given that the lead has not been assigned
when an analyst is on any application page (application, assessment, or history)
then they will see the "Select lead" on the right of the header
Given that the lead has been assigned
when an analyst is on any application page (application, assessment, or history)
then they will see the name of the lead (and the dropdown arrow) on the right of the header
#
- [x] **Date and user is tracked**
Given that an analyst assigns a lead
then the date it was selected is recorded
and the user is recorded
#
- [x] **Assigned analyst is static if list of analysts changes**
Given that the eligibility assessment has been assigned to someone
when their name is removed from the list of analysts
then their name will remain as a response to "assigned to"
#
## Technical Details
1. .
1. .

#### Dashboard
![Dashboard_Lead.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/434c4c2c-f708-4676-944d-00aa67fdeca4)

#### Header
![Header_unassigned.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/256453c7-838a-40d5-a5ef-5e318428fce9)
![Header_assigned.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/4684a884-08db-4c0f-95ca-5a349260b273)

zoeyli-46:
[...]

PR #893, opened 2022-11-14
758 assign lead

marcellmueller:
Implements #758


marcellmueller:
@AntBush this one is completed and ready to go. Though as I mentioned the use of relay fragments could be improved.

db/revert/tables/analyst_001.sql.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ccbc:tables/analyst_001.sql from pg
+
+behin;
```
AntBush, line 3: spelling error here, also filename has double sql

db/revert/tables/analyst.sql
```diff
@@ -1,7 +1,45 @@
--- Revert ccbc:tables/analyst from pg
+-- Deploy ccbc:tables/analyst to pg
 
 begin;
 
-drop table ccbc_public.analyst;
+create table if not exists ccbc_public.analyst(
```
AntBush, line 5: I think this should be undone, looks like there's a rework that was done but not reverted

db/test/unit/computed_columns/application_analyst_lead_test.sql
```diff
@@ -0,0 +1,22 @@
+begin;
+
+select plan(3);
+
```
AntBush, line 4: I think a test here that tests the return value would be useful, and at least one with multiple lead changes on the same application

AntBush (changes requested):
Just a couple of changes, looks good though!

db/deploy/computed_columns/application_analyst_lead.sql
```diff
@@ -0,0 +1,19 @@
+-- Deploy ccbc:computed_columns/application_analyst_lead to pg
+
+begin;
+
+create or replace function ccbc_public.application_analyst_lead(application ccbc_public.application) returns varchar as
+$$
+select concat_ws(' ', given_name, family_name) AS name from ccbc_public.application_analyst_lead as al
```
AntBush, line 7: I was going back and forth on whether I think this should return the whole row, and I think it would be more useful in future cases to return the row if we want to use it for joins later. Up to you since this is essentially designing for more than the ticket needs

VB-at-Bis (approved):
Looks safe

VB-at-Bis (approved):
Looks good

AntBush (approved):
LGTM!

PR #900, opened 2022-11-15
chore: release v1.15.0

marcellmueller:
Never had this issue before but the last release I just did won't [check sqitch plan release and deploy to test](https://github.com/bcgov/CONN-CCBC-portal/actions/runs/3473638332/jobs/5806761668) I believe because according to the docs we need to merge the PR locally using fast forward. 

Not sure why merging from the Github GUI worked in the past, either way our readme says to:

-once the pull request is approved, merge using fast-forward, from your local git repository. Only commits that are tagged can be deployed to test and prod.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,34 @@
+# [1.15.0](https://github.com/bcgov/CONN-CCBC-portal/compare/v1.13.0...v1.15.0) (2022-11-15)
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


AntBush (approved):

AntBush:
@marcellmueller I think the problem is that the tag was no longer matching the commit previously since you rebased. If we just merge from the GUI it should work this time

PR #905, opened 2022-11-16
fix: receive-applications cronjob utc/pst schedule

marcellmueller:
Implements # \<issue number\>

### DIFF
--- a/helm/app/templates/cronJobs/receive-applications.yaml
+++ b/helm/app/templates/cronJobs/receive-applications.yaml
@@ -5,7 +5,7 @@ metadata:
   labels: {{ include "ccbc.labels" . | nindent 4 }}
 spec:
   suspend: false
-  schedule: "0 9,21 * * *" # run cronjob at At 09:00 and 21:00
+  schedule: "0 5,17 * * *" # run cronjob at At 09:00 and 21:00
   jobTemplate:
     spec:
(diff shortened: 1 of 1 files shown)


helm/app/templates/cronJobs/receive-applications.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   labels: {{ include "ccbc.labels" . | nindent 4 }}
 spec:
   suspend: false
-  schedule: "0 9,21 * * *" # run cronjob at At 09:00 and 21:00
+  schedule: "0 1,13 * * *" # run cronjob at At 09:00 and 21:00
```
VB-at-Bis, line 8: I'd expect 5 and 17  :) .. we are PST (which is UTC +8), so 9am PST means 17 UTC (https://www.google.com/search?q=9am+pst+in+utc) and 9pm PST (21) means 5am.

VB-at-Bis (changes requested):
see inline

helm/app/templates/cronJobs/receive-applications.yaml
marcellmueller: Fixed, thanks Vlad appreciate it :)

VB-at-Bis (approved):
looks good

PR #906, opened 2022-11-16
fix: s3archive response

marcellmueller:
Eslint was complaining that I didn't return anything from async function so I threw a `return res.end()` which prematurely ended the function. Also added missing filenames.

### DIFF
--- a/app/backend/lib/s3archive.ts
+++ b/app/backend/lib/s3archive.ts
@@ -95,9 +95,7 @@ s3archive.get('/api/analyst/archive', async (req, res) => {
     sortAndAppendAttachments(jsonData, ccbcId);
   });
 
-  archive.finalize();
-
-  return res.end();
+  return archive.finalize();
 });
 
 export default s3archive;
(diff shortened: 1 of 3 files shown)


app/backend/lib/s3archive.ts
```diff
@@ -95,9 +95,7 @@ s3archive.get('/api/analyst/archive', async (req, res) => {
     sortAndAppendAttachments(jsonData, ccbcId);
   });
 
-  archive.finalize();
```
AntBush, line 98: Would adding an `await` here also have solved the problem? I assume it was returning before this had finished.

AntBush (approved):
Looks good!

Unrelated to this, if we want to be less CPU intensive, we can set the archiver option `{ store: true }` to turn off compression

app/backend/lib/s3archive.ts
marcellmueller: I did try that but it would still return the res.end before the archive finalized.

marcellmueller:
> Looks good!
> 
> Unrelated to this, if we want to be less CPU intensive, we can set the archiver option `{ store: true }` to turn off compression

Cool, I will keep that in mind when we do some optimizations :+1:

PR #911, opened 2022-11-17
fix/review file uploads

marcellmueller:
- fix: remove errors for file upload fields stored as array
- test: add validate test and update mock form data

Implements # \<issue number\>

### DIFF
--- a/app/formSchema/validate.ts
+++ b/app/formSchema/validate.ts
@@ -4,9 +4,35 @@ import schema from './schema';
 const validate = (formData) => {
   const errorSchema = validateFormData(formData, schema)?.errorSchema;
 
+  const fileUploadPages = [
+    'templateUploads',
+    'supportingDocuments',
+    'coverage',
+  ];
+
+  // Remove 'should be string' errors for file upload fields which are saved as an array
+  Object.keys(errorSchema).forEach((formPage) => {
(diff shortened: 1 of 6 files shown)


app/formSchema/validate.ts
```diff
@@ -4,9 +4,35 @@ import schema from './schema';
...
   const errorSchema = validateFormData(formData, schema)?.errorSchema;
 
+  const fileUploadPages = [
+    'templateUploads',
+    'supportingDocuments',
+    'coverage',
+  ];
+
+  // Remove 'should be string' errors for file upload fields which are saved as an array
+  Object.keys(errorSchema).forEach((formPage) => {
```
AntBush, line 14: Do we need this part? I commented it out locally and the bug isn't being reproduced

marcellmueller: Yes it is needed, try commenting that out and running the `ApplicationForm` test. Without it the validation will return those pages with empty objects causing the review page checkbox and alert section to show incorrectly.

If you want to test this past the tests you can see it by importing the `mockFormData` and running that through the validate function.

AntBush: Ah gotcha!

AntBush (approved):

PR #915, opened 2022-11-18
feat: add screening schema

AntBush:
Implements #  804


app/pages/analyst/application/[applicationId]/assessments.tsx
```diff
@@ -20,15 +32,55 @@ const getAssessmentsQuery = graphql`
...
+        <FormBase
+          schema={screening}
+          uiSchema={screeningUiSchema}
+          noValidate
+          onChange={(e) => {
+            setFormData(e.formData);
+          }}
+          formData={formData}
+          formContext={{ query }}
+          tagName="div"
```
Comment (marcellmueller) [inline app/pages/analyst/application/[applicationId]/assessments.tsx:73]: I have good but annoying news, all the source of our troubles was this `tagName="div"` which changes the `Form` to a `div`. The reason this was brought over for the analyst application page is it is taken from the review `FormBase` which we don't want to submit by default.

marcellmueller:
I spent a bunch of time with the so leaving a couple other notes for the morning:

`FileWidget` needs an update so it reads the `applicationId` query param as it was originally written for the applicant side of the portal. Easy fix is line 181 needs to be changed to something like:

` const formId = parseInt(router?.query?.id as string, 10) || parseInt(router?.query?.applicationId as string, 10);`

Also when you fix that and try and upload a file you'll notice that we need to give `ccbc_admin` and `ccbc_analyst` permissions to insert on table `attachment`.

Looking great other than that!

marcellmueller (approved):
Looks good!

Issue #881, opened 2022-11-06
Analyst - Create RFI


eazeredo:
## User Story
As an Analyst
I want to create an RFI 
So that the applicant will be able to upload the requested files.

## Acceptance Criteria:
- [x] **RFI in the sidebar**
	Given an analyst is on an application page where they can see the sidebar
	when they click on the RFI tab
	then they will see the RFI page
icon is [envelope from font-awesome](https://fontawesome.com/icons/envelope?s=regular&f=classic)
#
- [x] **Show 'New RFI' button**
	Given that I'm on the RFI page
	When I click on "New RFI"
	Then I will be directed to the RFI form
#
- [x] **RFI Number generated**
Given that an analyst is a person who exists in the world and has watched Lion King 16 times
when they click save
Then an RFI# will be autogenerated
and the format will be RFI-CCBC-XXXXYY-Z
RFI-`CCBC ID`-`#`
For CCBC-010023's first RFI, the # would be: RFI-CCBC-010023-1
For CCBC-010102's fifth RFI, the # would be: RFI-CCBC-010102-5
#
- [x] **RFI type** checkboxes
- [x] **Due by date** date-picker
- [x] **Email correspondence** file upload, no file type restrictions, multiple files allowed
- [x] **Checkboxes for requesting files**
#
- [ ] **Action: Save**
	Given that an analyst is on the RFI page
	When they click on Save
	Then the values are created in the database
	And I'm redirected to the page "List RFIs"
	And a success toast with the text "RFI saved"
#
- [x] **Cancel**
Given that an analyst is on the RFI form
when they press cancel
then any information is abandoned
and the RFI number can be used for the next RFI submitted for this application
#

## Technical Analysis
- [ ] On creation of an RFI, a new schema is stored in the ccbc_public.form table with type `rfi`
- [ ] Should create a new computed column to easily get `rfi` for an application
- [ ] .

## Wireframes
[Wireframes](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=7581%3A36769&t=Et08WUiZAvcXQQVj-4)

### New RFI
![RFI_create button.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/94b30350-70b0-43d6-9f58-adda9fc0353b)

### New RFI - Empty state
[...]

PR #929, opened 2022-11-28
feat: db changes for RFI

VB-at-Bis:
Implements backend support for #881

- [x] On creation of an RFI, a new schema is stored in the ccbc_public.form table with type rfi
- [x] Should create a new computed column to easily get rfi for an application

P.S. Tests were updated to include TRUNCATE ... CASCADE due to new foreign key dependencies.

### DIFF
--- a/db/deploy/mutations/create_rfi.sql
+++ b/db/deploy/mutations/create_rfi.sql
@@ -1,0 +1,36 @@
+-- Deploy ccbc:mutations/create_rfi to pg
+
+begin;
+
+create or replace function ccbc_public.create_rfi(application_row_id int, json_data jsonb)
+returns ccbc_public.rfi_data
+as $function$
+declare
+  result ccbc_public.rfi_data;
+  new_rfi_id int;
+  new_application_rfi_id int;
+  new_rfi_number varchar(1000); 
+begin
+  select 1 + count(*) into new_rfi_id from ccbc_public.application_rfi_data where application_id=application_row_id;
+  select CONCAT(ccbc_number,'-',new_rfi_id) into new_rfi_number from ccbc_public.application where id=application_row_id;
+
+
+  -- using nextval instead of returning id on insert to prevent triggering select RLS,
+  -- which requires the rfi_data record
+  new_rfi_id := nextval(pg_get_serial_sequence('ccbc_public.rfi_data','id'));
+  
+  insert into ccbc_public.rfi_data (id, rfi_number, json_data, rfi_data_status_type_id) overriding system value
+    values (new_rfi_id, new_rfi_number, json_data, 'draft');
+
+  insert into ccbc_public.application_rfi_data (application_id, rfi_data_id)
+   values (application_row_id, new_rfi_id);
+
+  select * from ccbc_public.rfi_data where id = new_rfi_id into result;
+  return result;
+end;
+$function$ language plpgsql strict volatile;
+
+grant execute on function ccbc_public.create_rfi to ccbc_analyst;
+grant execute on function ccbc_public.create_rfi to ccbc_admin;
+
+commit;
--- a/db/deploy/tables/application_rfi_data.sql
+++ b/db/deploy/tables/application_rfi_data.sql
@@ -21,27 +21,36 @@ $grant$
 begin
 perform ccbc_private.grant_permissions('select', 'application_rfi_data', 'ccbc_analyst');
 perform ccbc_private.grant_permissions('insert', 'application_rfi_data', 'ccbc_analyst');
+perform ccbc_private.grant_permissions('select', 'application_rfi_data', 'ccbc_admin');
+perform ccbc_private.grant_permissions('insert', 'application_rfi_data', 'ccbc_admin');
 
 end
 $grant$;
 
--- RLS: can only see and insert its own records in rfi_data
+-- RLS
 do
 $policy$
 begin
 
-perform ccbc_private.upsert_policy('ccbc_analyst_insert_application_rfi_data', 'application_rfi_data', 'insert', 'ccbc_analyst',
-'application_id in (select id from ccbc_public.application)');
-
-perform ccbc_private.upsert_policy('ccbc_analyst_select_application_rfi_data', 'application_rfi_data', 'select', 'ccbc_analyst',
-'application_id in (select id from ccbc_public.application)');
-
+perform ccbc_private.upsert_policy('ccbc_analyst_insert_application_rfi_data', 
+  'application_rfi_data', 'insert', 'ccbc_analyst', 'true');
+perform ccbc_private.upsert_policy('ccbc_analyst_select_application_rfi_data', 
+  'application_rfi_data', 'select', 'ccbc_analyst', 'true');
 perform ccbc_private.upsert_policy('ccbc_analyst_select_rfi_data',
-  'rfi_data', 'select', 'ccbc_analyst',
-  'id in (select rfi_data_id from ccbc_public.application_rfi_data)');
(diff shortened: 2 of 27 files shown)


db/deploy/tables/rfi_data.sql
```diff
@@ -0,0 +1,41 @@
+-- deploy ccbc:tables/rfi_data to pg
+
+begin;
+
+create table ccbc_public.rfi_data(
```
AntBush, line 5: Do we want to create a new table for the RFI data? I was thinking we'd reuse the `form_data` table to hold the RFIs and to differentiate them based on type

db/deploy/mutations/create_rfi.sql
```diff
@@ -0,0 +1,30 @@
...
+  new_rfi_id int;
+  new_rfi_number varchar(1000); 
+begin
+  select 1 + count(*) into new_rfi_id from ccbc_public.application_rfi_data where application_id=row_id;
+  select CONCAT(ccbc_number,'-',new_rfi_id) into new_rfi_number from ccbc_public.application where id=row_id;
+
+  insert into ccbc_public.rfi_data (rfi_number, json_data, rfi_data_status_type_id) 
+    values (new_rfi_number, '{}'::jsonb, 'draft')
+   returning id into new_rfi_id;
+
```
AntBush, line 19: We had a discussion about how we want the RFIs to be created, specifically how we want the applicant to see an RFI. Currently the RFI table will be holding the json data for an RFI form, but the resultant form isn't being created here. Are we planning on using the form_data to create the resultant schema on the applicant side dynamically then?

db/deploy/tables/application_rfi_data.sql
```diff
@@ -0,0 +1,56 @@
...
+begin
+
+perform ccbc_private.upsert_policy('ccbc_analyst_insert_application_rfi_data', 'application_rfi_data', 'insert', 'ccbc_analyst',
+'application_id in (select id from ccbc_public.application)');
+
+perform ccbc_private.upsert_policy('ccbc_analyst_select_application_rfi_data', 'application_rfi_data', 'select', 'ccbc_analyst',
+'application_id in (select id from ccbc_public.application)');
+
+perform ccbc_private.upsert_policy('ccbc_analyst_select_rfi_data',
+  'rfi_data', 'select', 'ccbc_analyst',
```
AntBush, line 40: If we're going to have the schema dynamically created when viewing and submitting the form, this is missing `ccbc_auth_user` select RLS to be able to view the RFI

AntBush (commented):
There are some changes that I think need to be made, but I also see that it's not included in the associated ticket, but should probably be taken into account here.

db/deploy/tables/rfi_data.sql
VB-at-Bis: I dont think re-using `form_data` would be feasible as it would require custom RLS (one form type belongs to customer, another to analyst). And application can have multiple RFIs .. Separate table seems to be cleaner solution.

db/deploy/mutations/create_rfi.sql
VB-at-Bis: I followed existing approach - application is created with empty json_data, and then gets updated.

db/deploy/tables/application_rfi_data.sql
VB-at-Bis: good point. I'll ad `ccbc_auth_user` select RLS

db/deploy/mutations/create_rfi.sql
AntBush: I meant moreso for the applicant side, the RFI form will only describe the type of schema that the applicant should see without building the schema itself. When an applicant views the RFI, they'll be needing to update or add uploads and we'll need a schema that reflects the RFI created by an analyst. 

It's likely out of scope of this ticket, but the question as to whether the creation of an RFI should build this schema or whether we should reference existing schemas in the `form` table (or some other third solution) isn't being addressed. I'm mostly concerned we sign up with this approach and then are in a corner with regard to how we display the form to the applicant

VB-at-Bis: I updated DB schema at https://miro.com/app/board/uXjVOnpFeNI=/ (Frame 3) and my implementation reflects that. My understanding is that RFI would be using schema created in `form` table. I do not think that RFI schema is dynamic in this sense. 
Please feel free to update DB schema on Miro board if you think that I've missed something

db/deploy/tables/application_rfi_data.sql
VB-at-Bis: Fixed

db/deploy/mutations/create_rfi.sql
```diff
@@ -0,0 +1,30 @@
+-- Deploy ccbc:mutations/create_rfi to pg
+
+begin;
+
+create or replace function ccbc_public.create_rfi(row_id int)
```
AntBush, line 5: Nit: For clarity, can we make this:  
```suggestion
create or replace function ccbc_public.create_rfi(application_row_id int)
```

AntBush (approved):
Looks good! Just one small optional change

db/deploy/mutations/create_rfi.sql
AntBush: As Marcel stated, this is missing adding in the json data. Please add that in there

AntBush (dismissed):

db/deploy/mutations/create_rfi.sql
VB-at-Bis: Done

db/deploy/tables/rfi_data.sql
```diff
@@ -0,0 +1,51 @@
...
+  rfi_number varchar(1000),
+  json_data jsonb not null default '{}'::jsonb,
+  rfi_data_status_type_id varchar(1000) references ccbc_public.rfi_data_status_type(name) default 'draft'
+);
+select ccbc_private.upsert_timestamp_columns('ccbc_public', 'rfi_data');
+
+alter table ccbc_public.rfi_data force row level security;
+alter table ccbc_public.rfi_data enable row level security;
+
+grant usage, select on sequence ccbc_public.rfi_data_id_seq to ccbc_analyst;
```
marcellmueller, line 16:
```suggestion
grant usage, select on sequence ccbc_public.rfi_data_id_seq to ccbc_analyst;
grant usage, select on sequence ccbc_public.rfi_data_id_seq to ccbc_admin;

```

db/deploy/tables/rfi_data_status_type.sql
```diff
@@ -0,0 +1,28 @@
...
+);
+
+insert into ccbc_public.rfi_data_status_type values
+ ('draft', 'Draft'), ('sent', 'Sent');
+
+do
+$$
+begin
+
+perform ccbc_private.grant_permissions('select', 'rfi_data_status_type','ccbc_analyst');
```
marcellmueller, line 17:
```suggestion
perform ccbc_private.grant_permissions('select', 'rfi_data_status_type','ccbc_analyst');
perform ccbc_private.grant_permissions('select', 'rfi_data_status_type','ccbc_admin');

```

marcellmueller: I was going to merge and update, it's working perfectly other than minor admin permissions. Though @AntBush needs to give the final approval since he requested changes :)

VB-at-Bis: Fixed

db/deploy/tables/rfi_data.sql
VB-at-Bis: Fixed

AntBush (approved):

PR #934, opened 2022-11-29
fix: location of freshclam.conf

VB-at-Bis:
Implements #275

Using dynamically created freshclam.conf file (sored in /tmp because it's the only writable folder available for lambda).
freshclam executable could not access freshclam.conf file located in the same /opt/bin folder.


marcellmueller (approved):
:rocket:

PR #943, opened 2022-12-01
881 create rfi

marcellmueller:
Front end ticket for #881 

Since we had to create new pages for this it also includes a somewhat hefty relay fragment refactor for all of the `/analyst` route subpages and some tests. 

Also fixed the typography so the font base size is 16px instead of 18px.


VB-at-Bis (approved):
LGTM

Issue #37, opened 2022-04-05
Analyst - List RFIs


eazeredo:
## User Story
As an Analyst
I want to see what RFIs were sent to an applicant
So that I can have a record.

## Acceptance Criteria:

- [x] **Sort RFIs by most recent created**
	Given I'm analyzing an application
	When I click on RFI in the sidebar
Or I press save in a new RFI
	Then I see a list of existent RFIs sorted by most recent created (created on - Descending)
#
- [x] **Scenario: display after save**
    Given that an analyst has filled fields on an RFI
    When they press save
    Then the fields will be displayed at the top of the RFI page
#
- [x] **Show 'not received' text for Files not received yet - Empty state**
    Given an existent RFI
    And one or more files have been requested
    When the applicant has _not_ uploaded the requested files
    Then the RFI page will state "not received" in light grey
#
- [x] **Show applicant's files**
    Given an RFI has been requested
    And one or more files have been requested
    When the applicant uploads the requested files
    Then the filenames will appear
#
- [x] **Download applicant's files**
    Given the applicant's files are displayed 
    When I click on the filename
    Then the file will begin downloading
    Applicant file names will be prefixed with CCBC ID (e.g. CCBC-010001 - Template 1.xlsx)

#
- [x] **Scenario: download email files**
    Given that an analyst has saved an RFI
    and uploaded a file
    When they click on the filename
    Then the file will begin downloading
    Email file names will be prefixed with CCBC ID (e.g. CCBC-010001 - Template 1.xlsx)
#

## Not included:
- Edit button is part of another ticket #895
-

## Technical Analysis
- What is the front-end complexity?
- Do it require creating or updating a component?
- Any work on the back-end?
- Integration?
- Any changes in database structure?
- What types of tests that we need to create and run?
- Impacts on DevOps?
- Other risks and potential side effects?
- Which documentation needs to be updated?

## Wireframes
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=8194%3A50768&t=PtcAayUyV4IigGnM-4)
**Files not received yet - Empty state**
![RFI_list.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/0b94a383-7af8-4768-beb1-79770b58443a)

**Files received**
[...]

PR #950, opened 2022-12-03
37 analyst list rfis db

VB-at-Bis:
Implements #37


VB-at-Bis:
Same can be achieved with graphql query

Issue #939, opened 2022-11-30
Virus scan - handle tagged files


MarsRomer-zz:
As an analyst 
I need to be sure files with virus' don't get downloaded to my computer
so I can keep the government system clean and dependable

Description:  When the virus scanner finds a tagged file, we need a way to prevent it from downloading  in any way to anyone's machine.

## Acceptance Criteria
 - [ ] We (CCBC button inbox and PO) should get email notification when there is a file with virus that is uploaded by the applicant
 - [ ] If there is one or more tagged file then the Download Attachments button will not work 
or 
- [ ] Given a file is infected
When an analyst clicks "export all"
Then the file is replaced with an empty file with the prefix 'Corrupted_'

## Technical Analysis
Steps how to react to email notification about infected file.
If there is an infected file, the CCBC ID and the file type and name will be determined by a query in Metabase (tbd) and the file will be deleted from the S3

MarsRomer:
@elliottSlee @zoeyli-46  I would like to bring this ticket into our next sprint, please take a look to see how we want to handle files with virus - something to discuss at sprint planning

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/939?planning-poker) with Zenhub @AntBush @marcellmueller @VB-at-Bis

PR #951, opened 2022-12-05
feat: update terraform to provision SNS topic for lambda notifications

VB-at-Bis:
Implements #939 (part 1)

 [x] We (CCBC button inbox and PO) should get email notification when there is a file with virus that is uploaded by the applicant


marcellmueller (approved):
Looks good!

PR #959, opened 2022-12-07
feat: use updater to manually link new formData object

AntBush:
Issue has to do with creating a new record and Relay not knowing what to link it to: https://github.com/facebook/relay/issues/1652#issuecomment-343332001 

This uses an updater to fix the problem, an optimisticUpdater should also be added to fix this issue since right now it switches back and forth from the old FormData (from the optimistic response not having the data) to the new FormData (after the updater runs with the response)


VB-at-Bis (approved):
Looks good

PR #965, opened 2022-12-08
895 edit rfi

AntBush:
Implements # 895

While in draft, is branched off of `37-list-rfi`, should be rebased onto master before being put into review

### DIFF
--- a/app/pages/analyst/application/[applicationId]/rfi/[rfiId]/index.tsx
+++ b/app/pages/analyst/application/[applicationId]/rfi/[rfiId]/index.tsx
@@ -9,6 +9,9 @@ import { RfiIdQuery } from '__generated__/RfiIdQuery.graphql';
 
 const getRfiIdQuery = graphql`
   query RfiIdQuery($rowId: Int!) {
+    rfiDataByRowId(rowId: $rowId) {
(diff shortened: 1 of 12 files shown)


app/pages/analyst/application/[applicationId]/rfi/[rfiId]/index.tsx
```diff
@@ -9,6 +9,9 @@ import { RfiIdQuery } from '__generated__/RfiIdQuery.graphql';
 
 const getRfiIdQuery = graphql`
   query RfiIdQuery($rowId: Int!) {
+    rfiDataByRowId(rowId: $rowId) {
```
Comment (marcellmueller) [inline app/pages/analyst/application/[applicationId]/rfi/[rfiId]/index.tsx:12]: This `rowId` is using the `applicationId` as you can see in the `variablesFromContext` at the bottom of the file so it returns the incorrect rfi to edit.

Easy to see how this would be confusing given the poorly named query due to relay naming conventions. Though this query still needs the `applicationId` as it is used in the `AnalystLayout_query` fragment.

Comment (AntBush) [reply app/pages/analyst/application/[applicationId]/rfi/[rfiId]/index.tsx:12]: Ah thanks, great catch

marcellmueller (approved):
:ship:

PR #969, opened 2022-12-08
chore: release v1.21.1

AntBush:
Release to include fix for logic error to fully fix assessments


AntBush:
@marcellmueller @VB-at-Bis no rush to merge this one, we can do it after the load testing

VB-at-Bis (approved):
Looks safe

PR #977, opened 2022-12-10
feat: add haproxy timeout to secure route chart

marcellmueller:
Implements # \<issue number\>


wenzowski (approved):
Confirmed fix for 504 timeout at 30s by manual testing

PR #992, opened 2022-12-13
fix: uploads are sometimes 0 bytes

AntBush:
Implements No issue


app/package.json
```diff
@@ -13,6 +13,7 @@
     "lint": "next lint",
     "relay": "mkdir -p __generated__ && relay-compiler",
     "pretest": "yarn run build:relay",
+    "postinstall": "yarn patch-package",
```
wenzowski, line 16: tech debt: patching is part of yarn 2

app/package.json
```diff
@@ -52,9 +53,11 @@
     "luxon": "^3.0.1",
     "next": "^12.1.6",
     "openid-client": "^5.1.6",
+    "patch-package": "^6.5.0",
     "path-to-regexp": "^6.2.1",
     "pluralize": "^7.0.0",
     "postgraphile": "^4.12.9",
+    "postinstall-postinstall": "^2.1.0",
```
wenzowski, line 60: Yarn runs the postinstall script after `yarn`, `yarn install` and `yarn add <package>` but not after `yarn remove <package>`. This postinstall hook will run even after a `yarn remove <package>`.

app/patches/fs-capacitor+6.2.0.patch
```diff
@@ -0,0 +1,18 @@
...
+             // If there were no more bytes to read and the write stream is finished,
+             // than this stream has reached the end.
+             if (this._writeStream._writableState.finished) {
+-                this.push(null);
++                // Check if we have consumed the whole file up to where
++                // the write stream has written before ending the stream
++                if (this._pos < this._writeStream._pos)
++                    this._read(n);
++                else
++                    this.push(null);
```
wenzowski, line 15: this fix for our 0 byte files issue is merged upstream but we need to abandon commonjs to pull it down as esm

wenzowski (approved):
Ship it with tracked tech debt? Nextjs is old, typescript is old, app is using commonjs not esm, yarn is old

Issue #999, opened 2022-12-13
Metabase query in prod not working


MarsRomer2:
The base query for all application data works with test data but switching it to prod does not work currently.  

This ticket is to allow the query to work with prod data as well as test data.

VB-at-Bis:
Permissions changed in Prod. Now Base Query works in Prod (no data in response because there are no submitted applications yet). https://ccbc-metabase.apps.silver.devops.gov.bc.ca/question/36-all-application-data

PR #1000, opened 2022-12-13
fix: added permissions to function application_status

VB-at-Bis:
Implements #999

Metabase uses ccbc_readonly user which did not have execute permissions on function `application_status`.
This PR fixes the issue.

### DIFF
--- a/db/deploy/computed_columns/application_status.sql
+++ b/db/deploy/computed_columns/application_status.sql
@@ -14,7 +14,10 @@ grant execute on function ccbc_public.application_status to ccbc_auth_user;
 grant execute on function ccbc_public.application_status to ccbc_job_executor;
 grant execute on function ccbc_public.application_status to ccbc_analyst;
 grant execute on function ccbc_public.application_status to ccbc_admin;
-grant execute on function ccbc_public.application_status to public; -- needed for Metabase
+
+-- ccbc_readonly user is used by Metabase and it is added outside of this project.
+-- statement below should be run manually to allow Metabase users to use the function.
+-- grant execute on function ccbc_public.application_status to ccbc_readonly;  
 
 comment on function ccbc_public.application_status is 'Computed column to return status of an application';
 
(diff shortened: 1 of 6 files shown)


db/sqitch.plan
```diff
@@ -118,3 +118,4 @@ tables/attachment_001_add_analyst_roles 2022-12-06T16:02:43Z Anthony Bushara <an
 @1.22.0 2022-12-08T20:08:49Z Anthony Bushara <anthony@button.is> # release v1.22.0
 @1.23.0 2022-12-11T04:15:58Z Marcel Mueller <marcel@button.is> # release v1.23.0
 @1.23.1 2022-12-13T03:39:09Z Anthony Bushara <anthony@button.is> # release v1.23.1
+computed_columns/application_status [computed_columns/application_status@1.20.0] 2022-12-13T18:14:55Z ,,, Vladimir <vladimir@button.is> # ccbc_readonly role
```
AntBush, line 121: I see a rework was done here, shouldn't there be a tagged version of the `application_status` file as well in the file changes?

VB-at-Bis: fixed

db/deploy/computed_columns/application_status.sql
```diff
@@ -14,6 +14,7 @@ grant execute on function ccbc_public.application_status to ccbc_auth_user;
 grant execute on function ccbc_public.application_status to ccbc_job_executor;
 grant execute on function ccbc_public.application_status to ccbc_analyst;
 grant execute on function ccbc_public.application_status to ccbc_admin;
+grant execute on function ccbc_public.application_status to public; -- needed for Metabase
```
VB-at-Bis, line 17: Please comment on that. I assume it is safe but please correct me if I'm wrong. Any alternative solutions are welcome.

AntBush: I'm tempted to say that it's safe as well, since this is neither volatile nor is it exposed as an endpoint in graphql for it to be abused. 

Alternatively, option 2 for this is to go right into the DB and grant execute function manually to the ccbc_readonly user. I'd rather have this saved in our codebase though. 

I'm okay with this though. Do we know why the application_status function works fine in test for metabase but not on prod? If we've already done option 2 for test may as well stay consistent

VB-at-Bis: I vote for option 2 :) . I will add commented out line here and update README just to keep track.

AntBush (approved):

PR #1001, opened 2022-12-13
986 save email uploads

marcellmueller:
Adds saving for the email uploads. Also refactored the `RFI` component to use a fragment.


AntBush (approved):
Tried it locally and it works well, appreciate the refactor to use a fragment!

PR #1019, opened 2022-12-15
fix/datetime

marcellmueller:
This is using a feature flag `show_subtracted_time` which returns a number which is set at 30. Either the number can be changed or the feature can be disabled and the time will show as normal.

### DIFF
--- a/app/pages/applicantportal/index.tsx
+++ b/app/pages/applicantportal/index.tsx
@@ -6,7 +6,7 @@ import Link from '@button-inc/bcgov-theme/Link';
 import styled from 'styled-components';
 import { useMemo } from 'react';
 import { Button, Callout } from '@button-inc/bcgov-theme';
-import { DateTime } from 'luxon';
+import dateTimeSubtracted from 'utils/dateTimeSubtracted';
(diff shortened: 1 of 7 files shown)


app/pages/applicantportal/index.tsx
```diff
@@ -89,9 +93,11 @@ const Home = ({
         </>
       );
 
-    const formattedClosingDate = DateTime.fromISO(openIntake.closeTimestamp)
-      .minus({ minutes: 30 })
-      .toLocaleString(DateTime.DATETIME_FULL);
+    // not ideal but checking type here for now due mockOpenIntake applicantportal/index test mocking useFeature and returning object
+    const formattedClosingDate = dateTimeSubtracted(
+      openIntake.closeTimestamp,
+      typeof showSubtractedTime === 'number' ? showSubtractedTime : 0
```
marcellmueller, line 99: Just flagging this here, I had to do a type check to get an unrelated test in `applicantportal/index.test.ts` that failed due to mocking `useFeature` with an unrelated feature that returned an object. Spent a bit of time trying to get it to pass without this so I'm open to suggestions, though in the absence of time it was easiest to do this.

AntBush (approved):

PR #1035, opened 2022-12-16
test: testing archive after invalid entry

VB-at-Bis:


app/config/index.js
```diff
@@ -191,7 +191,7 @@ const config = convict({
   CHECK_TAGS: {
     doc: 'Enable checking tags on s3 object',
     format: Boolean,
-    default: false,
+    default: true,
```
marcellmueller, line 194: I assume this is ok now that permissions have been added for object tags?

VB-at-Bis: Yes

marcellmueller (approved):
Looks good!

Issue #954, opened 2022-12-06
Virus - handle single file download 


marcellmueller:
As an analyst or applicant
I want the single file downloads not to allow viruses' to download to my computer
to keep my system safe

## Acceptance Criteria:
- [ ] As an analyst or applicant
Given that I have an infected file
When I click on it to download
Then a modal displaying an error will appear and the download will not happen

## Technical details:
- Back-end: Update S3 download route to fetch tag and error (500) if it is tagged 
- Front-end: Use the Toolkit modal
- Tests: Mock return of error on frontend for modal, mock s3 tag to test route in backend

## [Wireframe](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=9184%3A43533&t=ZuS50PheSOShTqzq-4)
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/74af90d1-5e52-46f6-a21a-a92ff4efe5c7)

elliottSlee:
Related to #962

MarsRomer2:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/954?planning-poker) with Zenhub @AntBush @marcellmueller @pdoc1111 @VB-at-Bis

MarsRomer2:
@MarsRomer2  update wording for prefix file name

MarsRomer2:
@elliottSlee @zoeyli-46 can you please add mockups for the modal here?

PR #1052, opened 2022-12-22
feat: handle single file download

VB-at-Bis:
Implements #954


marcellmueller (approved):
This is working well and thanks for moving the modal out of the file widget. Just needs eslint error fixed in s3download it looks like :)

PR #1064, opened 2023-01-03
fix: assessment delete attachment error

marcellmueller:
Fixes the error deleting attachments on the assessments page after it has been saved as well as allowing archived attachments to be deleted from the form data incase someone deletes then refreshes the page without saving.

### DIFF
--- a/app/pages/analyst/application/[applicationId]/assessments/screening.tsx
+++ b/app/pages/analyst/application/[applicationId]/assessments/screening.tsx
@@ -39,6 +39,9 @@ const ScreeningAssessment = ({
 
   const { applicationByRowId, session } = query;
   const [createAssessment, isCreating] = useCreateScreeningAssessmentMutation();
+  const [formData, setFormData] = useState(
+    applicationByRowId.assessmentForm?.jsonData
+  );
   const [isFormSaved, setIsFormSaved] = useState(false);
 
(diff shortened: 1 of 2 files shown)


app/pages/analyst/application/[applicationId]/assessments/screening.tsx
```diff
@@ -74,9 +77,10 @@ const ScreeningAssessment = ({
           schema={screening}
           uiSchema={screeningUiSchema}
           noValidate
-          formData={applicationByRowId.assessmentForm?.jsonData}
-          onChange={() => {
+          formData={formData}
+          onChange={(e) => {
             setIsFormSaved(false);
+            setFormData({ ...e.formData });
```
Comment (AntBush) [inline app/pages/analyst/application/[applicationId]/assessments/screening.tsx:83]: Do we have to make this change in other places as well?

AntBush (approved):
Tested and works fine, just a small question but it works as-is in all places where we use the file widget.

app/pages/analyst/application/[applicationId]/assessments/screening.tsx
Comment (marcellmueller) [reply app/pages/analyst/application/[applicationId]/assessments/screening.tsx:83]: No I don't believe so since the application form auto saves it doesn't have this issue though we will make sure the rest of the assessment pages use this as well. I think we can make a reusable assessments form when we go to do the rest of them.

Issue #1062, opened 2022-12-29
Type of Organization - wrong value showing


MarsRomer2:
Bug:  in Analyst portal - Type of Organization shows as 'True' instead of the value stored in the db.

This is what is shows:

![Image](https://user-images.githubusercontent.com/119887180/210003803-a59e104f-66b6-40e8-bb3b-b55d18e8ac30.png)

These are the values for example in the database that need to be displayed in the Analyst Portal
![Image](https://user-images.githubusercontent.com/119887180/210003440-96ca467f-2e39-4dc8-a099-7333e8e3a1f7.png)

MarsRomer2:
Good work!

PR #1065, opened 2023-01-03
fix: review schema type of organization display value

marcellmueller:
Fix for #1062 `typeOfOrganization` radio widgets field displaying as 'Yes' instead of the enum value on the review pages. You can see this is fixed in the Happo diffs here:

https://happo.io/a/336/p/1172/compare/67fd91f05835a7a6a050c9b5269cfdd5c319555a/586a6822219b30bc36bd66dd0c5a5d994b9b76da


VB-at-Bis (approved):
Looks good

PR #1066, opened 2023-01-04
848 table of assessments

marcellmueller:
This only has the row for screening assessment right now but the rest will be easy to add in. I will do that when I work on the assessments tabs ticket.


VB-at-Bis (approved):
Looks good

Issue #989, opened 2022-12-12
Implement tabs for the assessments


MarsRomer2:
As an analyst 
I want to see the assessments on separate tabs
so that I can get to them from the assessment tables

## Acceptance Criteria:

- [ ] Each assessment is in it's own tab as per design
![assessment tabs.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/f497fba8-323f-44da-8e3f-8dbcd2823c4b)

- [x] Screening assessment exists as a tab
- [x] Currently open table is bolded & standard text colour
- [x] Not opened tabs are not bold & grey (9B9B9B)

## Not included
- other tabs that are not yet created

## Technical Analysis
- [ ] Needs routing update
- [ ] Routing update requires testing to update
- [ ] Should pass formName to query from route

## Ultra refinement
- DB:
Nothing here.
- Frontend:
Yes. 
- Integration/API:
No
- Tests:
Yes
- Documentation:
No

MarsRomer2:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/989?planning-poker) with Zenhub @AntBush @marcellmueller @pdoc1111 @VB-at-Bis

MarsRomer2:
love it!

PR #1071, opened 2023-01-04
feat: add assessments tabs and placeholder pages

marcellmueller:
Implements #989 

Also I considered fixing it in this PR but I've opened an issue to refactor the session sub to use relay query fragments so we stop getting those warnings in our PRs. That refactor will touch every page so I think it's best to have it's own PR: https://github.com/bcgov/CONN-CCBC-portal/issues/1072


VB-at-Bis (approved):
Looks good; no need to refactor in this ticket :)

Issue #856, opened 2022-11-03
Enter 'Financial' Assessment


zoeyli-46:
As an Analyst 
I want to complete the Financial assessment 
So that I can manage it and record the outcomes.

## Acceptance Criteria:
- [ ] The PM assessment page will be designed as below:

![Untitled.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/11fbc866-dc7e-450e-9e09-743824ae25e9)

##
- [ ] The uploads will be multiple file uploads

- [ ] Behaviour of the dropdown, date field, radio buttons, uploads, and save will be the same as other assessment screens.

## Not included:
-
-

## Ultra refinement
- DB:
- [ ] .
- Frontend:
- [ ] . 
- Integration/API:
- [ ] .
- Tests:
- [ ] .
- Documentation:
- [ ] .

## Wireframes
https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=7963%3A39193&t=SZrAF09FkM4Z05Sl-4

elliottSlee:
@MarsRomer screenshot's updated

Issue #855, opened 2022-11-03
Enter 'PM' Assessment


zoeyli-46:
As an analyst 
I want to complete the PM assessment 
so that I can manage it and record the outcomes.

## Acceptance Criteria:
- [x] Project Management tab looks like design
![image.png](https://images.zenhubusercontent.com/638e5e789d9295d36fe268e9/33faf072-ad05-47d6-bfcf-b682248080ee)

- [ ] Multifile uploads

- [x] Components behave the same as the 'Screening' tab

elliottSlee:
@MarsRomer screenshot's updated

MarsRomer2:
Looks good, only one thing needs changing - the other files needs to be a multi file upload.  But I am going to put this in 'Done' for now and we can create a bug ticket for fixing that as it's not urgent

Issue #854, opened 2022-11-03
Enter 'Technical' Assessment


zoeyli-46:
As an analyst 
I want to complete the technical assessment 
So that I can manage it and record the outcomes.

## Acceptance Criteria:

- [x] Tech assessment tab looks like the design
![image.png](https://images.zenhubusercontent.com/638e5e789d9295d36fe268e9/598acf0b-7461-4551-b8ac-19a3c6046e51)

- [x] components behave the same as in 'Screening' tab
 - [ ] multifile uploads

- [x] Decisions are pass and fail
![image.png](https://images.zenhubusercontent.com/638e5e789d9295d36fe268e9/95bcc37d-aca0-4149-a635-ae27e3c4fc45)

zoeyli-46:
@eazeredo @elliottSlee is it fair to say that for each of the assessment tickets AC, we will just list out the "options", since otherwise they are all the same as #874?

elliottSlee:
Even thought AC are the same across tickets, they still need to be written out. We aren't creating features, we're creating user experiences. If interested, look up differences between test driven development, behaviour driven development, and other models. Our user stories needs to explain everything the user is doing.

elliottSlee:
@MarsRomer the screenshot has been updated.

MarsRomer2:
Looks great, just one minor thing needs changing is the 'Other files' should be a multi-file upload.  I will put this in the done column and we can create a bug ticket for it

PR #1074, opened 2023-01-05
Assessment forms

marcellmueller:
This PR handles #854, #855 and #856 since they are all are so similar with very minor differences.

This is all working and saving correctly, though will need a minor refactor in the refactor PR #1071.


AntBush (approved):
Tested and it works great, looks solid!

PR #1077, opened 2023-01-06
refactor: use assessments table

AntBush:
Implements # 1061


marcellmueller (approved):
Looks good, nice work

PR #1096, opened 2023-01-11
fix: reply rfi upload files button display correct on rfi due date

marcellmueller:
Fix for the replay RFI `Upload files` button not showing up on the `applicant` dashboard on the date the rfi is due by. It would only show up if you set the date to the day before, so I made a small change to `application_has_rfi_open` function to fix this.


VB-at-Bis (approved):
Looks good

Issue #370, opened 2022-08-02
Add intake column to applicant dashboard


MarsRomer-zz:
### User Story
As an applicant, 
I want to see what intake my applications were submitted in
so that I can better organize my submissions

### Acceptance Criteria
- [x] Dashboard shows intake column 
- [x] Submitted and withdrawn applications show the correct intake number in this column.  
- [x] Draft intakes should not have an intake number

![Screen Shot 2022-10-05 at 2.08.45 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/fe9f44c7-7ac4-47d7-aaa6-0011ffc60759)

[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=4343%3A17384)

```
- [ ] Daylight saving time change will be handled correctly.

Scenario Outline: Choose what intake dates to show
Given today is <today>
And the existing intake dates in the system
| intake | startDate  | endDate    |
| 1      | 2022-09-07 | 2022-11-06 |
| 2      | 2022-11-07 | 2023-01-04 |
When I open the application
Then the start date = <start> and the end date = <end>

Examples:
| today      | start      | end        |
| 2022-08-30 | 2022-09-07 | 2022-11-06 | # intake is not open yet
| 2022-09-07 | 2022-09-07 | 2022-11-06 | # first intake is running
| 2022-11-07 | 2022-11-07 | 2023-01-04 | # second intake is running
| 2023-02-01 | ?          | ?          | # no future intakes

```

AntBush:
> then the dashboard table will include a column with the intake #

We want a new column for the dashboard? Or do we just want a newly-submitted application (after the first intake) to have a CCBC-ID that is aligned with the new intake number

MarsRomer:
we actually want a new column called Intake @elliottSlee  can provide screen design

elliottSlee:
Wireframe added. Should make it more clear.

AntBush:
So I'm still a bit confused. For Sudden Valley its intake is 2, what gives it that intake? I am under the assumption the Intake column is the equivalent of the first two digits of the CCBC-ID (CCBC-**01**1234 would have intake value of **1**).

elliottSlee:
@AntBush Sorry for the confusion in the wireframe. I'll update it in a bit. Intake # should only be assigned after it's submitted. 
[...]

PR #1098, opened 2023-01-11
370 intake column

marcellmueller:
- feat: add applicant intake column and cleanup dashboard
- test: add applicant dashboard intake column tests

Implements #370


VB-at-Bis (approved):
Looks good

Issue #1106, opened 2023-01-12
Bug: If RFI files requested = null then do not show upload files button on applicant dashboard


MarsRomer2:
**Describe the bug**
Applicants can still see the upload files button when there are no files being asked for (but there used to be).  
Applicant still has the ability to upload a file that was unselected (when at least one other file is being asked for)

**To Reproduce**
Part 1
1. Go to RFI screen on analyst portal
2. Find an existing RFI that asks for files to be uploaded and the due date hasn't passed
3. Uncheck all of the boxes for the files being requested
4. On applicant dashboard the upload files button still appears
Part 2:
1. Go to RFI screen on analyst portal
2. Find an existing RFI that asks for MORE THAN ONE file to be uploaded and the due date hasn't passed
3. Uncheck ONE of the boxes for the files being requested
4. On applicant dashboard click the upload files button
5. On the upload page it still allows the user to upload the file unselected in step 3.

**Expected behaviour**
Part 1: If no files are selected then applicant should not see the upload documents button
Part 2:  If analyst unselects one file to be uploaded but they are still asking for one or more files then applicant should not have the ability to upload the file that was unselected

MarsRomer2:
Part 1 is not quite right yet.  
![image.png](https://images.zenhubusercontent.com/638e5e789d9295d36fe268e9/a467f2e8-c959-40ad-a95a-95dd928cd920)
Note above that I don't have any files selected for upload but in the below it still offers me an upload documents link on the applicant dashboard.
![image.png](https://images.zenhubusercontent.com/638e5e789d9295d36fe268e9/252192a3-09f4-459d-bade-9051d3a28535)

Part 2 works as expected! :)

marcellmueller:
@MarsRomer2 Can you please give me more information on this? I have tested this some more and I cannot reproduce it. I have gone back and forth checking and unchecking files for the RFI and the Upload files button appears when it should for me. Thanks!

Issue #1105, opened 2023-01-12
Bug: if RFI date = null, do not show upload documents button to applicant


MarsRomer2:
**Describe the bug**
When the RFI due date is null, the behavior for the applicant should be that they do not see the upload documents button on the their dashboard.  Currently it does.

**To Reproduce**
1. Go to RFI page on analyst portal
2. Click new RFI
3. Scroll down to Requested files and check off any of the files but do NOT put in a due by date
4. Save
5. ERROR is that applicant sees the upload documents button on their dashboard

**Expected behaviour**
The applicant should have upload abilities if the due date is null (or if the due date has passed)

PR #1111, opened 2023-01-13
1106 rfi files

marcellmueller:
This is branched off #1107 though we can probably just merge this one. Since we haven't made a release since the last changes to `application_has_rfi_open` no `sqitch rework` was needed.

Takes care a few issues from #1105 and #1106:
- #1105 
- #1106
- Also fixes bug if a user unchecks requested files RJSF would leave that field in the form data but set to false. So I created a function `removeFalseyValuesFromObject` to sanitize the form data of those false fields when we submit.
- Fixed a weird CSS bug on the analyst RFI side.


VB-at-Bis (approved):
Looks good

PR #1126, opened 2023-01-18
feat: add custom logout route

marcellmueller:
Implements # \<issue number\>

### DIFF
--- a/app/backend/lib/session.ts
+++ b/app/backend/lib/session.ts
@@ -24,7 +24,6 @@ const session = () => {
     secret: sessionSecret,
     store,
     resave: false,
-    // testing saveUninitialized set to false so we can totally clear cookies on logout
     saveUninitialized: false,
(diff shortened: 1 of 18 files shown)


app/backend/lib/session.ts
```diff
@@ -24,7 +24,8 @@ const session = () => {
     secret: sessionSecret,
     store,
     resave: false,
-    saveUninitialized: true,
+    // testing saveUninitialized set to false so we can totally clear cookies on logout
```
AntBush, line 27: nit: this seems to be working fine so we can either change or remove this comment

AntBush (approved):
Small nitpick about changing the comment, but looks great! Thanks for the hard work

app/backend/lib/session.ts
marcellmueller: Done. Thanks!

Issue #806, opened 2022-10-24
Remove Header/nav on the Analyst Portal


marcellmueller:
# User Story
As an analyst
I don't want to see the same navbar as the applicants
so that I don't perform actions that aren't relevant to me.

## Description
Now that we have 2 different login portals and dashboard we will need to link users to the correct dashboard in our nav menu.
We should also look at not using next router for nav links to trigger a reload since our redirection logic is handled in server side code.

## Acceptance Criteria
- [x] **No navbar for analysts**
Given that anyone is at any url beginning with `connectingcommunitiesbc.ca/analyst`
when they look at the top of the window, below the header
then there will be no navbar
#
~~- [ ] **Remove the login button in the header**
It takes the user to the keycloak page, asking to select an identity provider. This may be confusing for users, now that there is an IDIR option.~~
#
- [x] **Remove dashboard link for analyst**
The dashboard link in the header leads to the applicant's dashboard therefore it need to be removed _for analysts_
#
Includes e2e testing
#

![dashboard link.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/cfe417d5-a0b7-4861-ac4f-a168978b337e)

![Screen Shot 2022-12-15 at 1.33.17 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/181ca300-5456-4b49-a73a-00c0126824ba)

![Screen Shot 2022-12-15 at 1.33.18 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/6c0b065a-49d7-4394-a296-2a9bf11a5452)

## Technical Analysis:

- What is the front-end complexity?
- Do it require creating or updating a component?
- Any work on the back-end?
- Integration?
- Any changes in database structure?
- What types of tests that we need to create and run?
- Impacts on DevOps?
- Other risks and potential side effects?
- Which documentation needs to be updated?

eazeredo:
@MarsRomer , is this bug still valid?

MarsRomer:
yes

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/806?planning-poker) with Zenhub @VB-at-Bis @marcellmueller @AntBush

MarsRomer2:
[...]

PR #1129, opened 2023-01-20
806-analyst-header

marcellmueller:
Implements #806

I also added a redirect for logged in analysts/admin on the homepage `/applicantportal` route. It now redirects them to the analyst side as the dashboard button led to an error page (due to lack of permissions) as well as the issues with the nav we were trying to fix here.


VB-at-Bis (approved):
Looks good

PR #1147, opened 2023-01-24
feat: add auth_role to session

marcellmueller:
This was a small research spike turned into a PR for upcoming tickets. Adds the current `auth_role` to the session.

session


VB-at-Bis (approved):
Perfect!

Issue #1043, opened 2022-12-19
Export Attachment refactoring 


MarsRomer2:
Problem statement:
- based on existing process flow, Export Attachment should happen once after the intake is closed, and files from the downloaded archive is manually copied to SharePoint;
- existing implementation is based on real-time streaming files from S3, compressing data on-the-fly and streaming archive to the user directly from browser;
- due to yet unknown reasons current approach is failing in Production environment , and succeeded only once when it was executed during off hours;
- reasons could be either code issues, or AWS SDK (i.e. https://github.com/aws/aws-sdk-js/issues/2087), or OpenShift limits etc. Initial investigations did not identified exact cause or suspect;

Discussion points:
- should we spend time on further investigations (no tome/effort estimate possible);
- should we change the process flow and run archive generation not in real time (i.e. present message to the user that the request to generate export file is received, and read files from S3, add to archive, upload archive back to S3 and send user an email with pre-signed download URL);
- or should we think about any other approach?

Acceptance Criteria:
- [x] Zip file is created by cron job after the status is changed to 'Received' once the intake closes - zip file stored on S3 bucket
- [x] Zip file will have attachments for 'Received' applications for the intake that just closed, not all the intakes prior to that
- [x] Export Attachment button will download zip file from S3 bucket

Behavior of 'Export Attachment' button will be changed as part of #1120 (to download zip file with attachments for specific intake). 
P.S. Title of the ticket updated to reflect solution to the issue (Export attachment no working)

eazeredo:
Hey @VB-at-Bis , if you are working on this, can you move it to In Progress?

eazeredo:
The files in the zip file are not showing on the Windows default zip extractor.

PR #1160, opened 2023-01-26
feat: updated lambda and terraform 

VB-at-Bis:
Implements #1043

- updated lambda to use S3 events;
- updated terraform to remove SNS topic and subscribe lambda to S3 events instead;


marcellmueller (approved):
Nice work :partying_face:

Issue #1119, opened 2023-01-16
Add Admin Tab


MarsRomer2:
##
As an admin, 
I want to have a dedicated section where I can do all my necessary actions that's separate from the applications 
So that the UI is organized in a logical way.

## Acceptance Criteria
![Admin dashboard.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/f8b612fe-7302-4780-9966-b4ac26323c68)

### Create 1st level of tabs
- [x] The Dashboard screen now has "Admininstrative" tab as per screen design above
- [x] Analysts view does not have underline of Dashboard

**Validations**
- [x] Only show admin tab if they are admin - if easy (otherwise everyone sees the tab but only admin can access it)
- [x] Admin has all the same access as Analyst plus access to the admin tab/functionality

**Actions**
- [x] Clicking Administrative tab leads to Download Attachments subtab
- [x] Clicking Dashboard tab leads to Analyst/Admin Dashboard
#
![Admin pages.jpg](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/298d4908-447b-4d85-9388-0a9302c64128)

### Create 2nd level of tabs
- [x] Within the Administrative tab, there will be 3 subtabs as per design above
- [x] Admin can click on each subtab to view its contents (Each tab will only be a placeholder for future tickets and therefore not have any content.)

**Screen Behaviour**
- [x] When Clicking a grey sub-tab changes it to bold & 313132 and removes the bottom border so that it looks like a paper tab

**Analyst view**
<img src="https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/c8b8e5aa-effa-42d0-995c-287bac399b43" width=70%>

elliottSlee:
@MarsRomer2 card updated to reflect design changes

elliottSlee:
On the subtabs one minor thing is missing when selected "removes the bottom border so that it looks like a paper tab"

@marcellmueller

marcellmueller:
> On the subtabs one minor thing is missing when selected "removes the bottom border so that it looks like a paper tab"
> 
> @marcellmueller

Weird, that definitely got put in there as it is more difficult to do in css than you would think. Though I see it’s not working in test, it is working fine in dev.

carreenu:
Functionality is working as described.

PR #1162, opened 2023-01-27
1119 add admin tab

marcellmueller:
Implements #1119

This also updates the route authorizations. Previously `ccbc_analyst` and `ccbc_admin` users could access any route in the `/analyst/(.*)` side. Now analysts can only access `/analyst/dashboard` and anything in `/analyst/application/(.*)` so that only admins can access `/analyst/admin/(.*)`.

Not sure if there is a better way to do that so we don't have to update this if we add an `analyst/some-other-page` route.


VB-at-Bis (approved):
Looks good!

Issue #1146, opened 2023-01-24
Bug:  Submission Page shows 'Invalid DateTime'


MarsRomer:
**Describe the bug**
When no intake is currently open, the submission page shows 'Invalid DateTime'

**To Reproduce**
1. Go to Applicant portal
2. Click on any project where you can go to the submission page
3. on Time Machine set date to Jan 13 2023
4. See error:  After submission, you can continue to edit this application until the intake closes on Invalid DateTime

**Expected behaviour**

- [x] If the application was submitted, and the intake has closed then the closing date of the intake in which this application was submitted shows.

 - [x] if the application was submitted and the intake is still open then the closing date of the current  shows.

- [x] If the application has not submitted, and no intake is currently open then the sentence reads: 
After submission, you can continue to edit this application until the intake closes.

- [x] If the application has not been submitted, and no intake is currently open then the sentence reads:
After submission, you can continue to edit this application until the intake closes.

**Screenshots**
![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/1e2ae52f-2de1-4464-a8ba-b368b21852fd)

**Logs/tracebacks**

**Additional context**

carreenu:
Functionality demonstrated by Elliot and bug is fixed. Move to Sprint Done please.

PR #1166, opened 2023-01-27
1146 invalid datetime

marcellmueller:
Fixes bug #1146 

Pretty sure this was made before we had `received` applications so I've removed the `submitted` check to show the `intakeByIntakeId` if it exists.

I'm not sure if we will encounter this scenario unless using the time machine but if there is no open intake or no `intakeByIntakeId` we don't show a date at all.

Also I moved the submission form page tests to their own file as the `ApplicationForm` tests were getting out of control.


AntBush (approved):
LGTM!

Issue #1021, opened 2022-12-15
Error logging in export attachment function


AntBush:
As an analyst
I want to see any errors created in the Export Attachments feature, if they are to occur, in a text file
So that I can ask the devs to fix the problems

## Acceptance Criteria:

- [x] **Scenario: Error message structure**
Errors detected - please send this file to meherzad.romer@gov.bc.ca
[Todays timestamp]

These files were not included in the downloaded zip file due to errors:
[Error number] - [filename on S3] - [file name in db] - [filetype] - [size on S3] - [size according to database] - [CCBC-ID] - [upload date and time]  - [Error timestamp]
Error Details:
[error messages for developers to decipher]

[Error number] - [filename on S3] - [file name in db] - [filetype] - [size on S3] - [size according to database] - [CCBC-ID] - [upload date and time]  - [Error timestamp]
Error Details:
[error messages for developers to decipher]

[Error number] - [filename on S3] - [file name in db] - [filetype] - [size on S3] - [size according to database] - [CCBC-ID] - [upload date and time]  - [Error timestamp]
Error Details:
[error messages for developers to decipher]

- [x] if no errors found then text file says 'Download successfull'
.
.
.
#
- [x] **Error code definition**
     The error code maps to an enumerator in our control system.
#
- [x] **Scenario: Capture the errors in the Export Attachments feature**
    Given there are one or more errors in the Export Attachments feature
    When I try to export 
    Then each error captured will be saved 
    And these errors will be saved in the `errors.txt` with the logs
#
- [x] **Scenario: Add the error log file to the ZIP file**
    Given I try to Export Attachments on the Dashboard
    When an error occurs in one or more files
    Then an `errors.txt` file that includes a list of filenames and their respective errors description is added to the ZIP file.

## Not included:
-
-

## Ultra refinement
DB:
- no updates in the database.

Middleware:
- [x]  Map the error code in the JSON

Frontend:
- no updates in the front end. 

Integration/API:
- no integration.

Tests:
- [x] yes, ...

Documentation:
- [x] Create the Error code definition.

AntBush:
The errors.txt I think will be moreso for the devs, so that if one exists it can be given to the developers/us to look into.

MarsRomer2:
@marcellmueller @AntBush @VB-at-Bis @pdoc1111 please add a list of the errors that we do know we may get to this ticket so that I can add the desired wording

marcellmueller:
[...]

PR #1187, opened 2023-02-03
feat: pass more information about attachments

VB-at-Bis:
Implements #1021

### DIFF
--- a/Infrastructure/arc_lambda/index.js
+++ b/Infrastructure/arc_lambda/index.js
@@ -13,16 +13,18 @@ exports.handler = async(event, context, callback) => {
     return;
   }
   const record = event.Records[0];
-  if (record.s3.object.key.indexOf('Intake_')>-1) {
+
+  if (record.s3.object.key.indexOf('Intake_')>-1 && record.s3.object.key.indexOf('json')>-1) {
     const s3Object = await s3
     .getObject({
       Bucket: record.s3.bucket.name,
       Key: record.s3.object.key
     })
     .promise();
-    const message = s3Object.Body
+    const message = s3Object.Body;
     console.log(`Got list of attachments from S3; file size: ${message.length}`);
     const list = JSON.parse(message);
+    const issues = [];
     const key = record.s3.object.key.replace('.json','.zip');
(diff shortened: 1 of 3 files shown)


Infrastructure/arc_lambda/index.js
```diff
@@ -43,21 +45,82 @@ exports.handler = async(event, context, callback) => {
       output.on('end', function() {
         console.log('Data has been drained');
       });
-      
+      output.on('error', function (err) { 
+        console.log('[OUTPUT STREAM ERROR]', err);
```
wenzowski, line 49: can multiple output events be dispatched? what else can we do with the error other than logging to stderr? should we exit the function?

Infrastructure/arc_lambda/index.js
```diff
@@ -43,21 +45,82 @@ exports.handler = async(event, context, callback) => {
       output.on('end', function() {
         console.log('Data has been drained');
       });
-      
+      output.on('error', function (err) { 
+        console.log('[OUTPUT STREAM ERROR]', err); 
+      });
+
+      archive.on('error', function(err) {
+        console.log('[ARCHIVER ERROR]', err);
```
wenzowski, line 53: can multiple archive events be dispatched? what else can we do with the error other than logging to stderr? should we abort the response with a failure http status code?

Infrastructure/arc_lambda/index.js
VB-at-Bis: This is lambda function, so if we can create archive, we'll add all error logs into that errors.txt; if we fail to create archive then we'll check CloudWatch. This operation (create archive) happens once when intake is closed (invoked by the cron job), so I don't think we need to complicate the function by adding push to SNS to send email etc.

Infrastructure/arc_lambda/index.js
VB-at-Bis: This is lambda function invoked by cron job. Http code would be ignored anyway :) 
I will add documentation that describes steps how we close intake (two cron jobs need to run - one marks applications as received, second generates the archive of attachments)

marcellmueller (approved):

PR #1195, opened 2023-02-06
feat: add sentry custom spans for file uploads

marcellmueller:
This PR adds some Sentry custom spans to better track file uploads.


VB-at-Bis (approved):
Perfect!

PR #1198, opened 2023-02-06
fix: admin tabs border top css

marcellmueller:
Fix for the admin tabs Elliott pointed out for the bottom border of the selected admin tab wouldn't disappear. 

Before:
before

After:
after


VB-at-Bis (approved):
LGTM

PR #1213, opened 2023-02-09
fix: archive structure

VB-at-Bis:
Implements #1021

Updated lambda code to fix an issue with opening archive on Windows 11 where folders in the archive were not handled properly due to leading slash /


AntBush (approved):
Tried in test, working well!

Issue #1170, opened 2023-01-30
1110 - Improve health/readiness check 


VB-at-Bis:
When new pod is started, Kubernetes is checking if pod is alive and can serve traffic.
We need to improve readiness check by verifying if pod is  ready not only to serve traffic but also perform all DB and AWS actions. 
For example, code can execute simple SQL query to get list of intakes and download some pre-defined file from S3 (i.e. `freshclam.dat` from `fapi7b-XXX-ccbc-clamav` S3 bucket - file size is ~1K)

wenzowski:
The context of this ticket came from a discussion about potential slowness due to pods accepting traffic before being fully initialized. This sounds like an opportunity to improve the readiness check to ensure nextjs is fully loaded in addition to nodejs being booted. The liveness/readiness checks are already intended to be following expressjs best practices as we are using the [lightship](https://github.com/gajus/lightship#best-practices) library. Worth remembering that k8s best practice is to not check dependent services via readiness/liveness otherwise we potentially end up with race conditions. So this card will need to consider how we'll identify connection outages with postgres/s3 but should not include rebooting pods via liveness probe as a result of identified outages.

wenzowski:
Maybe the use case here isn't the health/readiness check per-se and more the warmup process itself? Here's [a sidecar pod example configuration the expedia team uses](https://opensource.expediagroup.com/mittens/docs/installation/running#run-as-a-sidecar-on-kubernetes) to help warm up app pods.

PR #1224, opened 2023-02-14
1170-readiness-check

AntBush:
Implements #1170


VB-at-Bis (approved):
Looks good; we could avoid adding new variable AWS_CLAM_S3_BUCKET by using AWS_S3_BUCKET (replace `-data` with `-clamav`)

PR #1226, opened 2023-02-14
fix: sort on analyst lead no longer returns empty string

AntBush:
Implements fix for column sorting


marcellmueller (approved):
Looks good! :ship:

PR #1234, opened 2023-02-15
fix: add large initial pagesize

AntBush:
Implements fix for page size on dashboard


marcellmueller (approved):
Nice work!

marcellmueller:
@AntBush I think it just needs `pageSize` removed from the default query variables in `analyst/dashboard.test.tsx` removed.

PR #1238, opened 2023-02-16
feat: add intake column to dashboard

AntBush:
Implements # 1167

Checked with Elliot for design changes


VB-at-Bis (approved):
Looks good

PR #1242, opened 2023-02-16
feat: enable graphql persisted operations

marcellmueller:
Implements graphql persisted queries so that we can use mitigate risks from using graphql filter connection plugin and also get a nice performance boost. This should work seamlessly for developers.

With `persistConfig` set in the relay options in our `package.json` the relay compiler will generate `schema/queryMap.json`. Afterwards the `yarn build:relay` script runs `utils/addToPersistedOperations.js` which splits the operations in `queryMap.json` into separate files stored in `.persisted_operations` which are named after the hashes we send instead of sending the entire query body over.

The postgraphile fetch operation had to be modified to send the id which is this hash to identify the query instead of the entire query body.

## Important:
 this will always work on a fresh project but I noticed that sometimes when switching back from branches without persisted queries I would have to delete my `__generated__` folder for relay compiler to populate `queryMap.json` and `.persisted_operations`. If you have any issues running this the first time delete that folder and run `yarn build:relay` again :)


VB-at-Bis (approved):
Looks good

PR #1255, opened 2023-02-23
fix: add flag for multiple uploads on widgets

AntBush:
Implements multiple uploads and file types for some fields


marcellmueller (approved):
Looks good!

PR #1256, opened 2023-02-23
fix: create extension 

VB-at-Bis:
Implements #1210

- fixed deployment issue


helm/app/templates/deployment.yaml
```diff
@@ -85,6 +85,7 @@ spec:
               psql -d $(DATABASE_NAME) <<EOF
                 grant pg_read_all_data to $(PGUSER_READONLY);
                 alter role $(PGUSER_READONLY) BYPASSRLS;
+                create extension if not exists "uuid-ossp";
```
AntBush, line 88: Is it for the uuid or is it failing on creating the audit extension?

AntBush (approved):

PR #1261, opened 2023-02-23
1240-ccbc-id-sequential

AntBush:
Implements # \<issue number\>


db/verify/tables/intake_004_drop_trigger.sql
```diff
@@ -0,0 +1,7 @@
+-- Verify ccbc:tables/intake_004_drop_trigger on pg
+
+BEGIN;
+
+-- XXX Add verifications here.
```
VB-at-Bis, line 5: Please remove files from /verify/

VB-at-Bis (commented):
we agreed do not create /verify/ sqitch files

db/verify/tables/intake_004_drop_trigger.sql
AntBush: Good catch! Fixed that up

VB-at-Bis (approved):
Looks good!

PR #1274, opened 2023-02-24
fix: add uncaughtException handler with sentry

AntBush:
Implements handler for uncaught exceptions, should prevent pod from crashing


VB-at-Bis (approved):
LGTM

PR #1276, opened 2023-02-24
docs: add table of contents and improve project setup documentation

marcellmueller:
Started a table of contents and improved some of the development setup documentation. Let me know what I can do to improve this thanks.

I think the one area of this particular ticket that could use some work is the environment variables section. Need to figure out which env vars are absolutely required to start development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,37 +2,81 @@
 
 # CONN-CCBC-portal
 
-## Database Setup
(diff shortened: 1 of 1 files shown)


VB-at-Bis (approved):
Nice! README looks much more useful now

Issue #1209, opened 2023-02-09
801 - History:  Backfill data for existing data


eazeredo:

VB-at-Bis:
Steps to backfill data :

1.  run `kubectl exec -it ccbc-hippo-ha-xzsq-0 -- /usr/pgsql-14/bin/pg_dump ccbc --inserts --on-conflict-do-nothing > backup.sql`

2. Edit the backup file:
- remove data for next tables :
ccbc_private.connect_session
ccbc_public.analyst 
ccbc_public.form_type (name, description) 
ccbc_public.form (id, slug, json_schema, description, form_type)
ccbc_public.form_data_status_type (name, description) 
ccbc_public.application_status_type (name, description, visible_by_applicant, status_order) 
ccbc_public.assessment_type (name, description)

and all sqitch.* tables




 
- add next statements before the `ccbc_public.users` data:
truncate ccbc_public.ccbc_user cascade;

ALTER TABLE ccbc_public.application DISABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.application_status DISABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.attachment DISABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.assessment_data DISABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.rfi_data DISABLE TRIGGER _100_timestamps;

- add next statements in the very end of the file:
ALTER TABLE ccbc_public.application ENABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.application_status ENABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.attachment ENABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.assessment_data ENABLE TRIGGER _100_timestamps;
ALTER TABLE ccbc_public.rfi_data ENABLE TRIGGER _100_timestamps;

3. Run 
psql -d ccbc -a -f backup.sql

Applying backup will generate historic records in `ccbc_public.record_version` table.

eazeredo:
Should we have a SQL script with this code and run it?

VB-at-Bis:
It will be that's backup.sql that will be created on step 1 - after we deploy new extension/triggers (other PR for #1210). There is no sense to create it before, or check it into GitHub (it would contain all data entered by the customers and analysts).
Manual step, that should be done once :)

PR #1282, opened 2023-02-24
feat: backfill history

VB-at-Bis:
Implements #1209 

- created functions to populate record_version table with existing records 
- no duplicate records are created in record_version table;
- after deployment import function should be run once;

### DIFF
--- a/db/deploy/import/attachment.sql
+++ b/db/deploy/import/attachment.sql
@@ -1,0 +1,58 @@
+-- Deploy ccbc:import/attachment to pg
+
+BEGIN;
+
+create or replace function ccbc_public.import_attachments() returns table(result_id int)
+as $function$
+    declare
+        current_app    ccbc_public.attachment%ROWTYPE;
+        cnt             int;
+        table_oid       int;
+	    attachments    refcursor;
+        pkey_cols       text[];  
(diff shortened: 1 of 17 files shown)


db/deploy/import/attachment.sql
```diff
@@ -0,0 +1,58 @@
...
+                where table_name='attachment' and op = 'INSERT');
+        
+    end loop;
+
+    close attachments;
+
+    return query select cnt as result_id; 
+end;
+$function$ language plpgsql volatile;
+-- grant execute on function ccbc_public.import_attachments to ccbc;
```
AntBush, line 57: I notice we have this commented out in every, I'm guessing it's because the `ccbc` account has execute access already? I think it's also not available in our dev environment

AntBush (approved):
Looks good! Tried it on my local and works well. We should probably implement a change in the future to change from the insert-only flow we've got to doing updates. Will make reading tables much easier

db/deploy/import/attachment.sql
VB-at-Bis: It is commented out because  `ccbc` is not available :( .. Should have removed the comment :)

PR #1283, opened 2023-02-24
1039 view and add analysts

marcellmueller:
- feat: add insert and update permissions for analyst table
- feat: add list of analysts to admin dashboard
- feat: add create and update analyst mutations

Implements # \<issue number\>


VB-at-Bis (approved):
Nice

PR #1287, opened 2023-02-28
intake filtering

marcellmueller:
- feat: enable intake column filtering
- fix: remove empty filter args

Implements # \<issue number\>


VB-at-Bis (approved):
Looks good

Issue #1285, opened 2023-02-27
Add Karina to list of analysts


MarsRomer:
As an analyst
I want the new analysts name to display in the drop downs 
So that I can assign them assessments and leads

**Description:**
Add Karina Boarato to the list of analysts that shows up in the drop downs in the assessments and lead on the analyst portal.

# Acceptance Criteria:
- [ ] List of analysts drop down includes Karina Boarato and can be used when assignment assessments and lead in the analyst portal

PR #1289, opened 2023-02-28
feat(db): add karina to list of analyst names

AntBush:
Implements #1285

Adds Karina to the list of Analyst Names


AntBush:
Name was manually added

Issue #1211, opened 2023-02-09
801 - History: Front-end implementation


eazeredo:

elliottSlee:
![Screen Shot 2023-03-07 at 9.52.58 AM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/1da915ba-76f2-49f2-b67a-38890351b5ae)
Times are incorrect in 2 ways:
(1) The actions in the screenshot were taken at different times but all display as 9:00pm
(2) time zones??? The actions in the screenshot were complete around 9:45am EST, thus they should display as 6:45am PT.

The order of events in the timeline is not correct, which could be a result of the above time issues.

MarsRomer:
Request:  Based on some user testing we did this morning, all users expected to see the most recent history at the top of the page, so can we revert back to sorting descending please?

PR #1296, opened 2023-03-01
feat: computed function for application history

VB-at-Bis:
Implements #1211

created function to return audit records related to application. Includes:
- record when application was created;
- records when application status was changed;
- records when attachments were added to the application;
- records when assessment entries were created;
- records when RFI entries were created;


marcellmueller (approved):
This is working well, nice work.

PR #1299, opened 2023-03-01
fix: inactive analyst is shown for select

AntBush:
Implements # 1295


marcellmueller (commented):
@AntBush nice solution, this is working really well. I did notice a small bug that should be fixed where if you unassign the analyst it now shows up blank instead of showing `unnassigned` as before.

AntBush:
> @AntBush nice solution, this is working really well. I did notice a small bug that should be fixed where if you unassign the analyst it now shows up blank instead of showing `unnassigned` as before.

Good catch, I've added a fix!

marcellmueller (approved):
Woo thanks! 🚀

PR #1302, opened 2023-03-01
session expiry

marcellmueller:
- feat: add on mousedown event to session expiry handler
- fix: add next router to session expiry handler resetOnChange

Implements # \<issue number\>

### DIFF
--- a/app/components/SessionExpiryHandler.tsx
+++ b/app/components/SessionExpiryHandler.tsx
@@ -62,6 +62,7 @@ const SessionExpiryHandler: React.FC = () => {
         <SessionTimeoutHandler
           modalDisplaySecondsBeforeLogout={120}
           onSessionExpired={handleSessionExpired}
+          resetOnChange={[router]}
           extendSessionOnEvents={{
             enabled: true,
             throttleTime: 200000,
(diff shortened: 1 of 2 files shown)


app/pages/_app.tsx
```diff
@@ -78,6 +76,7 @@ const MyApp = ({ Component, pageProps }: AppProps) => {
         <BCGovTypography />
         <Sentry.ErrorBoundary fallback={<Error500 />}>
           <RelayEnvironmentProvider environment={env}>
+            {typeof window !== 'undefined' && <SessionExpiryHandler />}
```
marcellmueller, line 79: Not sure if this will do anything though I checked the `cas` repo and this is where they have it.

app/components/SessionExpiryHandler.tsx
```diff
@@ -62,10 +62,11 @@ const SessionExpiryHandler: React.FC = () => {
         <SessionTimeoutHandler
           modalDisplaySecondsBeforeLogout={120}
           onSessionExpired={handleSessionExpired}
+          resetOnChange={[router]}
```
marcellmueller, line 65: This listens for next router changes and refreshes the session on each router change. This *should* severely mitigate the invalid session modal popup.

marcellmueller, line 65: This listens to changes in next router and refreshes the session. This alone *should* significantly mitigate the issue 🤞

AntBush (approved):
Looks safe, and keeping the timeout extended based on the router makes sense as well

PR #1308, opened 2023-03-03
1211 history front end

marcellmueller:
- feat: add basic history table and icons
- fix: history last child border
- feat: add history content
- feat: add more application history content types

Implements # \<issue number\>


VB-at-Bis (approved):
Perfect!

PR #1312, opened 2023-03-03
fix: use url query param instead of next router

marcellmueller:
Looks like we missed a minor issue where the first click of a sort column would not save the query param though it would save fine after that. For some reason next router was behind, though I am just grabbing it from the url now which seems to be working great.


marcellmueller:
@AntBush hey give this a try when you get a moment, Mars said it wasn't working though it was just the first click that wasn't saving as far as I can tell. This should fix that.

AntBush (approved):
Tested with my cookies cleared, works on first click

PR #1318, opened 2023-03-06
chore: deploy minimum app

AntBush:
Implements #  1178


AntBush:
Looks like this investigation is over

PR #1327, opened 2023-03-07
fix: check if what is passed is error before rethrowing

AntBush:
After looking at the implementation details of Upload.js, there seems to be an issue with overloading the method in the way I did. 

```javascript
    on(event, listener) {
        this.uploadEvent = event;
        return super.on(event, listener);
    }
```
Since the order of the call of `on` uses `uncaughtException`, later the notifyProgress function emits the same event assigned to `uploadEvent`. 
```javascript
    __notifyProgress(progress) {
        if (this.uploadEvent) {
            this.emit(this.uploadEvent, progress);
        }
    }
```

This reordering should prevent `notifyProgress` here to emit an `uncaughtException` 

I've also added a guard around the caught exception

### DIFF
--- a/app/backend/lib/graphql/resolveFileUpload.ts
+++ b/app/backend/lib/graphql/resolveFileUpload.ts
@@ -67,6 +67,13 @@ export const saveRemoteFile = async (stream) => {
       leavePartsOnError: false, // optional manually handle dropped parts
     });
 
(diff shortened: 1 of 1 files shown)


VB-at-Bis (approved):
Looks safe

PR #1333, opened 2023-03-08
1248 dashboard page position

marcellmueller:
- feat: save and set analyst dashboard scroll position
- feat: clear scroll position session storage on logout

Implements # \<issue number\>


AntBush (approved):
Good work!

PR #1343, opened 2023-03-14
feat: add logic for dynamic toggling in according

AntBush:
Implements # 1168


VB-at-Bis (approved):
Looks good (please see ESLint warning though)

app/components/Review/Accordion.tsx
```diff
@@ -112,7 +112,7 @@ const Accordion = ({
   // ugly and I don't like it
   useEffect(() => {
     setIsToggled(getToggledState(toggled, defaultToggled));
-  }, [toggled]);
+  }, [toggled, defaultToggled]);
```
VB-at-Bis, line 115: I, personally, find it weird to add to dependencies something that does not get updated in the component but only through props (meaning component re-render anyway) ... oh, those Lint rules

PR #1348, opened 2023-03-15
feat: add sentry data tags to get clearer picture

AntBush:
Implements nothing in particular, just thought it would be useful after looking at the sentry log info. Can help shed light as to whether this is a problem with our appilcation's resources (is our app hanging waiting to communicate with AWS or is it just taking a long time to setup the connection because the app in general runs slowly)

I'd love to add some log info of the pod load at the same time, but I'm not sure how to get that data in our app


marcellmueller (approved):
🚢

PR #1350, opened 2023-03-15
feat: remove all history items before received and reverse history order

marcellmueller:
Implements # \<issue number\>


VB-at-Bis (approved):
Looks clean

Issue #1352, opened 2023-03-15
Fix Export attachments AWS configuration


VB-at-Bis:
Existing implementation of AWS lambda and terraform scripts did not work properly due to the AWS rules regarding lambda notifications attached to S3 buckets - two lambdas cannot be subscribed to same event, and we have both "virus scanner" and "export attachment" lambdas using same bucket.
Code need to be updated to use different buckets to avoid the conflicts.

PR #1353, opened 2023-03-15
feat: changed bucket for list of attachments

VB-at-Bis:
Implements #1352

- updated Terraform scripts to attach lambda to different bucket;
- updated backend code to save list of attachments to different bucket;


marcellmueller (approved):
Looks good

PR #1428, opened 2023-04-05
fix(test): use mocks for creating rfi

AntBush:
Implements fix for spurrious diffs on happo screenshots for the applicant RFI page


marcellmueller (approved):
Nice!!

Issue #1393, opened 2023-03-27
Update updated at & archived by column


elliottSlee:
As a data analyst
I want to know more information about the updated at, archived at, and archived by columns in the database tables
So that 

**Description:**

# Acceptance Criteria:

## Implementation Details
- **DB:**
  - .
  - .
- **Middleware:**
  - .
  - .
- **Frontend:**
  - .
  - .
- **Tests:**
  - .
  - .
- **Documentation:**
  - .
  - .

elliottSlee:
@VB-at-Bis Here's the card

MarsRomer:
Ticket needs a spike to make sure all relevant tables have the archived by column

VB-at-Bis:
No other tables without `archived_at` columns in DB.

MarsRomer:
@VB-at-Bis  Is it the Application_status table that you added the 'Archived at' column?    I don't see any data in that column, should there be?
![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/fc53df26-49a3-4ee3-90c0-26f01f69d1cc)

VB-at-Bis:
Can you please try again? Did you scroll table? It seems most of first applications are left in draft status etc and were never archived.

MarsRomer:
I see data in there now, thanks!

PR #1436, opened 2023-04-06
feat: added archive columns and backfill

VB-at-Bis:
Implements #1393

- add archived_at/archived_by columns to application_status table;
- back fill archived_at/archived_by values using ccbc_public.record_version;
- add trigger to application_status table to archive old record when new is added;

### DIFF
--- a/db/deploy/util_functions/upsert_archive_trigger.sql
+++ b/db/deploy/util_functions/upsert_archive_trigger.sql
@@ -10,9 +10,6 @@ create or replace function ccbc_private.upsert_archive_trigger(
 returns void as $$
 
 declare
-  column_string text;
-  index_string text;
-  comment_string text;
   trigger_string text;
 
 begin
--- a/db/deploy/util_functions/upsert_archive_trigger.sql
+++ b/db/deploy/util_functions/upsert_archive_trigger.sql
@@ -5,9 +5,7 @@ begin;
 
 create or replace function ccbc_private.upsert_archive_trigger(
   table_schema_name text,
-  table_name text,
-  user_table_schema_name text default 'ccbc_public',
-  user_table_name text default 'ccbc_user'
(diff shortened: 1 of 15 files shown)


db/deploy/util_functions/upsert_archive_trigger.sql
```diff
@@ -0,0 +1,53 @@
+-- Deploy ccbc:util_functions/upsert_archive_trigger to pg
+-- requires: trigger_function/set_archived
+
+begin;
+
+create or replace function ccbc_private.upsert_archive_trigger(
+  table_schema_name text,
+  table_name text,
+  user_table_schema_name text default 'ccbc_public',
+  user_table_name text default 'ccbc_user'
```
AntBush, line 10: I notice `user_table_schema_name` and  `user_table_name` isn't used, should probably delete it but won't block approval

db/deploy/util_functions/upsert_archive_trigger.sql
```diff
@@ -0,0 +1,53 @@
...
+)
+returns void as $$
+
+declare
+  column_string text;
+  index_string text;
+  comment_string text;
```
AntBush, line 17: Looks like some of these variables are also unused

AntBush (approved):
Other than a few minor changes looks great!

db/deploy/util_functions/upsert_archive_trigger.sql
VB-at-Bis: indeed  user_table_schema_name and user_table_name isn't used. I deleted them.

db/deploy/util_functions/upsert_archive_trigger.sql
VB-at-Bis: Removed

Issue #1453, opened 2023-04-12
Change GIS Import to use row per CCBC ID and GIS Analysis


AntBush:
Change the GIS analysis import such that the JSON data is done row-by-row in the DB rather than storing the entire JSON array as a column. 

@VB-at-Bis feel free to edit and add some details

## Acceptance criteria
- [x] `application_gis_data` table exists in Metabase 
- [x] `parse_gis_data` mutation exists and can be called to populate  `application_gis_data` table with data from `gis_data` table

PR #1455, opened 2023-04-12
1453 GIS Import to use row per CCBC ID

VB-at-Bis:
Implements #1453

### DIFF
--- a/app/backend/lib/gis-upload.ts
+++ b/app/backend/lib/gis-upload.ts
@@ -76,41 +76,6 @@ const parseForm = (form, req) => {
   });
 };
 
-const persistRowsOld = async (batchId, json_data, req) => {
-  const error_list = [];
-  for (const row of json_data) {
-    try {
-      const { ccbc_number } = row;
-      const applicationId = ccbc_number ? parseInt(ccbc_number.substring(7),10) : null;
-      await performQuery(saveApplicationGisDataMutation, {applicationId, batchId, input: req}, req)
-      .catch((e) => {
-        error_list.push(e);
-      });
-    }
-    catch(e){
-      error_list.push(e);
-    };
-  }
-  return Promise.resolve(error_list);
-};
-
-const persistRows = async (batchId, json_data, req) => {
-  const error_list = [];
-  await json_data.reduce(async (promise, row) => {
-    // This line will wait for the last async function to finish.
-    // The first iteration uses an already resolved Promise
-    // so, it will immediately continue.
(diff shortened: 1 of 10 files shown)


db/deploy/mutations/parse_gis_data.sql
```diff
@@ -0,0 +1,47 @@
...
+    -- Loop over each object in the JSON array
+    for json_row in select * from jsonb_array_elements(json_file)
+    loop
+        ccbc_id := json_row->>'ccbc_number'; 
+
+        -- Look up the application_id based on the ccbc_number value
+        select id into application_id from ccbc_public.application where ccbc_number = ccbc_id;
+        -- If no row was found, add to list 
+        if not found then
+            result := coalesce(result || ',','') || ccbc_id;
```
AntBush, line 31: Are we planning on saving even if the file has non-valid `ccbc_id`s?

AntBush (commented):
Looks good! Just a question or two, with the important one being the behaviour of parse_gis_data

AntBush (approved):
After changing the stringify/parse, should be good to go. Good stuff!

app/backend/lib/gis-upload.ts
```diff
@@ -92,12 +96,19 @@ gisUpload.post('/api/analyst/gis', limiter, async (req, res) => {
   }
    
   // time to persist in DB
-  await performQuery(saveGisDataMutation, {input: data}, req)
+  const result = await performQuery(saveGisDataMutation, {input: data}, req)
   .catch((e) => {
     return res.status(400).json({ error: e }).end();
   });
+  
+  const newRow = JSON.parse(JSON.stringify(result));
```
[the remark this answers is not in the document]
VB-at-Bis: Typescript throws error:
```
Property 'data' does not exist on type 'Response<any, Record<string, any>, number> | ExecutionResult<{ [key: string]: any; }, { [key: string]: any; }>'.

Property 'data' does not exist on type 'Response<any, Record<string, any>, number>'.ts(2339)
```
I guess its' because it's result from mutation, not query.

db/deploy/mutations/parse_gis_data.sql
VB-at-Bis: Yes, file already saved anyway (in `gis_data` table. List of non-valid ccbc_ids returned from the function, so UI can display if needed. I used Prod JSON in Test environment, and, as expected, some ccbc_id were not found.

PR #1461, opened 2023-04-14
fix: updated dependency

VB-at-Bis:
Fixes Dependabot alert 28 https://github.com/bcgov/CONN-CCBC-portal/security/dependabot/28 (weakness CWE-1321)


eazeredo:
Vlad, is this ready for review?
If yes, can you move it to the Dev Review column?

AntBush (approved):

PR #1467, opened 2023-04-17
fix: update sqitch.plan

VB-at-Bis:
set correct chronological order


AntBush:
We can close this right?

PR #1478, opened 2023-04-18
fix: replace schemaspy snapshot tag with latest

marcellmueller:
Replaced the `snapshot` tag with `latest`. Snapshot is the latest development build which is volatile by nature, while latest is the latest release and only behind by a few weeks at the moment. If this issue comes up again we can lock in a specific version in the future.

Successful test run:

https://github.com/bcgov/CONN-CCBC-portal/actions/runs/4736614960/jobs/8408408417


AntBush (approved):
:shipit:

PR #1480, opened 2023-04-19
fix: update archiver to use new role for archiving

AntBush:
Implements # \<issue number\>


marcellmueller (approved):
Nice work, this seems to be working well.

PR #1482, opened 2023-04-19
feat: update relay store for external status change

marcellmueller:
External status in header should update instantly on change now


AntBush (approved):
Looks good, thanks!

Issue #1466, opened 2023-04-17
Provide ISED access to the Portal with Business BCeID


MarsRomer:
As an ISED member
I want to access the analyst portal
So that I can better collaborate with NWBC on the applications and projects

**Description:**
This ticket is to provide ISED with access to the Analyst Portal using their Business BCeID

# Acceptance Criteria:
- [x] Get access to a test Business BCeID account for the team 
- [x] ISED can log in with their Business BCeID (use login button in top right corner of login)
- [x] ISED can view and make changes the same as the analysts role
- [x] History table shows the name of ISED person for 'who' made changes

Note:  https://github.com/bcgov/sso-keycloak/blob/dev/docs/migration-guide.md 
## {Page Name}

**Screen**

[wireframe 1]
[wireframe 2]

- [ ] .

**Page Behaviours**

- [ ] .
- [ ] .
- [ ] .

**Validations**

- [ ] .
- [ ] .
- [ ] .

**Actions**

- [ ] .
- [ ] .
- [ ] .

### Not included:
- .
- .

## Implementation Details
- **DB:**
  - .
  - .
- **Middleware:**
  - .
  - .
- **Frontend:**
  - .
  - .
- **Tests:**
  - .
  - .
- **Documentation:**
  - .
  - .

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1466?planning-poker) with Zenhub @AntBush @marcellmueller @VB-at-Bis

MarsRomer:
I don't know if it's the same redirect issue we are having with logging in internally to the analyst portal but I can't seem to be able to get to the analyst dashboard after logging in with the Business BCeID.  After logging in, I get redirectd to the Applicant dashboard and I can't figure out how to get to get to the analyst portal

marcellmueller:
Did you add an analyst role to your BCeID in the SSO dashboard?

MarsRomer:
hah, that was it!  Thank you! It works now!

marcellmueller:
Haha nice, at least you got to confirm that people with regular BCeID with no role added can’t login to the analyst side 🙂

PR #1488, opened 2023-04-20
1466 ised access

marcellmueller:
Implements #1466

To test this you will need a dev BCeID business account (ask Raphael) then add an `analyst` role in [SSO](https://bcgov.github.io/sso-requests). Once the `analyst` role has been added you will be taken to the analyst dashboard using the BCeID business account.

Once you log in you should be able to see that `external_analyst` has been set to true in `ccbc_user` table which we use in the history table to display the analyst name since previously we were just using the session sub including `idir` for `isAnalyst`.


db/deploy/mutations/create_user_from_session.sql
```diff
@@ -8,15 +8,22 @@ as $function$
 declare
   jwt ccbc_public.keycloak_jwt;
   result ccbc_public.ccbc_user;
+  is_external_analyst boolean;
 begin
   select * from ccbc_public.session() into jwt;
 
+  select true into is_external_analyst
```
VB-at-Bis, line 15: Perfect!

VB-at-Bis (approved):
Looks good to me

Issue #1486, opened 2023-04-19
Fix Revert Scripts


MarsRomer:
Business value:  Debugging errors with deploying the database will be a lot quicker, saving time and resources.  Dev are happy
Business Risk if not implemented:  Dev sad.  Slower to find breaking changes and be able to diagnose deployment issues.  Longer downtime for application which may impact business and clients.

Description:
Revert will help the dev team debug issues with the database.  This will allow the dev team to revert the database to a state where there were no errors.

AC:
- [x]  Able to correctly revert and re-deploy dev and test database without errors

MarsRomer:
Business Value:  
Helps reduce time to debug issues.

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1486?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

PR #1494, opened 2023-04-24
1486 fix revert scripts

VB-at-Bis:
Implements #1486

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -26,7 +26,7 @@ jobs:
       RENOVATE_PRIVATE_KEY: ${{ secrets.RENOVATE_PRIVATE_KEY }}
 
   deploy:
-    # if: github.event.ref == 'refs/heads/main'
+    if: github.event.ref == 'refs/heads/main'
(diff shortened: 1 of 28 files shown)


helm/app/values-dev.yaml
```diff
@@ -9,6 +9,8 @@ db:
     EOF
 
   postMigrationCommand: |
+    sqitch revert;
+    sqitch deploy;
```
marcellmueller, line 13: I don't think we need this, we will test revert in CI

.github/workflows/main.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       RENOVATE_PRIVATE_KEY: ${{ secrets.RENOVATE_PRIVATE_KEY }}
 
   deploy:
-    if: github.event.ref == 'refs/heads/main'
+    # if: github.event.ref == 'refs/heads/main'
```
marcellmueller, line 29: This needs to be undone :)

marcellmueller (approved):
Perfect!

PR #1497, opened 2023-04-25
1371 show gis data

marcellmueller:
- feat: add application gis data component
- fix: gis-schema and gis-data naming error
- feat: add gis last updated section
- feat: add gis map icon
- feat: update application gis data breakpoints
- feat: add gis input onchange
- feat: format numbers with decimals only if they are floats
- feat: add tooltip component
- fix: tooltip overflow issue
- refactor: motify application_gis_data computed column to return row
- chore: update graphql schema
- feat: use computed_column update to show last updated time
- feat: add table application_gis_assessment_hh
- feat: add save_gis_assessment_hh postgres mutation
- feat: add application_gis_assessment_hh computed column
- chore: update graphql schema
- feat: add save gis assessment hh graphql mutation
- fix: change eligible and eligible_indigenous column types to float
- feat: save eligible and eligibleIndigenous on change
- fix: null values showing as 0 for gis assessment hh inputs

Implements # 1371


app/components/LoadingCheck.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+interface Props {
+  checked: boolean;
+}
+
+const LoadingCheck: React.FC<Props> = ({ checked }) => {
+  return (
+    <StyledLoading>
+      <input type="checkbox" id="check" checked={checked} />
+      <label htmlFor="check">
```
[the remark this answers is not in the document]
marcellmueller: This has an ignore at the top of the file. Basically it's misleading as it does have an associated control, though this rule wants the input as a child of the label which would make this not work correctly.

db/deploy/computed_columns/application_gis_data.sql
```diff
@@ -2,13 +2,14 @@
 
 begin;
 
-create or replace function ccbc_public.application_gis_data(application ccbc_public.application) 
-returns ccbc_public.gis_data_item as
+drop function ccbc_public.application_gis_data(ccbc_public.application);
+
+create or replace function ccbc_public.application_gis_data(application ccbc_public.application)
```
VB-at-Bis, line 7: Just wondering if we should rename it to ‘application_gis_latest_data’ to avoid conflicts with the table with same name?

marcellmueller: Is there conflict? `gis_data` is separate from application, this is the `gis_data` for the application.

VB-at-Bis (approved):
Looks perfect

Issue #1496, opened 2023-04-25
1416 - GIS Analysis History DB only


VB-at-Bis:
As a  developer working on ticket 1416
I want to get data with total counts and list of CCBC numbers where GIS analysis data was changed, unchanged or provided in JSON file first time.
So that I can present the data as required in mockups in #1416

Description:
This card is about providing the data to be displayed to the user as described in #1416 

Acceptance Criteria:
[x] Provide DB function and corresponding GraphQL statement that returns # of records (and list of CCBC numbers) that were added for the first time,   # of records (and list of CCBC numbers) that were updated (new GIS analysis numbers) and   # of records (and list of CCBC numbers) that were unchanged

PR #1498, opened 2023-04-25
feat: GIS History DB only

VB-at-Bis:
Implements #1496 

- added type `gis_data_count`;
- updated `parse_gis_data`;
- added function `gis_data_counts(batch_id int)`;
- added and tested revert scripts;
- added tests

### DIFF
--- a/app/schema/schema.graphql
+++ b/app/schema/schema.graphql
@@ -34146,6 +34146,12 @@ type Mutation {
     """
     input: CreateUserFromSessionInput!
   ): CreateUserFromSessionPayload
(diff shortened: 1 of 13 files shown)


marcellmueller (approved):
Nice work, took me a minute to figure out how to use it properly thanks :)

VB-at-Bis:
After discussion with Marcel it was decided to change logic a bit and do not require `parse_gis_data` to be executed before `gis_data_counts` (otherwise if user choose do not continue after reviewing the data we need to delete or archive correspondent records in `application_gis_data` table)

Issue #1371, opened 2023-03-21
Show GIS analysis data


MarsRomer:
As an analyst
I want to be able to see the GIS analysis data
So that when I am doing a tech assessment I have the facts I need

# Acceptance Criteria:

## GIS assessment page

**Screen**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/0de483a2-f99d-41c5-b979-1f678b8ac79e)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=10966-96261&t=8WfWl8r66JdO9RT4-4)

- [x] Displays data that was input during JSON input in the GIS analysis row - 2 decimal places
- [x] The line under the column headers has breaks in it, include these breaks
- [x] Displays data from the application in the In application row - 2 decimal places
- [x] GIS analysis last updated date & time are set to the Created/Updated at of JSON upload - Pacific Time - ONLY IF the numbers for the GIS analysis were updated for this application
- [x] Input box only allows numbers including 2 decimal places
- [x] Input box auto-saves same time after keystroke as on applicant side (250ms debounce?)
- [x] auto-save notification appears to the right of the input box. [Saving spinner change to check mark](https://www.codingnepalweb.com/circle-loader-with-check-mark-animation/)
![Screen Shot 2023-04-10 at 11.53.28 AM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/6803138a-a3e0-4db2-9500-667a315cdd97)
![Screen Shot 2023-04-10 at 12.00.35 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/feb6ad3f-b23a-4287-84d8-d89fca8823a9)
- [x] When "Applicant is contesting the area map" checkbox on eligibility screening page is checked for this application, then the map warning icon is visible beside In application HH count
![Screen Shot 2023-04-10 at 11.57.03 AM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/96067b1d-7580-4264-9f7e-ec99035d939b)
The icon is actually two icons which are [stacked](https://fontawesome.com/docs/web/style/stack): Bottom is [map](https://fontawesome.com/icons/map?f=classic&s=solid) size `fa-lg` colour #C38A00 and top is [exclamation](https://fontawesome.com/icons/exclamation?f=classic&s=solid) size `fa-sm`
[...]

PR #1511, opened 2023-05-02
fix/gis

marcellmueller:
Adds a couple of fixes for the application gis table that Elliott mentioned [here](https://github.com/bcgov/CONN-CCBC-portal/issues/1371).

- Break between both sides of the Overbuild header (had to do a weird solution due to limited table styling
- Map icon should only appear when the checkbox is checked on the Screening page. 
- Fixed collapsing table cells when they were empty


VB-at-Bis (approved):
Looks good

PR #1519, opened 2023-05-03
fix: separate conditional for gis contesting map and household data

marcellmueller:
Implements # \<issue number\>


rafasdc (approved):
Looks good!

Issue #1416, opened 2023-04-03
GIS Analysis History


elliottSlee:
As an analyst, admin, or data team member
I want to see which GIS analysis outputs changed in the history of a project
So that I can understand the story of it and make informed judgments during negotiations and announcements

**Description:**
This card is about showing the history of GIS analysis data that's input with a JSON in #1142

# Acceptance Criteria:
- [x] On GIS analysis import success page, show # new, # updated, and # that will be ignored because they're the same
- [x] On GIS analysis import success page show the CCBC ID's under each of the categories "#New", "#updated", "#unchanged".
- [x] By default, toggle is closed toggle arrow is https://fontawesome.com/icons/chevron-right?f=classic&s=solid
- [x] Click on toggle arrow and it turns (e.g., css `transform: rotate(90deg);`) and div containing all CCBC IDs slides down
- [x] On History page, if it's the first time a GIS analysis is uploaded for the project, show just the new values, stating `<User> uploaded the first **GIS analysis** on <mmm dd, yyyy at hhmm am/pm>`
- [x] On History page, if. the GIS analysis is updated for the project, show both the new & old values, stating `<User> updated **GIS analysis** on <mmm dd, yyyy at hhmm am/pm>` -  Reason for change will be in another ticket that follows this one

## GIS Import Success

**Screen**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/ebf05ece-ad61-43a3-8e45-5f884d55785a)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?node-id=11144-100458&t=gNmtoHECWRA0rRbj-4)

## History page

**First time**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/4dece34c-8da9-4b18-9e28-5a049f311cc5)
**Updates**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/b74901e2-786a-4b0e-9e87-df0977f516a5)

**Page Behaviours**

- [ ] .
- [ ] .
- [ ] .

**Validations**

- [ ] .
- [ ] .
- [ ] .

**Actions**

- [ ] .
- [ ] .
- [ ] .

### Not included:
- [ ] Reason for change - will be part of a new ticket

## Implementation Details
- **DB:**
  - .
  - .
- **Middleware:**
  - .
  - .
- **Frontend:**
  - .
  - .
[...]

PR #1542, opened 2023-05-09
feat: gis analysis history

rafasdc:
Implements #1416


app/components/DiffTable.tsx
```diff
@@ -20,23 +20,26 @@ const StyledTable = styled.table`
   }
 `;
 
-const format = (value) => {
-  if (typeof value === 'number') {
+const format = (value, type) => {
+  if (typeof value === 'number' && type === 'number') {
     return formatMoney(value);
   }
   if (typeof value === 'boolean') {
```
AntBush, line 27: I know it's late in the game concerning this component, but this file is quite large and has many conditions and doesn't have any comments. Obviously not needed for this ticket but something we should do in the future.

AntBush (approved):
Looks good! Small comment about the DiffTable but not related to this feature

app/components/DiffTable.tsx
rafasdc: Agreed, I'll create a small ticket for it 👍

Issue #1395, opened 2023-03-27
1131 - Record Announcements - Read and Delete Announcement


AntBush:
DB Stuff:

- [x] Create a delete mutation that takes in an announcement and CCBC ID and archives that ccbc_public.application_announcement
- [x] Create a computed column for the ccbc_public.application that will display all non-archived ccbc_public.announcement that are relevant to it. 
  - [x] May not even need to with the filter by null ticket and the many to many plugin we're already using - requires some tests and checks

## Delete Announcement

**Screen**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/baa9cc86-2d73-4d73-9a17-c308aa24dda9)
- [x] Clicking red X beside an announcement opens the modal
- [x] In the modal, clicking "Delete from all projects" removes the announcement from all projects that are linked including the current one
- [x] In the modal, clicking "Remove from this project" will remove the announcement from the current project but it will remain for all other projects linked to it.
- [x] In the modal, clicking "Cancel" or the X closes the modal with no changes made.

## Not included
~- [ ] clicking on 'and more' shows all the CCBC IDs in some way (TBD)~

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1395?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

PR #1543, opened 2023-05-10
feat: delete announcement

VB-at-Bis:
Implements #1395


marcellmueller (commented):
This is working well, except when I click 'Remove from this project' it removes the announcement from all projects.

PR #1553, opened 2023-05-11
feat: do not include the ccbc id of your own application

AntBush:
I noticed a quirk in the announcements form. Right now, if you've added a project to an announcement (so in the dev data, if you're on CCBC-010001 and you've added CCBC-010003) you'll see the added project in its own project. 

I've changed it so that it always adds its own ccbcNumber as the value, and additionally removes the reference to itself in the success toast and in the autocomplete.


VB-at-Bis (approved):
Looks safe

PR #1558, opened 2023-05-15
fix: toast renders on each save

AntBush:
Implements # 1529

### DIFF
--- a/app/components/Analyst/Project/Announcements/AnnouncementsForm.tsx
+++ b/app/components/Analyst/Project/Announcements/AnnouncementsForm.tsx
@@ -195,9 +195,11 @@ const AnnouncementsForm = ({ query }) => {
   const removeSelfReference = (ccbcList: Array<any>) => {
     return ccbcList.filter((ccbcId) => ccbcId.ccbcNumber !== ccbcNumber);
   };
+  const handleUpdateToastKey = () => {
+    setToastKey((key) => key + 1);
+  };
 
   const handleSubmit = () => {
-    setToastKey((key) => key + 1);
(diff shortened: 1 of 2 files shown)


app/components/Analyst/Project/Announcements/AnnouncementsForm.tsx
```diff
@@ -194,6 +195,9 @@ const AnnouncementsForm = ({ query }) => {
   const removeSelfReference = (ccbcList: Array<any>) => {
     return ccbcList.filter((ccbcId) => ccbcId.ccbcNumber !== ccbcNumber);
   };
+  const handleUpdateToastKey = () => {
+    setToastKey((key) => key + 1);
+  };
 
   const handleSubmit = () => {
     hiddenSubmitRef.current.click();
```
marcellmueller, line 203: Nice work on this toast, it looks great.

I think this solution is slightly overcomplicated. I think we can simply set `updatedCcbcItems` to null at the top of the `handleSubmit`?
```suggestion
  const handleSubmit = () => {
    setUpdatedCcbcItems(null);
    hiddenSubmitRef.current.click();
```

AntBush: Worked perfectly! Much better than my convoluted method

AntBush:
> I am getting an error on announcements that don't have any additional projects (ccbc numbers) added to them:
> 
> Screenshot 2023-05-15 at 8 17 17 AM

Fixed the error

marcellmueller (approved):
🚢

Issue #1504, opened 2023-04-27
show application history row only when something changed


rafasdc:
As an analyst
I want the application form to not allow me to click save when I don't make a change
So that I don' t see unnecessary  history changes

**Description:**
An analyst can edit a section of the application, click save without modifying anything and a new history row will be created and shown under application history. 

Intention is to not show this history at all on the history page; alternatively an empty save could be disallowed. 

The analyst can also save without a change reason. 

# Acceptance Criteria:
- [x] Save button is actually unclickable and text reads 'Saved' until user makes an edit

# Not included:
Fixing past 'Save' clicks that didn't change any data

**Screen**

[wireframe 1]
[wireframe 2]

- [ ] .

**Page Behaviours**

- [ ] .
- [ ] .
- [ ] .

**Validations**

- [ ] .
- [ ] .
- [ ] .

**Actions**

- [ ] .
- [ ] .
- [ ] .

### Not included:
- .
- .

## Implementation Details
- **DB:**
  - .
  - .
- **Middleware:**
  - .
  - .
- **Frontend:**
  - .
  - .
- **Tests:**
  - .
  - .
- **Documentation:**
  - .
  - .

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1504?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

PR #1559, opened 2023-05-15
1504 application history row

marcellmueller:
- feat: only allow save when edit application data has changed
- test: add test for edit application save button

Implements #1504


AntBush (approved):
Works well! :shipit:

PR #1570, opened 2023-05-16
fix: ccbc id widget invisible colour

marcellmueller:
I'm unsure what the purpose of that code I removed here was...and I wrote it 😅 Though it seems to be working great now.


AntBush (approved):
Easy fix!

PR #1571, opened 2023-05-16
1562 announcement validation

marcellmueller:
- fix: datepicker and select widget error clearing
- fix: url widget error clearing

Implements # \<issue number\>


AntBush (approved):
🚀

PR #1572, opened 2023-05-16
feat: change ui element titles and allow multiple

AntBush:
Implements # 1370

### DIFF
--- a/app/lib/theme/components/FileComponent.tsx
+++ b/app/lib/theme/components/FileComponent.tsx
@@ -12,6 +12,7 @@ const StyledContainer = styled('div')`
   border: 1px solid rgba(0, 0, 0, 0.16);
   border-radius: 4px;
   padding: 16px;
+  flex-wrap: wrap-reverse;
 `;
 
 const StyledDetails = styled('div')`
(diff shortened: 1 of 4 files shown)


marcellmueller (commented):
I noticed that the upload button jumps to the other side when you upload with this change. What do you think about enabling this with a `ui:option` flag then setting `flex-direction: column-reverse` maybe with a bit of a padding under the button?

AntBush:
> I noticed that the upload button jumps to the other side when you upload with this change. What do you think about enabling this with a `ui:option` flag then setting `flex-direction: column-reverse` maybe with a bit of a margin under the button?

Yeah I noticed in the happo screens. I added the `ui:option` flag, but I didn't put a margin under the button, I'll go ahead and do that!

app/lib/theme/components/FileComponent.tsx
```diff
@@ -184,6 +194,7 @@ const FileComponent: React.FC<FileComponentProps> = ({
       </StyledDetails>
       <div>
         <StyledButton
+          addBotttomMargin={wrap}
```
marcellmueller, line 197:
```suggestion
          addBottomMargin={wrap}
```

marcellmueller (approved):
Thanks, nice work! I noticed 1 minor issue that caused the margin to not work though other than that this is great!

app/lib/theme/components/FileComponent.tsx
AntBush: <s>The extra t is for Tony</s> Fixed!

PR #1577, opened 2023-05-17
feat: workflow to create release PR on merge to main

AntBush:
Implements a job to recreate the release PR whenever a merge to main is done

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -56,6 +56,16 @@ jobs:
           git branch -D chore/release || true
           git push origin --delete chore/release || true
           git checkout -b chore/release
+      - name: Close Previous PRs
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const { owner, repo } = context.repo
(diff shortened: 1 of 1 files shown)


marcellmueller (commented):
The code looks fine but can you please explain the idea behind it. Are we now creating a release every single time we merge a PR to main? What about when the release branch gets merged to main?

Sorry I'm sure it's fine I think I am just missing something.

AntBush:
> The code looks fine but can you please explain the idea behind it. Are we now creating a release every single time we merge a PR to main? What about when the release branch gets merged to main?
> 
> Sorry I'm sure it's fine I think I am just missing something.

Yeah! No worries. 

The code does create a release PR every time we merge a PR to main. When a PR is merged in, the code deletes the old chore/release branch and recreates a new chore/release branch/pr to keep a release PR ready. So whenever we want to create a release, it's ready for us rather than one of the devs needing to initiate the process. Just a way to eliminate toil was the goal. 

The release branch would also recreate a PR for a new release

marcellmueller (approved):
Sweet, thanks for the explanation. I kind of wish it wouldn't create one when a new release got merged but it's not the end of the world.

PR #1580, opened 2023-05-17
feat: change link color to white

AntBush:
Implements # 1529 White text on sucess toast links


marcellmueller (approved):
Nice work. One thing I noticed (unrelated to this ticket) though is the tooltip cuts off at the bottom of the screen. Maybe we should move the success toast up a bit?

AntBush:
> Nice work. One thing I noticed (unrelated to this ticket) though is the tooltip cuts off at the bottom of the screen. Maybe we should move the success toast up a bit?

Good point, yeah we can move it up some pixels. Will just use the large spacing instead

PR #1587, opened 2023-05-17
1491 crunchy postgres helm

marcellmueller:
- refactor: replace postgresCluster chart with crunchy-postgres
- feat: update crunchy-postgres name override
- fix: add crunchy postgres name override value
- fix: helm subchart name issue
- test: deployment

Implements # \<issue number\>


helm/app/values-prod.yaml
```diff
@@ -11,23 +11,18 @@ resources:
...
-  requests:
-    cpu: 250m
-    memory: 16Mi
-  limits:
-    cpu: 250m
-    memory: 32Mi
+proxy:
+  pgBouncer:
+    requests:
+      cpu: 250m
```
rafasdc, line 17: suggest cpu request be 100m or less

marcellmueller: Sorry I missed this message. Did you think it should be `1m` like we have for the rest in `values.yaml`? Also the app requests just above this are the same as the limits as well at `500m`, should I update that?

helm/app/values.yaml
```diff
@@ -159,3 +102,85 @@ loadTest:
...
+  postgresVersion: 14
+  instances:
+    name: hippo-ha # high availability
+    replicas: 3
+    dataVolumeClaimSpec:
+      storage: 2Gi
+      storageClassName: netapp-block-standard
+    requests:
+      cpu: 1m
+      memory: 1Gi
```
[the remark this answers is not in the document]
marcellmueller: Fixed.

helm/app/values-prod.yaml
rafasdc: Should be fine at `1m - 10m` and yes, we should update that one as well if you haven't already. The intention is that the pod is "burstable" and we are being good community members by just requesting ~average the usage and setting the limit to allow it to burst during "high" traffic or expensive transactions.

marcellmueller: Sounds good. Yeah that is what we have been doing though I can't remember why, Alec and I changed the prod values to increase the requests hoping it would solve some issue though I will revert it.

rafasdc (approved):
I think we are good to merge this in, it looks it will solve the PDB issue, we'll likely just have to remove the extra ones.

marcellmueller:
> I think we are good to merge this in, it looks it will solve the PDB issue, we'll likely just have to remove the extra ones.

Sounds good thanks!

Issue #1549, opened 2023-05-11
SOW - Auto Import Tab 2


MarsRomer:
As an analyst
I want the SOW data to be auto populated in the database
So that I can do a better job of administering and announcing the project

**Description:**
This ticket is for the auto import of just TAB 2 of the SOW worksheet
[SOW template with yellow highlighted fields for import](https://bcgov.sharepoint.com/:x:/r/teams/04708-ScrumTeam/Shared%20Documents/Scrum%20Team/UBF-AA-00000-SOW_tbl_New_Jan12_2023%20-%20Justin%202023-04-26%20(1).xlsx?d=w45fcb100f15a489c866d10d89400ef6a&csf=1&web=1&e=9f0Grs)

[Sample Data](https://bcgov.sharepoint.com/:x:/r/teams/04708-ScrumTeam/Shared%20Documents/Scrum%20Team/CCBC-020118%20-%20Statement%20of%20Work%20Tables%20-%2020230517.xlsx?d=w3793701bcabb40669f342c7c049634ce&csf=1&web=1&e=LYO8A7)

[POC](https://github.com/bcgov/CONN-CCBC-portal/compare/main...1369-import-sow-data-from-excel)

# Acceptance Criteria:
- [x]  Yellow highlighted data in SOW xls Tab 2 is auto populated in database when imported using the portal (separate ticket for UI component)
- [x]  SOW Tab 2 data is available in parsed format in a metabase query in separate columns so that it's easy to use.

**Validations**

- [x] Milestones dates are stored as dates
- [ ] .
- [ ] .

### Not included:
- .
- .

## Implementation Details
- **DB:**
  - .
  - .
- **Middleware:**
  - .
  - .
- **Frontend:**
  - .
  - .
- **Tests:**
  - .
  - .
- **Documentation:**
  - .
  - .

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1549?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

MarsRomer:
I am not sure if I am seeing this data in metabase
![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/007d1b38-ee7d-404b-b949-fb14c8018203)

AntBush:
[...]

Issue #1533, opened 2023-05-09
SOW - Auto-import SUMMARY tab


MarsRomer:
As an analyst
I want the SOW data to be auto populated in the database
So that I can do a better job of administering and announcing the project

**Description:**
This ticket is to import the yellow highlighted data in SUMMARY tab of the SOW xls
[SOW template with yellow highlighted fields for import](https://bcgov.sharepoint.com/:x:/r/teams/04708-ScrumTeam/Shared%20Documents/Scrum%20Team/UBF-AA-00000-SOW_tbl_New_Jan12_2023%20-%20Justin%202023-04-26%20(1).xlsx?d=w45fcb100f15a489c866d10d89400ef6a&csf=1&web=1&e=9f0Grs)

[Sample Data](https://bcgov.sharepoint.com/:x:/r/teams/04708-ScrumTeam/Shared%20Documents/Scrum%20Team/CCBC-020118%20-%20Statement%20of%20Work%20Tables%20-%2020230517.xlsx?d=w3793701bcabb40669f342c7c049634ce&csf=1&web=1&e=LYO8A7)

[POC](https://github.com/bcgov/CONN-CCBC-portal/compare/main...1369-import-sow-data-from-excel)

# Acceptance Criteria:
- [x] Yellow highlighted data in SOW xls SUMMARY tab is auto populated in database when imported using the portal (separate ticket for UI component)
- [x] SOW SUMMARY data is available in parsed format in a Metabase query in separate columns so that it's easy to use.
- [x] The following field names are the same in the database project table as they are in the application form table (`form_data`)

| Excel file             | form_data          |
|------------------------|--------------------|
| Applicant name         | organizationName   |
| Project title          | projectTitle       |
| Application number     | ccbc_number        |
| Backbone technologies  | backboneTechnology |
| Last mile technologies | lastMileTechnology |

**Validations**

- [x] CCBC ID format matches ours

### Not included:
- Check that CCBC ID matches the application it's being uploaded to

## Implementation Details
- **DB:**
  - .
  - .
- **Middleware:**
  - .
  - .
- **Frontend:**
  - .
  - .
- **Tests:**
  - .
  - .
- **Documentation:**
  - .
  - .

MarsRomer:
[...]

PR #1604, opened 2023-05-19
feat: import sow

VB-at-Bis:
Implements #1549


app/package.json
```diff
@@ -90,7 +90,8 @@
     "relay-runtime": "^13.2.0",
     "styled-components": "^5.3.5",
     "typography": "^0.16.21",
-    "validator": "^13.9.0"
+    "validator": "^13.9.0",
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.19.3/xlsx-0.19.3.tgz"
```
AntBush, line 94: Probably better to just use the npm registry rather than the CDN here, atl east for consistency's sake

app/backend/lib/sow-upload.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const filename = Object.keys(files)[0];
+  const uploaded = files[filename]; 
+
+  const buf = fs.readFileSync(uploaded.filepath);
+  const wb = XLSX.read(buf); 
+
+  // check if we have all needed worksheets
+  let missingSheet = '';
+  sheetNames.forEach(x=>{
```
AntBush, line 48: Probably should have a better name for the variable here

Maybe something like: 
```suggestion
  sheetNames.forEach(sheet =>{
```

app/backend/lib/sow-upload.ts
```diff
@@ -0,0 +1,89 @@
...
+      if(wb.SheetNames.indexOf(x) === -1) {
+        missingSheet = x;
```
AntBush, line 50: Not a huge issue, but if there multiple missing sheets we will only get one here

app/backend/lib/sow_import/tab_2.ts
```diff
@@ -0,0 +1,78 @@
...
+      }  
+      if (tableDetected) { 
+          validLine = sheet[row]['O'] && typeof(sheet[row]['O']) === 'string'; 
+          value = sheet[row]['O'];
+          // now reading line
+          if (validLine && (value === 'Complete')) {
+            const lineData = 
+            {  
+              entryNumber: sheet[row]['A'], 
+              projectSiteName: sheet[row]['B'],
```
AntBush, line 40: Something to check with @elliottSlee, part of the acceptance criteria is that any data that refers to the same json key in the application form data should be the same

AntBush (commented):
Only one small change with the package.json I'd say is absolutely necessary. Also running prettier would be appreciated

app/backend/lib/sow_import/tab_2.ts
VB-at-Bis: yes, #1533 mentions other fields, nothing about projectSiteName

AntBush: Sorry projectSiteName was just a random one I decided to use for commenting on this part here. I'm also not sure here either if any of these are duplicated in the application data though. Probably not but wouldn't be a bad idea checking in with Elliot

app/package.json
VB-at-Bis: I had to use cdn - see https://docs.sheetjs.com/docs/getting-started/installation/nodejs/

> Older releases are technically available on the public npm registry as xlsx, but the registry is out of date. The latest version on that registry is 0.18.5
> This is a known registry bug

app/backend/lib/sow-upload.ts
VB-at-Bis: I guess we'll handle all validations in some other ticket - we discussed that we will have one pass to validate file and return list of all errors, and second pass to actually import in DB. This ticket is all about import.

app/backend/lib/sow-upload.ts
VB-at-Bis: Updated as suggested

app/package.json
AntBush: Ah sorry didn't know this, thanks for letting me know!

AntBush (approved):
Good stuff!

PR #1620, opened 2023-05-25
fix: handle links without protocol

rafasdc:
Bug fix for announcements link previews.

Implements # \<issue number\>


VB-at-Bis (approved):
Looks safe

PR #1623, opened 2023-05-25
1544 sow data input ui

marcellmueller:
- feat: add project_information_data table
- feat: add create project information db mutation
- feat: add computed column application_project_information

Implements # \<issue number\>


rafasdc:
Looks pretty good, just noticed the following:
- When you go cancel the files disappear from the "read only" if you refresh the page they reappear. 
- When you go cancel the files you had previously uploaded disappear from the "read only" view, if you click edit one more time and then "save & import" everything gets reset.

marcellmueller:
> Looks pretty good, just noticed the following:
> 
> * When you go cancel the files disappear from the "read only" if you refresh the page they reappear.
> * When you go cancel the files and info you had previously entered disappear from the "read only" view, if you click edit one more time and then "save & import" everything gets reset/overwritten.

Nice catch thanks, it should be good now. I was completely resetting the form data on cancel instead of replacing with the old data if it existed, as well as had to update the relay store on save.

rafasdc (approved):

PR #1627, opened 2023-05-29
feat: import tab 1

AntBush:
Implements # 1548


app/backend/lib/sow_import/tab_1.ts
```diff
@@ -1,34 +1,146 @@
...
-   
+const LoadTab1Data = async (sow_id, wb, sheet_name, req) => {
+  const data = readData(sow_id, wb, sheet_name);
+  const input = { input: { sowId: sow_id, jsonData: data } };
   // time to persist in DB
-  const result = await performQuery(createSomeMutation, {input: data}, req)
-  .catch((e) => {
-    return { error: e };
-  });
+  const result = await performQuery(createSomeMutation, input, req).catch(
```
VB-at-Bis, line 137: should it be `tab1Mutation` ?

app/tests/backend/lib/sow_tab_1.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (BOOLEAN_KEYS.includes(key)) {
+      rowData[key] = 'Yes';
+    } else {
+      rowData[key] = `Value_${index}`;
+    }
+  });
+  return rowData;
+};
+
+describe('sow_tab_1 parsing tests', () => {
```
VB-at-Bis, line 33: maybe we need to check if graphql mutation was called with expected input?
i.e. check https://github.com/bcgov/CONN-CCBC-portal/pull/1604/files#diff-97d71ed34fde88b1575c5db66b79a727bcb0edb0c96b6ff57a4594ad53894774R33

VB-at-Bis (commented):
Please see comments inline

app/tests/backend/lib/sow_tab_1.test.ts
AntBush: Added!

app/backend/lib/sow_import/tab_1.ts
AntBush: Good call, changed it

PR #1631, opened 2023-05-29
1551-Auto-import SoW tab 8

VB-at-Bis:
Implements #1551


marcellmueller (approved):
Nice work! This is working well.

Issue #1632, opened 2023-05-30
SoW upload - Validation - Backend 


VB-at-Bis:
sow-upload.ts should return validation errors to be used by Front End to display to the user as described in [#1617](https://app.zenhub.com/workspaces/ccbc---delivery-board-61c0c1204c8bcb001491c5f3/issues/gh/bcgov/conn-ccbc-portal/1617)

MarsRomer:
Validations will be PO reviewed with #1617 since this is a back end ticket

PR #1646, opened 2023-05-31
feat: validate-only mode for sow-upload

VB-at-Bis:
Implements #1632
- sow-import now accepts ?validate=true parameter and performs file parsing/validation without calls to DB


AntBush (approved):
Looks good!

PR #1647, opened 2023-05-31
fix: receive application attachment spread application bug

marcellmueller:
Implements # \<issue number\>


AntBush (approved):
Looks good!

PR #1657, opened 2023-06-02
feat/project information feature flag

marcellmueller:
Add a feature flag for project information. 

Also snuck in a small improvement to the file widget which allows a file of the same name to be reuploaded if the upload fails. This has always been a thing with our file widget, or anything using html file input but especially noticeable here if someone was say uploading a SoW excel with the wrong CCBC number.


VB-at-Bis (approved):
Looks safe

PR #1661, opened 2023-06-02
fix: use parseint on application number in summary tab import

marcellmueller:
Implements # \<issue number\>


VB-at-Bis (approved):
Looks safe

PR #1662, opened 2023-06-02
chore: release v1.73.1

marcellmueller:
Had to get in a parseInt fix for application id in the sow api to allow it to work, thought it was coming in another PR.


VB-at-Bis (approved):
Looks good

Issue #1653, opened 2023-06-01
Verify form fields in applicant form on save


marcellmueller:
Due to a bug on the applicant form where form fields can be saved in the incorrect form page we need to verify they belong in that section against the form schema before saving.

- [x] Verify all fields belong to that form page on save in `ApplicationForm` component and only save correct fields
- [x] Capture Sentry log of incorrect fields?

MarsRomer:
Nothing for PO to review here, moving the Sprint Done

PR #1666, opened 2023-06-05
1653 verifiy form fields on save

marcellmueller:
implements #1653 

Takes a list of the incoming form data field keys and compares it against a list of that form pages keys then filters out fields that don't belong. 

I had to redo how I was getting the schema keys so it would support complex form schemas with `oneOf` which includes arrays of objects. Let me know if you have any questions or need help testing this.


VB-at-Bis (approved):
Looks perfect

PR #1676, opened 2023-06-07
chore: update openid client to node 18 compatible version and nodejs and node types to v18.16.0

marcellmueller:
Also had to update `openid-client` since the version we were using only supported up to node 16.


VB-at-Bis (approved):
Looks safe

PR #1677, opened 2023-06-07
chore/nextjs 13

marcellmueller:
Branched off nodejs update PR #1676 
 

- had to remove all child `<a>` tags from the `Link` component as that is depreciated now. 
- the updated `next/image` component strangely added a margin to the bottom of it so I had to fix that in the three places we use it.
- tests needed to be updated due to the `next/link` change and `useRouter` mocks had to be updated.
- the main logo in the nav is causing happo diffs for some reason. I have compared it and can't find any meaningful differences.

Maybe have a look through the happo diffs and see if you can spot something I haven't, as well just click through the page and make sure everything is working well thanks!


rafasdc (approved):
Amazing, runs perfect.

Issue #1630, opened 2023-05-29
Update internal & external status


elliottSlee:
As an analyst 
I want to change the external status
So that the applicant can see that their project is approved on their dashboard

**Description:**
_if needed_

# Acceptance Criteria:
- [x] Add dropdown to external status. Options in External status dropdown
- Received
- Conditionally Approved
- Approved
- Complete
- Cancelled
- Closed
- On Hold

- [x] **Internal** status options always enabled: All previous ones, closed, on hold
- [x] Internal status options disabled:
- If Internal status == Received, then Screening is enabled but Assessment, Recommendation, Cond Appr, Approved, Complete, and Cancelled are disabled
- If Internal status == Screening, then Received and Assessment are enabled but Recommendation, Cond Appr, Approved, Complete, and Cancelled are disabled
- If Internal status == Assessment, then Received, Screening, and Recommendation are enabled but Cond Appr, Approved, Complete, and Cancelled are disabled
- If Internal status == Recommendation, then Received, Screening, Assessment, and Cond Appr are enabled but Approved, Complete, and Cancelled are disabled
- If Internal status == Cond Appr, then Received, Screening, Assessment, Recommendation, and Approved are enabled but Complete, and Cancelled are disabled
- If Internal status == Approved, then Received, Screening, Assessment, Recommendation, Cond Appr, Complete, and Cancelled are enabled
- [x] External status. All options enabled.
- [x] External status. If user selects conditionally approved, it can only change if (1) Internal Status is Cond Appr and the conditionally approved form has (2) Minister's decision = Approved and (3) Applicant's response = Accepted. If these 3 are not met, a modal appears:
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/cfa0ad6f-fff1-4991-b8a6-a81ebeb6e226)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=11649-114341&t=0PaYkf2EIQh9lpLz-4)
- [x] "Take me there" button links to Conditional Approval form
- [ ] Click x or close, closes the modal
- [x] External status. If user selects Approved, Complete, Cancelled, Closed, or On Hold, it can only change if Internal Status is the same. If it is not the same, show the following modal:
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/9d5d8ffa-d6b5-49bc-89d5-be184e2f092b)
[...]

PR #1710, opened 2023-06-13
1630 update internal external status

marcellmueller:
Let me know if you can find any bugs thanks!!

Takes care of #1630

### DIFF
--- a/app/components/Analyst/ChangeStatus.tsx
+++ b/app/components/Analyst/ChangeStatus.tsx
@@ -118,7 +118,7 @@ const ChangeStatus: React.FC<Props> = ({
     conditionalApprovalData?.response?.applicantResponse === 'Accepted';
 
   const disabledStatuses =
-    disabledStatusList && disabledStatusList[currentStatus.name];
+    disabledStatusList && disabledStatusList[currentStatus?.name];
 
   // No dropdown for withdrawn applications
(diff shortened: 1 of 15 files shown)


app/components/Analyst/ChangeStatus.tsx
```diff
@@ -68,112 +67,193 @@ const ModalDescription = ({ currentStatus, draftStatus }) => {
...
+    setDraftStatus(getStatus(e.target.value, statusTypes));
+    const isAllowedExternalChange =
+      isExternalStatus && e.target.value === analystStatus;
+
+    if (!isExternalStatus) {
+      // open modal for internal status change
+      window.location.hash = `#change-status-modal`;
+    } else if (
+      // open modal for external status change
+      !isAllowedExternalChange ||
```
AntBush, line 195: I think it might be easier to read if we refactor this above. Something like:

```javascript
const isAllowedExternalChange = isExternalStatus && e.target.value === analystStatus
const isInvalidConditionalApproval = e.target.value === 'conditionally_approved' && !isAllowedConditionalApproval
...
if(!isAllowedExternalChange || isInvalidConditionalApproval){
  window.location...
}
```

AntBush (approved):
Looks good! Two suggestions that are more style than anything, good stuff though

PR #1711, opened 2023-06-13
feat: clear project data on no

rafasdc:
Implements # 1680

### DIFF
--- a/app/components/Analyst/Project/ProjectInformation/ProjectInformationForm.tsx
+++ b/app/components/Analyst/Project/ProjectInformation/ProjectInformationForm.tsx
@@ -190,7 +190,11 @@ const ProjectInformationForm = ({ application }) => {
       }}
       formData={formData}
       handleChange={(e) => {
-        setFormData({ ...e.formData });
+        if (!e.formData.hasFundingAgreementBeenSigned) {
+          setFormData({ hasFundingAgreementBeenSigned: false, main: {} });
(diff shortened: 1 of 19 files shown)


app/components/Analyst/Project/ProjectInformation/ProjectInformationForm.tsx
```diff
@@ -44,6 +86,87 @@ const ProjectInformationForm = ({ application }) => {
...
+        }) => {
+          const archivedAt = new Date().toISOString();
+          const applicationSowInput = {
+            id: node.id,
+            applicationSowDataPatch: { archivedAt },
+          };
+          if (node.archivedAt === null) {
+            archiveSowData({
+              variables: {
+                input: applicationSowInput,
```
AntBush, line 111: We usually create a mutation to archive on creation of another record, to keep it all as part of one transaction to prevent any network errors preventing the DB from being in an undefined state (tabs 1 and 2 are archived, but the request for 7 and 8 didn't go through for example). 

Not exactly the case here, but would be a good idea to create a "saveProjectInformationAndArchive" db mutation function to handle updates in the future as well. Right now we're using the `set_archived` trigger function on some places to automatically archive when we create a new record - can use that for the project_information and the sow_data table but the individual tabs will need some logic in the function itself

VB-at-Bis (approved):
Looks safe

PR #1721, opened 2023-06-14
1617-sow-validation-frontend

AntBush:
Implements # \<issue number\>


marcellmueller (commented):
This is working well but the design of the file widget validation does not look like Elliott's designs, the `Checking validation...` completely replaces all the file widget content instead of appearing to the right. Also I don't see it in the acceptance criteria but it looks like there is a validation complete status message that is supposed to replace the `Checking validation...` message.

Let me know if the designs changed, maybe this was how Elliott wanted it.

Also this is a total nit but <p> tags add a margin to the bottom that should be removed here:
Screenshot 2023-06-20 at 10 00 12 AM

app/components/Analyst/Project/ProjectInformation/ProjectInformationForm.tsx
```diff
@@ -10,6 +10,66 @@ import { useCreateProjectInformationMutation } from 'schema/mutations/project/cr
 import ProjectTheme from 'components/Analyst/Project/ProjectTheme';
 import MetabaseLink from 'components/Analyst/Project/ProjectInformation/MetabaseLink';
 import Toast from 'components/Toast';
+import validateFormData from '@rjsf/core/dist/cjs/validate';
+import { Alert } from '@button-inc/bcgov-theme';
+
+const StyledAlert = styled(Alert)`
+  margin-bottom: 8px;
+  margin-top: 8px;
```
marcellmueller, line 18:
```suggestion
p {
   margin-bottom: 0;
}
```

not a huge deal but there is still that weird margin inside the alert. either this or using a div or span instead of `<p>` would work.

marcellmueller (approved):
Nice work, thanks for updating that!

app/components/Analyst/Project/ProjectInformation/ProjectInformationForm.tsx
AntBush: I am using a div now, I think the margin might be from the styling of the container for the content? That's my guess

app/components/Analyst/Project/ProjectInformation/ProjectInformationForm.tsx
```diff
@@ -10,6 +10,66 @@ import { useCreateProjectInformationMutation } from 'schema/mutations/project/cr
...
+  }
+  return (
+    <StyledAlert
+      key={errorMessage}
+      variant="danger"
+      closable={false}
+      content={
+        <>
+          <div> {title}</div>
+          <p>{errorMessage}</p>
```
marcellmueller, line 67: This `<p>` 😅

AntBush: Ah! Sorry the test file I'm using was only rendering the divs. Thanks haha

PR #1737, opened 2023-06-16
refactor: clean up simple sonarcloud code smells

marcellmueller:
This takes care of the most basic code smells. I will likely clean up some more in a separate PR but I didn't want to risk introducing any bugs.

Also had to fix a large dependency cycle error mostly involving the review page components and write a few tests to make sonarcloud happy.


VB-at-Bis (approved):
Looks safe

PR #1741, opened 2023-06-20
feat(ci): add test to ensure that no sqitch tags are deleted

AntBush:
Implements # \<issue number\>


marcellmueller (approved):
Nice work! Thanks for doing a test run 🥇

Issue #1702, opened 2023-06-13
Add links to Metabase on GIS import pages


elliottSlee:
As an analyst
I want the GIS import pages to have a link to the Metabase GIS analysis dashboard
So that I can see the GIS analysis for all projects

# Acceptance Criteria:

## Input page
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/621b3755-655f-4467-86ec-3e2cd101a543)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=11639-123483&t=9bEtzcEfUrnL8iMZ-4)

- [ ] Add card for GIS Metabase link
https://ccbc-metabase.apps.silver.devops.gov.bc.ca/dashboard/87-gis-analyses

background: #FFFFFF;
border: 1px solid #D6D6D6;
box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
border-radius: 8px;

- [ ] Convert upload label & description to card

## GIS Import Success Page
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/abb49157-4ccf-4cb0-a249-7cf6cc810e52)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=11144-100458&t=9bEtzcEfUrnL8iMZ-4)
- [ ] Add card to GIS import success page
https://ccbc-metabase.apps.silver.devops.gov.bc.ca/dashboard/87-gis-analyses

## Implementation Details
- **DB:**
  - .
  - .
- **Middleware:**
  - .
  - .
- **Frontend:**
  - .
  - .
- **Tests:**
  - .
  - .
- **Documentation:**
  - .
  - .

MarsRomer:
@VB-at-Bis  can all the acceptance criteria be checked off?

PR #1746, opened 2023-06-21
feat: adding metabase link to gis page

VB-at-Bis:
Implements #1702


marcellmueller (commented):
Looks different than the ticket:
Screenshot 2023-06-23 at 7 25 12 AM

Screenshot 2023-06-23 at 7 46 04 AM

I think link has different colour blue (#1A5A96 in figma) and font is bold. Not sure why this design is a bit different than on project information page, we can ask Elliott in standup.

marcellmueller (approved):
Okay thanks for letting me know, I still think we need to update the link colour (I should have got it right when I made MetabaseLink component) but not a big deal.

PR #1757, opened 2023-06-22
fix(ci): use release-it to determine tag

AntBush:
Implements fix for the release action by specifying the tag version to be what release-it is attempting to commit. 

Also added the sqitch config for the ccbc service account's username and email so we don't get something like: 

`@1.69.0 2023-05-19T22:45:23Z ,,, <runner@fv-az482-816> # release v1.69.0`

The error it faced last time was due to the tag deleted and being recreated. Note that the tag for release 1.69.0 is not associated with any branch in the repository https://github.com/bcgov/CONN-CCBC-portal/commit/b3ee81ed99d6920814464b3748634398e8192120


marcellmueller (approved):
🕺

Issue #1751, opened 2023-06-22
Bug:  Summary data is not getting uploaded into the database from SOW data


MarsRomer:
**Describe the bug**
The summary data on the first tab of the SOW table is not getting uploaded into the database

**To Reproduce**
1. Upload SOW
2. View in metabase table 'Application SOW Data'
3. new application not in the table

**Expected behaviour**
SOW data should show up in the Application SOW Data

**Screenshots**

**Logs/tracebacks**

**Additional context**

MarsRomer:
I seems this still a bit buggy, I deleted and uploaded a SOW to application 020189 a couple of times today, the latest one doesn't show in the Application SOW table
![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/90d6cc54-c2b4-4f63-9c5d-9366962836dd)

PR #1763, opened 2023-06-23
fix: correct type for mutation input

VB-at-Bis:
Implements #1751

- fixes graphql error which prevented saving data in DB;


AntBush (approved):
Looks safe

Issue #1732, opened 2023-06-15
Uploaded Change Request SOW data is in DB


MarsRomer:
As an analyst 
I want the uploaded change request SOW data in the database
So that I can view the changes to the project in reports

**Description:**
The data from the change request should be in the database.  The SOW template will be updated and uploaded into the portal as part of #1691 

# Acceptance Criteria:
- [ ] All necessary data from change request SOW will be visible in metabase
- [ ] Original SOW at signing has an ID of 0
<s>- [ ] First change request has an ID of 1</s>
<s>- [ ] Subsequent new change request uploads count up - gapless</s>
- [ ] Replacing a SOW at signing or change request should retain it's ID and archive the data that is being replaced
- [ ] Metabase data from change request can be easily viewable in parsed format so that it can be used for reporting
- [ ] Metabase user can tell the difference between SOW at signing and SOW from change requests (true/false column)
- [ ] Each change request can be uploaded again if there was a mistake and metabase user can see the lastest version of that particular change request
- [ ] There can be multiple change requests for each project and metabase user can tell how many change requests there have been and what the latest project information is

MarsRomer:
I'm not sure if this is replicable but I wasn't able to upload a second change request just now, the spinner has been spinning for approx 10-15 minutes now.  
![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/083faace-08fa-41f3-bb72-aa472f8966a8)

MarsRomer:
I cancelled out of the upload (see previous comment) using the 'Cancel' button and tried again but still didn't upload, same spinning.  The app logged me off and I logged back in and the upload worked on first try but then the save and import is non responsive as I had encountered earlier this morning with the SOW upload.  
![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/27644666-d78c-48a7-aee5-1586b25af198)

MarsRomer:
I'm gonna put this to in progress, not sure if there is a bug or if there is something else at play here but worth exploring

marcellmueller:
Okay we can look into this. Though I don't think this ticket is the right place for this discussion as it was focused on database changes.

MarsRomer:
agreed, not sure if this is a database thing or a UI thing at this point

marcellmueller:
I'm wondering if it's an issue with our `test` namespace because I'm in there now and I am getting errors just trying to view applications.

MarsRomer:
[...]

PR #1768, opened 2023-06-26
1732 change request data db

marcellmueller:
Implements #1732


AntBush (approved):
Looks good! Thanks for adding the tags on postgraphile

PR #1783, opened 2023-06-27
feat: cosmetic changes for sow

AntBush:
Implements # 1617
1. Moved messages to appear right below sow upload field
2. "Checking the data..." is now a fixed width to prevent it from sliding
3. The moment you upload a new file, the errors disappear rather than waiting for the vlidation to occur
4. Even if errors occur, the onChange is committed to allow for the form to still be required *and* for that file to not be valid


marcellmueller (approved):
Looks good! 🚢

Issue #1664, opened 2023-06-05
GIS analysis upload display depends on status


elliottSlee:
As an analyst
I want the GIS analysis to update on the Assessment page only when the status is Received, Screening, Assessment
So that we can have the right snapshot in time for the application.

# Acceptance Criteria
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/b9092200-ab0a-4145-b8d2-5b003b82f5cf)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=11144-100458&t=9bEtzcEfUrnL8iMZ-4)

- [x] If the status of an application is Received, Screening, Assessment, when the GIS analysis JSON is uploaded, then the #s are updated on the GIS assessment page
- [x] If the status of an application/project is Recommendation, Conditionally Approved, Approved, Complete, Cancelled,  Withdrawn or anything other than those mentioned in the previous AC when the GIS analysis JSON is uploaded, then the #s are _NOT_ updated on the GIS assessment page
- [x] Uploader only imports the above data into the DB - Block upload of data that is at or beyond the Recommendation state
- [x] After importing, display which projects were not uploaded due to status rules
<img src="https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/adf7fc2f-77bc-421c-be4b-1b52160e3b89" width="50%">

- [x] Unchanged, not imported sections: colour F1F2F3, radius 8
- [x] Successful import sections: colour DFF0D8, radius 8, icon [font-awesome circle-check](https://fontawesome.com/icons/circle-check?f=classic&s=solid) 2E8540

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1664?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

PR #1797, opened 2023-06-28
1664 gis upload status

marcellmueller:
implements #1664 

Please roast my database code. 

Also fixed the metabase link colour and added a breakpoint for it on the GIS success page since I was there.


db/deploy/mutations/parse_gis_data.sql
```diff
@@ -22,20 +23,27 @@ begin
...
+        select id into application_row_id from ccbc_public.application where ccbc_number = ccbc_id;
+
+        -- Look up the application_status based on the application_id value
+        select status into application_status_name from ccbc_public.application_status
+        where application_id = application_row_id
+        order by id desc limit 1;
+
+        -- If no row was found, add to list
+        if not found or
+            (application_status_name != 'received' and application_status_name != 'screening' and application_status_name != 'assessment')
```
AntBush, line 38: nit: you can add in a `application_status_name not in ('received', 'screening', 'assessment')` I think. But more of a style choice, might make things easier when we want to add in other statuses

db/deploy/computed_columns/gis_data_counts.sql
```diff
@@ -50,12 +53,32 @@ return query
...
+    not_imported as (
+       select
+            (json_row->>'ccbc_number') as ccbc_number,
+            a.id as application_id,
+            json_row as json_data
+        from json_rows, ccbc_public.application a
+        join ccbc_public.application_status s on s.application_id = a.id
+        where a.ccbc_number = (json_row->>'ccbc_number')
+        and s.archived_at is null
+        and (s.status != 'received' and s.status != 'screening' and s.status != 'assessment')
```
AntBush, line 75: nit: can be replaced with same comment done in parse_gis_data.sql

AntBush (approved):
No roasting here! Looks good, just a couple of nits but otherwise good stuff!

db/deploy/mutations/parse_gis_data.sql
marcellmueller: Thanks, I definitely think this is simpler. Updated!

PR #1805, opened 2023-06-29
feat: add new hook to cancel operations on hook change

AntBush:
Implements # 1652

I've added a reusable hook to be able to dispose of requests made to graphql (and thus cancelling the onComplete that initiates once it ends) on a route change.


marcellmueller (approved):
This is very slick, nice work! I wasn't able to reproduce that form data bug with this.

PR #1812, opened 2023-06-29
fix: session timeout

rafasdc:
Implements # 1690

- Updated logout for inactivity to be handled by sso-express
- Separate session-idle-remaining and extend into their own endpoints
- session-idle will no longer refresh token to prevent countdown from 1700+
- Refresh the timeout when session is extended


marcellmueller (approved):
Looks good!

Issue #1829, opened 2023-06-30
Bug: SoW not saving in database


marcellmueller:
Looks like a bug was introduced with the validation, the SoW is validating but not saving in the database on `Save & Import Data`.

MarsRomer:
the Save and import button is non responsive, not sure if it's related to this ticket 
![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/0bdf0877-0836-42f6-a810-4e89d2e98d77)

PR #1830, opened 2023-06-30
fix: sow validate query param boolean bug

marcellmueller:
Implements #1829

This must have slipped by in a validation ticket. Since query params are all strings this made `validation = true` all the time so it wouldn't save in the database.


VB-at-Bis (approved):
Looks safe

rafasdc (approved):

PR #1834, opened 2023-06-30
feat: provide information about internal and external status of applications

rafasdc:
uses mui dialog/modal for accessibility

Implements # 1716


VB-at-Bis (approved):
Perfect

PR #1866, opened 2023-07-07
1673-redirect-sentry-metrics

AntBush:
Implements  sentry error handling


rafasdc (approved):
I'm really curious to see the full tokenSet that is being attempted to get refreshed. Not sure how we'll get past SonarCloud, might have to use a patch or something?

Issue #1828, opened 2023-06-30
Read only view of Change request metadata - merge project information and change request


MarsRomer:
As an analyst
I want to see the change request and project information together 
So that it's easy for me to see and get information about the project 

**Description:**

_if needed_  

# Acceptance Criteria
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/2163795d-12af-4a7d-9444-a37a2943bc64)
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/d897d202-8fcc-4b60-959d-5e8a95bc2d0d)

[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=11918-127943&mode=design&t=4oABshMK52dHeeT6-4)

- [x] Files links only show in the UI if they were uploaded
- [x] Hover over file links shows filename
- [x] If file isn't uploaded, elements still appear in the grid (see Amendment 4 in the wireframes)
- [x] Date is the date change request approved
- [x] Description area is 451x60 (if this needs to change to make it look better, go ahead). `text-overflow: ellipsis;`
- [x] Hover over truncated description shows full description
- [x] Clicking view more slides down the contents using similar animations as forms on this page and rotates chevron to point down (like on GIS upload success page)
- [x] If change request form filename is longer than box width, wrap the text _not the containers_
- [x] Click edit pencil re-opens that form

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1828?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

marcellmueller:
@elliottSlee can you please clarify acceptance criteria ` Files links only show in the UI if they were uploaded` and `If file isn't uploaded, elements still appear in the grid`, I'm sure I'm missing something but these seem at odds with each other.

marcellmueller:
@elliottSlee hey sorry to bug you again but what is supposed to happen to the metabase link once the `Add amendment` button disappears? I can try hiding it as well but I think it's going to look weird.

elliottSlee:
[...]

Issue #1770, opened 2023-06-26
Change request - Metadata


MarsRomer:
As an analyst
I want to enter in metadata for the change request 
So that I can have an understanding and be able to categorize the change requests for reporting

**Description:**
There are some pieces of information that we need to capture for the change requests so that the ministry can understand and categorize the change requests

# Acceptance Criteria:
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/89993663-e9a0-4724-b52c-77b522c3c3f1)
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/5113a64b-9c7d-431b-8711-619ce9123028)

- [x] Amendment # is 3 digit integer only. Mandatory.  (validation that it's a unique # is ticket #1777)
- [x] Description of changes: Text  Max 10,000 characters.  Mandatory 1 characters at minimum
- [x] Date requested - Date today or prior to.  validate that it's not in the future
- [x] Date approved - Date today or prior to.  Validate that it's not in the future
- [x] Level of amendment: Radio buttons - all unselected when form opens
- [x] When a SoW Excel file is uploaded, the Save button changes to "Save & Import"
- [x] Additional comments:  text max 10,000 characters

Notes:  
Radio buttons should be designed and built in a way that it enables changes that don't require a lot of effort.

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1770?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

marcellmueller:
@MarsRomer @elliottSlee what do we mean by `validate that it's not in the future` for both of the new date fields here? Do we want visual feedback and to not allow the form to be saved?

MarsRomer:
I was imagining that the date picker would simply not allow it in the sense that the dates in the future would be grayed out and not selectable.  nip it in the bud.

marcellmueller:
Yeah we can definitely set a max date of the current date. Another thing to consider with date validations is you can now copy/paste or type the dates.

 I think I am going to get started on this ticket soon since it's going to block some other upcoming tickets, I will play around with setting a max date with the new datepicker.

MarsRomer:
[...]

PR #1868, opened 2023-07-07
1828 merge project information change request

marcellmueller:
This ticket is to merge project information + change request, add change request read only form and ability to edit change request.

Will still need to pass the correct data to the read only change request once #1770 add change request metadata is complete though I think we should get this merged in sooner than later as there is a lot of incoming SoW/change request tickets.

Implements #1828


VB-at-Bis (approved):
LGTM

Issue #1851, opened 2023-07-05
sow-upload api 'filepath' error


marcellmueller:
<img width="1016" alt="Screenshot 2023-07-04 at 5 06 49 PM" src="https://github.com/bcgov/CONN-CCBC-portal/assets/14259474/95f45d95-587a-46d7-8cb5-b85fec875c35">

Something is causing our sow-upload api to crash with an `undefined filepath` error. This is causing pods to crash and I believe other strange behaviour in test. 

Edit: this can be reproduced by saving a SoW form, then refreshing the page, hitting the edit button then saving again. This is because it tries to send the file to the api but there is no actual file attached to the file input, just the file data in the form.

We could fix this by not posting to the api if there is no actual file in the file input as well as better error handling in the api.

AntBush:
Seems like it's also happening in the gis-upload: https://bcgov-ccbc-yr.sentry.io/issues/4109830104/?environment=test&project=4504079285944320&query=is%3Aunresolved&referrer=issue-stream&stream_index=15

MarsRomer:
not sure if it's related to this ticket but the GIS upload now does not work.  I uploaded the attached file, but the continue button is unresponsive now.

MarsRomer:
I was able to make it to the preview screen with a different json and so I cancelled out and tried again with the previous json but again the continue button was unresponsive.  Then when I tried to go to the dashboard I got this ![image.png](https://images.zenhubusercontent.com/63c8370399512153492c8e97/94de6b2a-51a4-417b-82eb-e25de743fb00)

marcellmueller:
@MarsRomer can you please send us the JSON you are using? I just tried with a couple different ones and can make it through in test. However I can see that something is causing pods to crash from gis upload page.

Your JSON might be fine but it would help us eliminate 1 extra variable thanks :)

PR #1869, opened 2023-07-07
feat: failsafe to prevent filepath error

VB-at-Bis:
Implements #1851


app/backend/lib/gis-upload.ts
```diff
@@ -71,7 +71,9 @@ gisUpload.post('/api/analyst/gis', limiter, async (req, res) => {
 
   const filename = Object.keys(files)[0];
   const uploaded = files[filename];
-
+  if (!!uploaded === false) {
```
AntBush, line 74: Couldn't this be simplified with just doing: 
```suggestion
  if (!uploaded) {
```

since it'll still implicitly convert anything falsey to a true.

AntBush (approved):
🚀

PR #1893, opened 2023-07-12
chore: set margins for inputs and questions

rafasdc:
set margins for inputs to 8px top to separate them, set margin bottom to container div to 16px to separate questions

Implements # 1802


marcellmueller (approved):
Nice work!

PR #1895, opened 2023-07-12
chore: connect via pgbouncer

rafasdc:
Implements # 888


rafasdc:
Need to grant access to pgbouncer role before merging this in.

marcellmueller (approved):
🕺

Issue #1777, opened 2023-06-26
Validate Amendment Number in Change Request


MarsRomer:
As an analyst
I want want to be warned when I have used the same amendment number as before
So that I don't make a mistake entering the same amendment twice or numbering incorrectly

**Description:**
This ticket is to add validation to the Amendment Number in the Change Request form 

# Acceptance Criteria:
- [x] Validate that the amendment number is not already in use, on keystroke (consistent with how the url on announcements is validated)
- [x] If same number then highlight in red and and add text message :'Amendment number already in use'
- [x] when they type in the correct number or delete the number the text message disappears on keystroke
- [x] Validate the amendment number on Save with same behavior as above i.e. red highlight and text message.

![image](https://github.com/bcgov/CONN-CCBC-portal/assets/93939894/bb2d9b8a-b946-40c1-ba69-f9545ab30982)

[wireframe 1]

## Implementation Details
Validate in the database on save

PR #1909, opened 2023-07-13
1777 validate amendment number

marcellmueller:
Implements #1777 

Also updated some change request form styles. Please check this closely for bugs, especially in the `handleSave` section there was a lot more logic I had to add than expected 🙇‍♂️


rafasdc (commented):
When you add a new amendment and then edit the number, the UI will temporarily show correct as an update but a new one gets created. For example, if I create amendment 19, then realized it should have been 18, go to edit and change to 18 it will show ok on the UI but after a refresh both 18 and 19 will be visible and exist on the DB.

marcellmueller:
> When you add a new amendment and then edit the number, the UI will temporarily show correct as an update but a new one gets created. For example, if I create amendment 19, then realized it should have been 18, go to edit and change to 18 it will show ok on the UI but after a refresh both 18 and 19 will be visible and exist on the DB.

Damn, nice catch. I updated the mutation to accept the old rowId of the previous change request and use that to archive instead of the amendment. 

Only small caveat I noticed is that if you edit a change request to a new amendment number, then try and reuse the old one it says it is taken until you refresh the page. Though I kind of think that could be it's own ticket if we see it as an issue.

marcellmueller:
@rafasdc can you check this out again when you have time, I think the edit change request amendment number issue is taken care of thanks.

rafasdc:
> [ ](/marcellmueller)

The edit is still being weird for me, I can reuse a previous amendment number on the same screen before refreshing, seems the UI is updating properly but the DB is still not archiving or updating the edited amendment. 

image

marcellmueller:
@rafasdc ah sorry about that 😓  I guess I had too many change requests and didn't notice. It was a couple issues, I needed to use the `rowId` for the change request plus it was using the wrong `old_change_request_id` in the postgres mutation itself.

app/components/Analyst/Project/ProjectForm.tsx
```diff
@@ -14,7 +14,7 @@ const ToggleRight = styled.div`
   align-items: center;
   margin-left: auto;
   margin-top: auto;
-  margin-bottom: auto;
+  margin-bottom: autoo
```
rafasdc, line 17: Possible typo? auto;

marcellmueller: Fixed!

app/components/Analyst/Project/ProjectInformation/ProjectInformationForm.tsx
```diff
@@ -191,29 +223,57 @@
       } else if (isSowUploaded) {
         delete newFormData?.isSowUploadError;
       }
-
       if (isChangeRequest) {
+        console.log('old change request id', currentChangeRequestData?.rowId);
```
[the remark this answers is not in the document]
marcellmueller: Fixed!

rafasdc (approved):
Working great now, awesome work! 🥂

PR #1930, opened 2023-07-18
change request read only

marcellmueller:
This is just passing the change request metadata to the read only form to complete ticket #1828


db/test/unit/computed_columns/application_amendment_numbers_test.sql
rafasdc, line 1: Got some possible merge stuff in here?

marcellmueller: Looks like it, thanks!

rafasdc (approved):
👍

Issue #1122, opened 2023-01-16
Admin - Add Intake


MarsRomer2:
As an administrator
I want to be able to add an intake into the future
So that I can directly control when intakes open and close

# Acceptance Criteria:

## Page: View Intakes dates
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/80f3eae9-8f09-40a9-b6ac-0c4bb5dbacab)

### Fields and info
- [x] "Add Intake" button
  
### Button Behaviour
- [x] **Visible when there is no future intake already (Future intake = start date > current date).**  
  _i.e. It can only have one intake in the future - one more than the current open intake.  
  If no intake is currently open, they can still only have one intake in the future. Once the intake opens, then they are allowed to add another intake into the future._
- [x] Hide "Add Intake" button
- [x] **Transition:** slide down - similar to: https://codepen.io/mbxtr/pen/OJPOYg

## Page: Add Intake
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/45e2852e-2543-4110-bd3b-7bc1a17d4469)

### Fields
- Intake # (Autogenerated) 
- Start Date
- End Date 
- Description (optional)
#
- [x] **Show next Intake number ** Gapless, sequential even if intake is soft-deleted
#
- [x] **Date and time are Pacific Timezone (PT)**
 (ensure this accounts for daylight savings time when necessary)

### Validations
- [x] Live validations on dates being correct
- [x] When pressing Save validate that the two dates are filled in
#
- **Required fields:**
- [x] Start date = null
Error message:  "Start date and time must be entered"
- [x] End date = null
Error message:  "End date and time must be entered"
#
- **Error state design:**
- [x] Field outlined red
- [x] Red warning text appears below each field
- [x] "Description (optional)" 
- [x] Description field - 150 characters
#
- [x] Scenario: Start date & Time > End Date from the 'last in the future' existent intake. 

- [x] The intake period cannot overlap with other existent intakes.
The `Start Date` must be `greater than` the `End Date from the chronologically previous intake`.
I.e. two intakes cannot be open at the same time.
Error message: "Start date & time must not overlap with other intakes"
#

- [x] Start date & time > current date and time
Error message: "Start date & time must be after current date & time"
- [x] End date & time > start date and time
Error message: "End date & time must be after start date & time"
#

[...]

PR #1950, opened 2023-07-21
1122 admin add intake

marcellmueller:
Implements #1122


rafasdc (commented):
This is working great, only thing I noticed is that you create an intake where start date === end date: 
image

marcellmueller:
> This is working great, only thing I noticed is that you can create an intake where start date === end date: image

I added a check in the database mutation that will throw if the start date is greater than or equal to the end date, as well as a pgtap test for it. Let me know if you can reproduce this on the front end again though! Thanks.

rafasdc:
> 

The validation doesn't seem to trigger when there is an intake open, this is what I did: 

- Create a new intake that starts in 1 to 2 minutes
- Refresh the page once the new intake "starts"
- Create a future intake
- No validation

Result:

image

marcellmueller:
> 

Screenshot 2023-08-01 at 10 58 34 AM

I can't reproduce this

rafasdc:
> > 
> 
> Screenshot 2023-08-01 at 10 58 34 AM
> I can't reproduce this

I think this is probably fine the way it is, it is a very very edge case.

rafasdc (approved):

PR #1964, opened 2023-07-24
feat: add info tooltip on applicant dashboard

rafasdc:
Implements # 360


AntBush (approved):
I think we have a Tooltip component, but I'm not entirely against using the MUI one either

PR #1969, opened 2023-07-25
fix: set the sequence value to match what is set on override

AntBush:
Implements fix where creating an intake produces an error in deev data. @marcellmueller

### DIFF
--- a/db/data/dev/001_intake.sql
+++ b/db/data/dev/001_intake.sql
@@ -31,4 +31,6 @@ open_timestamp = excluded.open_timestamp,
 close_timestamp = excluded.close_timestamp,
 ccbc_intake_number = excluded.ccbc_intake_number;
 
+select setval('ccbc_public.intake_id_seq', 2, true);
+
 commit;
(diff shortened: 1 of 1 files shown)


db/data/dev/001_intake.sql
```diff
@@ -31,4 +31,6 @@ open_timestamp = excluded.open_timestamp,
 close_timestamp = excluded.close_timestamp,
 ccbc_intake_number = excluded.ccbc_intake_number;
 
+select setval('ccbc_public.intake_id_seq', 2, true);
```
marcellmueller, line 34: Hey I appreciate this, I just tried it but looks like we also need to set the gapless counter id sequence. I'll approve this but it needs this small update 🙇‍♂️ 

```suggestion
select setval('ccbc_public.intake_id_seq', 2, true);
select setval('ccbc_public.gapless_counter_id_seq', 2, true);
```

marcellmueller (approved):

db/data/dev/001_intake.sql
AntBush: Done!

Issue #1124, opened 2023-01-16
Delete future Intake 


eazeredo:
## User Story
As an administrator
I want to be able to delete the future intake
so that I can maintain an organized list of intakes

## Acceptance Criteria
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/16626302-b1f7-48dc-88b5-a9e486269a73)

- [x] Applicant can delete the intake that hasn't started yet.  (There should only be one intake in the future)

### Validations
- [x] **Show Delete action** for intakes if intake's start date & time > current date  & time
<s>- [ ] When click delete, a confirmation modal opens stating, "Are you sure you would like to delete Intake {# Start date & time - End date & time}. </s>

### Action
- [x] Soft delete the intake
<s>
- [ ] When "Delete" is clicked in the modal, the modal closes and an updated list of intakes are shown
- [ ] When "Cancel" is clicked in the modal, the modal closes and no changes are made
</s>
## Technical Details:
<s>- Will require a custom mutation to delete an intake and then update the remaining intakes</s>

elliottSlee:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1124?planning-poker) with Zenhub @rafasdc

PR #1988, opened 2023-07-27
1124 delete future intake

marcellmueller:
Implements #1124 \<issue number\>

### DIFF
--- a/app/components/Admin/Intake.tsx
+++ b/app/components/Admin/Intake.tsx
@@ -1,6 +1,9 @@
 import { graphql, useFragment } from 'react-relay';
 import styled from 'styled-components';
 import { DateTime } from 'luxon';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faClose } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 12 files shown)


app/components/Admin/Intake.tsx
```diff
@@ -71,28 +101,59 @@ const Intake: React.FC<IntakeProps> = ({ currentIntakeNumber, intake }) => {
...
   );
 
-  const closeDate = DateTime.fromISO(closeTimestamp).toLocaleString(
+  const formattedCloseDate = DateTime.fromISO(closeTimestamp).toLocaleString(
     DateTime.DATETIME_FULL
   );
 
+  const currentDateTime = DateTime.now();
+  const endDateTime = DateTime.fromISO(closeTimestamp);
+  const isAllowedDelete = currentDateTime <= endDateTime;
```
rafasdc, line 116: Should this be using the open timestamp instead of the close one, it is currently letting me delete in progress/open intakes. 

Something like this: 

` const startDateTime = DateTime.fromISO(openTimestamp);`
 ` const isAllowedDelete = currentDateTime <= startDateTime;`

marcellmueller: Ah yeah good catch, fixed!

rafasdc (approved):

PR #1992, opened 2023-07-28
Allow applicant to delete drafts on dashboard

rafasdc:
Implements # 703


app/components/Dashboard/Table.tsx
```diff
@@ -40,8 +41,11 @@ type Props = {
 
 const Table = ({ applications }: Props) => {
   const [withdrawId, setWithdrawId] = useState<null | number>(null);
+  const [archiveId, setArchiveId] = useState({ rowId: null, id: null });
 
-  const applicationNodes = applications.allApplications.nodes;
+  const applicationNodes = applications.allApplications.edges
+    .map((edge) => edge.node)
+    .filter((node) => node !== null);
```
marcellmueller, line 48: Yep I am doing this as well when deleting nodes from a connection, wonder if there is a way we can avoid this.

PR #2026, opened 2023-08-01
Re unite orphaned rfi files

rafasdc:
Implements # 1907


app/lib/theme/components/FileComponent.tsx
```diff
@@ -209,23 +238,79 @@ const FileComponent: React.FC<FileComponentProps> = ({
...
+            >
+              <StyledDatePicker
+                id={id}
+                sx={getStyles(false)}
+                disabled={false}
+                readOnly={false}
+                onChange={(d: Date | null) => {
+                  if (d) {
+                    const originalDate = new Date(d);
+                    const realDate = new Date(originalDate.toDateString());
```
AntBush, line 265: Is this to set it without Hours/Minutes/Seconds? Why don't we just do something like `originalDate.setHours(0,0,0,0)` to zero-out everything. Would read more clear to me

app/lib/theme/widgets/FileWidget.tsx
```diff
@@ -120,6 +123,7 @@ const FileWidget: React.FC<FileWidgetProps> = ({
           name,
           size,
           type,
+          ...(useFileDate ? { fileDate } : {}),
```
AntBush, line 126: I thought we were going to pass back the `created_at` if we weren't using the `useFileDate` here? Or is that just in a different ticket

app/lib/theme/components/FileComponent.tsx
```diff
@@ -209,23 +238,79 @@ const FileComponent: React.FC<FileComponentProps> = ({
...
+                enUS.components.MuiLocalizationProvider.defaultProps.localeText
+              }
+              dateAdapter={AdapterDayjs}
+            >
+              <StyledDatePicker
+                id={id}
+                sx={getStyles(false)}
+                disabled={false}
+                readOnly={false}
+                onChange={(d: Date | null) => {
```
AntBush, line 262: When I choose a date on the date picker, I get a an error of `TypeError: value.isBefore is not a function`. I remember this is from not using dayjs when setting the date I think was the problem.

AntBush (commented):
Looks good, running into a bug when setting the date

app/lib/theme/components/FileComponent.tsx
```diff
@@ -209,23 +238,79 @@ const FileComponent: React.FC<FileComponentProps> = ({
...
-        >
-          {loading ? (
-            <LoadingSpinner color={isSecondary ? '#000000' : '#fff'} />
-          ) : (
-            buttonLabel()
-          )}
-        </StyledButton>
+      <div
+        style={
+          useFileDate
```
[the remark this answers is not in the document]
rafasdc: Agreed, changed

app/lib/theme/widgets/FileWidget.tsx
rafasdc: That's a different ticket, it should make more sense now that the other ticket is merged. I'm using `fileDate` to capture the "Date file received" field which might be different than the uploaded date, if that makes sense.

app/lib/theme/components/FileComponent.tsx
rafasdc: I believe this is to handle the time zone stuff and set it in the correct order, it was taken directly from the DatePickerWidget file, only difference is the setter.

app/lib/theme/components/FileComponent.tsx
```diff
@@ -144,12 +159,25 @@ const FileComponent: React.FC<FileComponentProps> = ({
...
 }) => {
   const hiddenFileInput = useRef() as MutableRefObject<HTMLInputElement>;
   const isFiles = value?.length > 0;
   const hideIfFailed = !!error && hideFailedUpload;
   const isSecondary = buttonVariant === 'secondary';
 
+  // eslint-disable-next-line react/no-unstable-nested-components
+  const ClearableIconButton = () => {
+    return (
+      <button type="button" onClick={() => setFileDate(undefined)}>
```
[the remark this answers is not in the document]
rafasdc: This is a workaround as the datepicker won't accept a component with props, based of the original DatePickerWidget.

app/lib/theme/components/FileComponent.tsx
rafasdc: This should be good, give it another try when you get a chance :)

PR #2027, opened 2023-08-02
feat: cannot save outdated data

AntBush:
Implements # 1212


marcellmueller (approved):
Nice work, this seems to be working well!

Issue #1900, opened 2023-07-13
Show RFI attachments in Summary page


MarsRomer:
As an analyst 
I want to be able to see the uploads the applicant did in the summary page
So that all the info I need to assess an application is in one place

**Description:**
Now that the analyst can download the attachments from the application on the summary form, there is a request to also be able to see the RFI attachments in the summary page in the appropriate places.  

# Acceptance Criteria:
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=12202-132683&mode=design&t=pTgJ1NRuwszdA7LM-4)
- [x] Analyst can see the RFI uploads by the applicant (and analyst) in the appropriate places on the summary page
- [x] The RFI uploads are visible as hyperlinks to the documents

**RFI replaced a single file upload**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/c952b7f0-2a13-4d19-8441-9b060c75b6e4)
**Two RFIs for the same single file upload**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/7c37c92f-bf9b-44a6-8a1f-2810ac6044e3)
**One RFI for multi-file upload**
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/d634a7b9-0b78-4d3f-88de-ef848985bf20)

- [x] Below each file, right aligned, display the date & time the file was uploaded by the applicant or the time entered by the analyst
- [x] Below a group of files from the same RFI, display the RFI #
> Received from RFI-CCBC-000000-0 
- [x] no spacing between filename & "Received from..." or "Original"
- [x] 8px between groups of files
- [x] icon is https://fontawesome.com/icons/star?f=classic&s=solid&sz=2xs **size: 2xs**
- [x] icon only appears on most recent if the section is single file upload
- [x] icon does not appear on any RFI of multi-file uploads
- [x] After an RFI is received, add the word "Original" to the original file(s)
- [x] filename and text below ("Original" or "Received from...") are left-aligned
- [x] icon is to left of the filename and text below
- [ ] indent any rows with**out** a star icon (it may be 14px indent if icon is 12px wide and there's 2px spacing between icon & text

MarsRomer:
[...]

PR #2030, opened 2023-08-02
1900 show rfi attachments

marcellmueller:
- feat: add sort rfi file util function
- feat: display rfi files on application page

Implements #1900


app/components/Review/widgets/FileWidget.tsx
```diff
@@ -1,22 +1,144 @@
...
+        if (!isFileField) return null;
+
+        const attachments = rfi?.attachments?.nodes;
+
+        return (
+          <StyledFile key={rfi}>
+            {rfiFiles[fieldName]?.map((el, index) => {
+              // loop through the list of files for the current field
+              const isSingleFile = !options?.allowMultipleFiles;
+              const isDisplayIcon = isSingleFile && i === 0;
```
rafasdc, line 73: What's the purpose of the i === 0? Wouldn't this prevent the latest RFI to have the star? Something like this? 
image

rafasdc (commented):
This is working great, just a small question. I also added a general comment to the issue as no stars will show once this is merged as all rfi files are now multi file uploads.

app/components/Review/widgets/FileWidget.tsx
marcellmueller: In the graphql query we are sorting them by `orderBy: RFI_DATA_ID_DESC` and the `i === 0` is so we display the icon for the first rfi (and newest rfi) that we map. 

The files below the rfis are mapped separately so I am confused as to what you mean. I just double checked and it's working for me:

Screenshot 2023-08-11 at 9 27 56 AM

rafasdc: Sorry! This makes sense now, I was looking at it from the wrong way.

rafasdc (approved):

app/components/Review/widgets/FileWidget.tsx
marcellmueller: No worries thanks 🙂

Issue #1954, opened 2023-07-24
Community Progress Report (CPR) - Import data from Excel file


MarsRomer:
As an analyst
I want the CPR data from the excel to be imported into the database
So that I don't have to spend valuable time typing it in

**Description:**
 This ticket is to import the CPR data from the excel file

[Excel file with highlighted fields to import](https://bcgov.sharepoint.com/:x:/r/teams/04708/Shared%20Documents/Research/Community%20Progress%20Report_mock.xlsx?d=w6778cac29f544933b66c50bc4e8c9284&csf=1&web=1&e=KGahCH)

# Acceptance Criteria:
- [x] Excel file is imported and ingested into the database
- [x] When excel file is uploaded and validation is successful, then the Save button becomes the 'Save and Import data' button 
- [x] If the excel file is subsequently removed, the Save and Import Data button becomes the 'Save' button again
- [x] When the 'Import and Save' button is clicked, it imports data from the excel file (spinner and success toast are in #1959)

## Implementation Details

DB - one table ApplicationCommunityReportData

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1954?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc

PR #2042, opened 2023-08-03
1954 import community progres report data

marcellmueller:
Changed `SowImportFileWidget` to `ExcelImportFileWidget` as well as the `sowValidate` function was refactored and changed to `excelValidate`. Maybe just double check that the SoW import and Change Request import are working well still, thanks.

Implements #1954


app/components/Analyst/Project/ProjectInformation/widgets/ExcelImportFileWidget.tsx
```diff
@@ -187,7 +187,10 @@ const SowImportFileWidget: React.FC<SowImportFileWidgetProps> = ({
   const [createAttachment, isCreatingAttachment] = useCreateAttachment();
   const [deleteAttachment, isDeletingAttachment] = useDeleteAttachment();
   const [isImporting, setIsImporting] = useState(false);
-  const [isValidSow, setIsValidSow] = useState(false);
+  const [isValidExcel, setIsValidExcel] = useState(false);
+  const excelImportOptions = options?.excelImport;
+  const successHeading = excelImportOptions['successHeading']?.toString();
```
AntBush, line 192: nit: I think the `toString()` here can cause our code to be brittle. If at one point the value is a number which would pass the `?` check we'll get an error. I'm cool with leaving it in since I doubt we'll ever have the success heading be a number but can cause an error down the line.

I think it might also produce weird behaviour if we provide it with JSX, probably get a `[object Object]` as the heading.

Could throw some checks in the `renderStatusLabel` instead?

AntBush (approved):
Good stuff! Kudos for refactoring the `SowFileImportWidget`, I like the generalized ExcelImportWidget. 

Just a tiny nit, up to you if you want to make the change

PR #2065, opened 2023-08-10
Community progress report import spinner and success toast

rafasdc:
Implements # 1959

Note: you might need to add a "fake" two second timeout to see the spinner as the import is very quick.


marcellmueller (approved):
Looks good. I did notice a bug that isn't related to this ticket though it is pronounced by it, if I upload 3 community progress reports in a row the spinner will stall on the third one.

I tried it in another branch and I had to hit save a couple times for it to work, not sure if it is due to the rate limiting or what.

rafasdc:
> Looks good. I did notice a bug that isn't related to this ticket though it is pronounced by it, if I upload 3 community progress reports in a row the spinner will stall on the third one.
> 
> I tried it in another branch and I had to hit save a couple times for it to work, not sure if it is due to the rate limiting or what.

I did notice that when I was testing the view, my network tab was showing "Too many requests, please try again later" I believe this is the rate limiting against the API route.

PR #2073, opened 2023-08-11
fix: big file upload boxes

rafasdc:
Implements # 2074


marcellmueller (approved):
Thanks for fixing this!

PR #2096, opened 2023-08-14
fix: add greater buffer time to prevent modal popup on intake form

AntBush:
Implements # \<issue number\>


marcellmueller (approved):
Looks good!

Issue #255, opened 2022-06-24
RJSF custom widget and props documentation


marcellmueller:
We have now created quite a bit of custom functionality into our RJSF setup so I think it would be worth documenting, especially if we plan on adding some of this to the toolkit.

PR #2106, opened 2023-08-15
2055 suppress aws logs

marcellmueller:
Adds a new env var called `ENABLE_AWS_LOGS` which is set to true by default in `test` and `prod`.

Implements #255


AntBush (approved):
:100:

PR #2109, opened 2023-08-15
feat: ui does not change when selecting no

AntBush:
Implements # 1961


marcellmueller (approved):
This is working well, nice work!

PR #2110, opened 2023-08-15
Login changes

rafasdc:
Implements # 1673


AntBush (approved):
Login flow still works here on dev, hoping this will resolve the issues :crossed_fingers:

PR #2119, opened 2023-08-16
feat: reduce buffer to only two seconds

AntBush:
Implements # 1212 reducing the buffer, but won't be merged in until we conclude that this is what we want.


AntBush:
Tested again without time machine, going to close this since it is not required

PR #2134, opened 2023-08-18
feat: archive scraped data when deleting file 

AntBush:
Implements # 2089


marcellmueller (approved):
Cool solution, I was expecting to use archive mutations separately on save.

Issue #2002, opened 2023-08-01
Claims - Add


elliottSlee:
As an analyst
I want to enter the claim & progress report information & file
So that we can have it in the portal with all other quarterly reports

**Description:**
This is for the form of entering & uploading the claim & progress report Excel file

# Acceptance Criteria:
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/ab30a05d-f51b-4726-b172-e5397f6a4a41)
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/89262efe-0932-4bd6-ab09-abb875aee094)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=12009-130547&mode=design&t=awLlJr6x7ZLrfYN8-4)

- [x] Click add claim & progress report, animation similar to announcements & funding agreement forms where this slides up and the form slides down
- [ ] Claim period from & to dates are required
- [x] Analyst can upload the report - for this ticket the upload only saves the file, does not import the data - not mandatory. (Note: saves it on the file widget and form data, but hitting submit will go nowhere as there is no backend or DB table)
- [x] When uploading the file, the spinner shows
- [x] Click cancel and the form closes, nothing is saved

elliottSlee:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2002?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

MarsRomer:
Looks good, the only thing I noticed is that the to and from dates are not required.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbEpWIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--dbc4f56275141dc7f16a3de1bd558d02f1d37806/image.png)

marcellmueller:
@MarsRomer I'm confused by this comment. We closed a ticket on Claims form validation:
https://github.com/bcgov/CONN-CCBC-portal/issues/2147

Also we have a ticket in progress to remove these fields altogether and pull that data from the excel:
[...]

PR #2145, opened 2023-08-21
2002 add claims

marcellmueller:
To save time in upcoming claims PRs there is  a decent amount of commented code since I based this off of the `Community progress report` form which is nearly identical.

I have also added the `application_claims_excel_data` table and mutations along with a basic api route that save but has no validation or anything to do with the excel file.

Implements #2002


marcellmueller:
Closing since we are merging in #2148 which is branched off of this PR.

Issue #2003, opened 2023-08-01
Claims - View


elliottSlee:
As an analyst
I want to see the read-only version of the claims that were submitted
So that I can view the claims submitted

**Description**
This ticket is for the user to see the list of claim and be able to download, and edit 

# Acceptance Criteria:
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/57d29309-3975-456e-903b-48e90ee6cfbe)
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=12010-129103&mode=design&t=awLlJr6x7ZLrfYN8-4)

- [x] After pressing save on the form, this view slides in and the form slides out (same animation as announcements & the funding agreement forms)
- [ ] Ordered by claim #, highest on top & lowest on bottom
- [ ] Claim # is imported from Excel
- [x] Dates are the claim period from & to dues. Display only the months & year, not day. If the time spans more than 1 year, just display the "to" year
- [x] Click on "Claim & Progress Report" link in the view begins the download of the original file
- [x] When user clicks the edit on the view page - the add form opens with form data

elliottSlee:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2003?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

MarsRomer:
ok sounds good @marcellmueller , if you do want to sort the the claims in the mean time feel free to use the claim dates.  Oldet at the bottom newest on top

PR #2146, opened 2023-08-22
2003 claims view

marcellmueller:
Branched off of #2002 

Implements #2003


marcellmueller:
Closing since we are merging in #2148 which is branched off of this PR.

Issue #2092, opened 2023-08-14
Investigate increasing rate limit for excel api routes


marcellmueller:
We should investigate increasing our rate limiting for uploading excel documents. Currently if you upload about 3 in a row you will trigger the rate limit, causing the spinner to stall. We need to find a nice balance that won't overload or crash our pods but that can allow a user to submit multiple files in a row.

- [x] monitor resources used while uploading excel in openshift
- [x] adjust rate limit if our resources allow

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2092?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc

marcellmueller:
@MarsRomer Not sure if you wanted to review/test this you can now upload 3 (or more) excel files per minute in a row without the spinner timing out.

MarsRomer:
I tried, but the uploads are so quick I can't keep up! hahaha

PR #2158, opened 2023-08-25
chore: increase excel api rate limit

marcellmueller:
Increased the limit to 20 connections per IP per minute, which seems like a lot though the reason we were maxing out our 5 during testing is that we hit the API twice per upload, once to validate and the second to save. By the time we would get to saving our third excel file we would be at the limit of 5, so it would timeout and hang on save.

10 uploads per minute should be more than enough as it would be hard to hit even intentionally. 

Implements #2092


VB-at-Bis (approved):
Looks good

PR #2176, opened 2023-08-30
2006 claims excel validation

marcellmueller:
- feat: add claims excel validation
- fix: don't show community progress excel spinner if there are validation errors
- test: update claims tests due to validation changes
- test: add claims excel validation tests


rafasdc (approved):
👍

Issue #2161, opened 2023-08-28
Claim number validation and display


MarsRomer:
As an analyst
I want to see the claim number
So that I know what order the claims were received

**Description:**

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcmRaIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--bda3c07b73ae858c7cbb3e85f96974bcf3e22d3f/Screenshot%202023-08-31%20at%2010.21.57%20AM.png)

# Acceptance Criteria:
- [x] Display claim number from the uploaded claims document on the Claims view
- [x] Display list of claims in descending order on claim number
- [x] Validate that the same claim number was not already used when they upload a new claim (show error message), see related #2006
- [x] Validate that they upload the same claim number document when they edit an already existing claim (show error message), see related #2006

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2161?planning-poker) with Zenhub @marcellmueller @VB-at-Bis

PR #2191, opened 2023-08-31
2161 claim number validation and display

rafasdc:
Implements #2161


marcellmueller (approved):
Nice job, this is working really well. Only tiny thing I noticed is the example claims excel I've been using has claims no 999 and looks like 3 digit claims numbers display kind of weird. Not sure what kind of numbers to expect for claims.

Screenshot 2023-09-06 at 2 59 38 PM

PR #2192, opened 2023-09-01
fix: claims from year width overlap

marcellmueller:
Fix for dates overlapping claims excel file name due to increased width from year being added. 
Screenshot 2023-09-01 at 10 13 07 AM


VB-at-Bis (approved):
Looks safe 👍

Issue #400, opened 2022-08-10
Application question changes in future intake


zoeyli-46:
### User Story
As an analyst
I want to be able to remove the estimated project employment page
so that I don't waste ISP time collecting info we don't use

### Acceptance Criteria
- [x] Remove Estimated project employment page from new applications for intake 3
- [x] Remove Estimated project employment link from stepper for analyst and applicant
- [x] Keep data from past intakes in metabase and front end for applicant and analyst

Technical details:
- [x] Make sure form schema is versioned

zoeyli-46:
@MarsRomer this ticket is still convoluted, but logging this here for now as we work on the design for it

MarsRomer:
Thanks @zoey, please let me know what decisions you need from me to move this forward

zoeyli-46:
I added a screenshot in the description for an alert. 

When an applicant edits a draft application created in a previous intake, what are your thoughts on having this alert appear "Questions in this application may differ from previous intakes' applications. Please review your responses to ensure completeness."?

cc @elliottSlee

MarsRomer:
it's a good start.  We can also have a note about what the changes are on the landing page and it should also show red in the review page.  Right now I don't know of any new questions to be added to the second intake so lets not spend too much time on this.

elliottSlee:
To discuss: can applicants see changes in their draft applications? Or are the question changes only visible after intake open date? Latter requires #367 - Time Travel

zoeyli-46:
If an applicant is working on their draft application during the gap between an intake, they may feel their time is wasted if when the intake opens the questions that they spent time on have changed 

Though, if they receive updates to their draft applications during the gap between an intake, they may see there are discrepancies between the application questions document and the form on the site, and they may be confused why there are differences

MarsRomer:
Do we need to sync on this ticket?

elliottSlee:
Can be combined with #40.

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/400?planning-poker) with Zenhub @marcellmueller @VB-at-Bis

MarsRomer:
[...]

PR #2193, opened 2023-09-01
2182 change application spike

marcellmueller:
This actually went pretty well, this needs some cleanup but I just made a `schemaV2` which is imported into the database at application startup using `importJsonSchemasToDb`. From there any application created using `create_application` mutation will use the most recent `intake` schema.

I had to make a few minor changes such as passing the `jsonSchema` associated with that application to the stepper and a couple other places but it seems to be working great. Applications created before `schemaV2` were imported will display the estimated project employment page, applications after will not and it all works pretty seamlessly. There may have been a few areas I didn't think about.


marcellmueller:
Closing as this was used and completed in #400

PR #2195, opened 2023-09-01
fix: fiscal year warning showing up while editing community report

marcellmueller:
Fix for the fiscal year validation error showing when editing a community progress report. The warning was showing since it was including the fiscal year from the report being edited.

The warning will still show up when editing if another report exists that is using the same fiscal quarter.


VB-at-Bis (approved):
Looks good

Issue #1886, opened 2023-07-11
GIS upload - error handling for invalid JSON


marcellmueller:
As an analyst
I want feedback when the JSON file is not uploading due to an error
so that I can resolve the issue
Description:
Invalid JSON files are creating unhandled errors `gis-upload` api and causing our pods to crash. In addition to that there is no user feedback for it so most people would just try again and then give up when it isn't working.

- [x] Screen design to follow spike
- [x] Properly handle broken JSON files so that pods don't crash
- [x] Use eslint to get error message and display on front end so user knows it has errors:
   "Error detected on JSON file:
    [message] at line # and column#
    Please check your file and try again.
e.g.  Error detected on JSON file:
        Parsing error: Unexpected token at line 20 and column 2
        Please check your file and try again.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBc1ViIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--7fcc80039c569370b27164dd9b8e2bdc07edf6e5/image.png)

marcellmueller:
@MarsRomer I created a ticket here for invalid JSON error handling.

MarsRomer:
Blocked until #1978 is completed

MarsRomer:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1886?planning-poker) with Zenhub @VB-at-Bis

AntBush:
We have a branch with some sample code `1978-granular-json-errors`

PR #2201, opened 2023-09-05
Handling Gis JSON file upload errors

VB-at-Bis:
Implements #1886 

- added JSON validation based on code from `eslint-plugin-json` ;
- updated Gis page to match screen design;


marcellmueller (approved):
seems to work well, nice work!

PR #2202, opened 2023-09-05
test: increase project test excel api route timeout

marcellmueller:
Increasing the timeout for excel upload tests. This doesn't add to the total time of the test unlike `cy.wait`, just gives more breathing room before it fails due to runners being slow.


rafasdc (approved):
👍

PR #2211, opened 2023-09-06
fix: multiple tabs error being shown erroneously

marcellmueller:
Partial fix for the tabs error. I discovered from logging that the time difference that was causing this error to happen was very large, up to 30-45 seconds some times. I was unable to discover what was the cause of this, it seemed like relay was updating the store correctly and from what I could tell the `useDebouncedMutation` was disposing of the previous call correctly, though I suspect it has to do with the debounce. 

Anyways I have kept the 3 second buffer allowed before triggering the error when `last_edited_page` is different than the one being saved. This will still mitigate the potential loss of data (ie losing the acknowledgements checklist due to the form being saved in another tab).

What is changed is I have drastically increased the buffer time to 60 second when the page being saved is the same as the `last_edited_page`. This should hopefully stop the multiple tab warning being incorrectly triggered while filling out the form. There is more of a chance of data loss due to this but just on the same page. If they have the same page opened in two tabs and edit one quickly (within a minute) of the other then they would still lose data.

Seems like a fair trade off for now and which is why I am considering this a partial/incomplete fix.


VB-at-Bis (approved):
LGTM

Issue #1980, opened 2023-07-26
Keep spacing consistent between errors/helper labels


rafasdc:
<s>Right now the helper text and error messages will take space on the forms even though they might have no content which adds extra space in between inputs. 

- [ ] Make helper/error text not render if empty on applicable widgets

</s>

- [ ] Keep the spacing between helper and error messages consistent between existing and non existing errors/helper texts.

MarsRomer:
duplicate of #736

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1980?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc

Issue #1802, opened 2023-06-29
Update spacing of input elements


elliottSlee:
The spacing between input elements on all our forms is inconsistent. Let's make the default pattern 8px between label & input (input, textarea, radio, checkbox, date, etc.) and 16px between questions, unless otherwise stated.

Current:
![Screen Shot 2023-06-29 at 12.57.30 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/9e354850-39b6-42b5-9470-cb538ff122cb)

Desired:
![Screen Shot 2023-06-29 at 12.57.11 PM.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/46837ebb-8e85-4b7d-a18b-39a95131f46f)

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1802?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc @VB-at-Bis

marcellmueller:
@elliottSlee I'm not going to die on this hill but I looked into this and I think this will harm the general readability of our forms and make all the forms across our site feel more busy and claustrophobic. All of our form designs were made with this `32px` margin bottom in mind for the widgets. 

It's only a couple of main widgets that this margin was forgotten about such as the checkbox widget groups and textarea. I think it's worth considering just addressing the few widgets with missing margins than to reduce the space between all widgets.

elliottSlee:
Part of the issue we're trying to solve here is that forms with only a small handful of questions are longer than the screen, it shouldn't be this way. A prime example is the project management assessment which only has 5 inputs, but I can only see the top 2.5 when I go there
<img width="1552" alt="Project management assessment screen with too much padding" src="https://github.com/bcgov/CONN-CCBC-portal/assets/93939894/e2e66b73-6b7f-4423-95d3-a5bb93c4f0ec">
Reducing padding on the top of the main container and reducing size of the header are some things that would help. However, the spacing between inputs is quite big and taking up too much room too. I can't find much documentation or research on this, but I found several forms on BC gov sites that have 10, 15, 16, & 20px between input elements and 5 and 8px between label & input. I'd like ours to be similar, hence what's in the card.

marcellmueller:
[...]

PR #2230, opened 2023-09-11
chore: add space for helper/error labels

rafasdc:
Implements #1980

### DIFF
--- a/app/components/Form/Label.tsx
+++ b/app/components/Form/Label.tsx
@@ -2,6 +2,8 @@ import styled from 'styled-components';
 
 const StyledLabel = styled('label')`
   color: rgba(49, 49, 50, 0.7);
+  display: inline-block;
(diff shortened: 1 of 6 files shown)


marcellmueller (approved):
This looks good but I thought we were moving the help/errors to the `FieldTemplate` as is probably the proper way to go about this (we closed the previous ticket about that and combined it with this one)? 

Will approve since this still works either way.

marcellmueller:
@rafasdc was the goal to remove all margins between form fields? They look really squished together now. 

Screenshot 2023-09-11 at 4 07 15 PM

marcellmueller:
> Yeah, that's the intention coming from #1802

Yeah it's just that the reason this gap between form fields exists was to prevent buggy looking jumping if an error is displayed.

marcellmueller (approved):

Issue #2014, opened 2023-08-01
Milestone report - Excel Validation


elliottSlee:
As an analyst
I want to see feedback when there is something wrong with the file I am uploading
So that I can correct my mistake and upload again

# Acceptance Criteria:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcTBGIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--f050c5932c73e85b43185925f76e1bae255d6b9c/image.png)

- [x] Check CCBC # matches
- [x] Check for existence of tabs as expected
- [x] If validation fails then provide feedback (similar to SOW validation)
- [ ] <S>Errors persist onto read-only view</S>

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2014?planning-poker) with Zenhub @marcellmueller @rafasdc @VB-at-Bis

Issue #2009, opened 2023-08-01
Milestone report - Delete


elliottSlee:
As an analyst
I want to be able to delete a milestone row
So that I can correct for my mistakes

# Acceptance Criteria:
[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=12418-21645&mode=design&t=bgdLnqaU8Hg6gMyA-4)
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdEVTIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--de301cddd38094d272e3a3946f74f073aca555ba/image.png)

- [x] When user clicks on the red x on the view page for the claim, user sees Modal asking for confirmation
- [x] When user clicks 'Yes Delete' then modal disappears and the row with the claim is deleted (soft delete in backend)
- [x] All data that were imported from the Excel file to the database are archived
- [x] When user clicks 'No Keep' then the modal disappears and the row with the milestone report stays the same and data remains in database

MarsRomer:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2009?planning-poker) with Zenhub @VB-at-Bis

PR #2243, opened 2023-09-12
2014 milestone excel validation

marcellmueller:
Implements #2014 

Branched off #2009


VB-at-Bis (approved):
Looks good.
Expect conflicts on merge - I see same new files added like db/deploy/tables/application_milestone_data_001.sql etc.

Issue #2209, opened 2023-09-06
Dependabot


MarsRomer:
Enable dependabot PRs for all dependency updates, not just dependencies with potential security issues as is currently setup.

- [x] Dependabot opens PRs for outdated dependencies

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2209?planning-poker) with Zenhub @marcellmueller @rafasdc @VB-at-Bis

AntBush:
Updated env settings for dependabot, CI now running fine

PR #2245, opened 2023-09-12
chore: create dependabot.yml

marcellmueller:
Implements #2209 

This may require more tweaking after we merge it in, let me know if you think we are missing any useful settings. Set up from example here:

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#yarn


VB-at-Bis (approved):
Perfect

Issue #1979, opened 2023-07-26
Project accordions open or closed depending on forms


elliottSlee:
As an analyst
I want the project accordions to be open or closed depending on the stage of the project
So that my attention is directed to where necessary

# Acceptance Criteria:

![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/7343d69c-3072-4ec0-8bb9-28024846082f)
- [x] If conditional approval fields are not all filled, when a user comes to the page, that accordion is open and all others are closed
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/b2cb150a-220b-4ed2-8390-d949b23be382)
- [x] If conditional approval has all fields filled, when a user comes to the page, that accordion is closed and the FA, SoW, & map accordion is open
- [x] If  Has the funding agreement been signed by the province? = Yes  when a user comes to the page, then that accordion is closed
- [ ] Add amendment shows even when the accordion is closed
![image.png](https://images.zenhubusercontent.com/61ae5dab07758299775d9d78/4643abe3-d204-4298-ac96-7fd1ce388c95)
- [x] If Has the funding agreement been signed by the province? = Yes and current date is between Mar 1-Mar 30,  June 1-June 30, Sept 1-Sept 30, December 1-Dec 31 then Community Progress Report accordion is open
- [x] ⁠If  Has the funding agreement been signed by the province? = Yes and current date is between Mar 15-Apr 15, June 15 -July 15, Sept 15-Oct 15, December 15-Jan 15 then Milestone reports accordion is open
- [x] ⁠If Has the funding agreement been signed by the province? = Yes and current date is between Apr 15-May 30, then Claim & progress reports accordion is open

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/1979?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc

marcellmueller:
As we discussed I didn't do acceptance criteria `Add amendment shows even when the accordion is closed` as it doesn't really make sense without an updated design since everything is inside of the accordion.

PR #2261, opened 2023-09-14
1979 project accordion

marcellmueller:
Implements #1979

I also did a pretty big refactor of the project page jest tests and moved all of them except for the announcements to their own form tests in their `components` folder. This will make it a lot easier to work with as that file was getting way too large.


VB-at-Bis (approved):
Perfect

PR #2263, opened 2023-09-14
2213 get sharepoint access

rafasdc:
Implements # 2213

### DIFF
--- a/app/backend/lib/sharepoint.ts
+++ b/app/backend/lib/sharepoint.ts
@@ -1,0 +1,56 @@
+import { Router } from 'express';
+import * as XLSX from 'xlsx';
+import * as spauth from 'node-sp-auth';
+import config from '../../config';
+
+const SP_SITE = config.get('SP_SITE');
+const SP_DOC_LIBRARY = config.get('SP_DOC_LIBRARY');
+const SP_FILE_NAME = config.get('SP_MS_FILE_NAME');
+const SP_SA_USER = config.get('SP_SA_USER');
+const SP_SA_PASSWORD = config.get('SP_SA_PASSWORD');
+
+const sharepoint = Router();
+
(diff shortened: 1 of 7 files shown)


app/backend/lib/sharepoint.ts
```diff
@@ -0,0 +1,65 @@
...
+const SP_FILE_NAME = config.get('SP_MS_FILE_NAME');
+const SP_SA_USER = config.get('SP_SA_USER');
+const SP_SA_PASSWORD = config.get('SP_SA_PASSWORD');
+
+const sharepoint = Router();
+
+// eslint-disable-next-line consistent-return
+sharepoint.get('/api/sharepoint/masterSpreadsheet', (req, res) => {
+  const authRole = getAuthRole(req);
+  const isRoleAuthorized = authRole?.pgRole === 'ccbc_admin';
```
AntBush, line 18: Not needed for now, but do we want this to only be accessible to admins?

app/config/index.js
```diff
@@ -230,6 +230,36 @@ const config = convict({
     default: '',
     env: 'METABASE_SITE_URL',
   },
+  SP_SITE: {
```
AntBush, line 233: Worth updating the `.env.example` file with these? Since the endpoint will fail without them

AntBush (approved):
Looks good! Just some minor points that can be fixed in the future

app/backend/lib/sharepoint.ts
rafasdc: Agreed, I didn't want this to be open without any kind of role checks. 

I was thinking we can get the service account to get a token with the proper role, this can be admin, analyst or a custom one only for the service account.

app/config/index.js
rafasdc: Agreed, adding a sample.

Issue #2017, opened 2023-08-01
Milestone report - Add


elliottSlee:
As an analyst
I want to enter the milestone report information & file
So that we can have it in the portal with all other quarterly reports

# Acceptance Criteria:
[Add](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcFFGIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--6cbb42bcb9f0c6a5c1763b28f6faf37aa364387f/image.png)
[Form](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBZ1p4IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1aaece97b78d51946c7c32b9a7dd720f934fd4d0/image.png)

[**Wireframe**](https://www.figma.com/file/nhOquOVmWD1P3BxEY0KIh5/CCBC-Intake?type=design&node-id=12418-21611&mode=design&t=bgdLnqaU8Hg6gMyA-4)

- [x] Click add milestone report, animation similar to announcements & funding agreement forms where this slides up and the form slides down
- [x] Analyst can upload the report - for this ticket the upload only saves the file, does not import the data - not mandatory. (Note: saves it on the file widget and form data, but hitting submit will go nowhere as there is no backend or DB table)
- [x] Due date is entered on the form
- [x] Due Date is mandatory field.  Turns red on 'Save'
- [x] Date Received is not entered on the form - it will be collected from the excel upload (do not show on form)
- [x] When uploading the file, the spinner shows
- [x] Milestone report upload is single file
- [x] Evidence of milestone completion upload is optional - multiple files
- [x] Click cancel and the form closes, nothing is saved

MarsRomer:
Needs designs that remove the Due Date filed on the form

marcellmueller:
@elliottSlee why does the date field look like a dropdown in the wireframe?

elliottSlee:
@marcellmueller sorry, it should be a date.
The dropdown was intended to be a list of dates from the SoW, but I didn't end up doing that.

marcellmueller:
No worries thanks!

MarsRomer:
[...]

PR #2271, opened 2023-09-15
feat: allow multiple file uploads for milestones evidence of completion

marcellmueller:
Add support for multiple uploads for #2017


VB-at-Bis (approved):
Looks safe

PR #2330, opened 2023-09-26
chore: update owasp image for new version

rafasdc:
Implements # \<issue number\>


marcellmueller (approved):
Nice work! Thanks for figuring this out.

Issue #2214, opened 2023-09-07
Automate Updating of Master Spreadsheet


MarsRomer:
As an analyst
I want the master spreadsheet data in the portal database be updated automatically on a regular basis
so that I don't need to spend time managing that myself

Description:
This is to automate the updating of the master spreadsheet data in our database the xls on sharepoint

AC:
- [x] Upload from sharepoint on a regular basis, twice a day, noon and 430pm Pacific Time
- [x] Data is updated in our database 
- [x] Only upload the data IF the file was modified since the last time it was uploaded
- [x] our db captures the 'created at' based on our upload date and time
- [x] our db records the 'modified date and time' from sharepoint for when the xls was last modified by user

Implementation notes:
- [x] look into libraries for automation?

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2214?planning-poker) with Zenhub @marcellmueller @rafasdc @VB-at-Bis

rafasdc:
For automation an OpenShift CronJob is being used that runs a very basic nodejs script that triggers the API. To confirm max frequency.

MarsRomer:
@rafasdc will the change in bi-annual time zone be taken into consideration?

MarsRomer:
Hey Rafa can you please confirm that the AC for: Only upload the data IF the file was modified since the last time it was uploaded is in effect?  The test spreadsheet wasn't updated since yesterday but the script still ran today as I can see in the Error list  
[Uploading... image.png]

rafasdc:
Yes, the AC is working as expected. On the last triggers in test it returned `Response {"message":"No new data to import"}` 

Note that I believe that DEV also pushes to the same list and since that DB gets cleared very often it will always "think" the data has never been imported.

The DB on test has the correct data no imported twice, we might have to create a DEV list or disable the trigger on dev to avoid the list overly growing.

MarsRomer:
ok that's good to know, yes I think we should probably disable it in dev somehow, but keep it flexible to bring it back in dev if we need to

rafasdc:
I've set it to once a year which is they max we can give it, but we can keep pushing forward indefinitely. Next run is scheduled for October 15th, 2024.

PR #2336, opened 2023-09-27
2214 automate updating of master spreadsheet

rafasdc:
Implements #2214 

To test this locally: 

1. Do a post request to https://dev.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/token with basic auth where the username and password is the client id and client secret for the service account (you can get it at CSS). Set the body to be `x-www-form-urlencoded` with key `grant_type` set to `client_credentials`
2. Use the access_token received from step 2 to do a get request to http://localhost:3000/api/sharepoint/cron-cbc-project set the Authorization header to Bearer {access_token} 

You will also need to set the following in your .env to match the values you used in 1: 
KEYCLOAK_SA_CLIENT=''
KEYCLOAK_SA_CLIENT_SECRET=''


marcellmueller (approved):
Thanks for the walkthrough on this whopper, appreciate it

Issue #2170, opened 2023-08-29
Project Description 


MarsRomer:
As an analyst
I want to add and see a brief project description
So that our organization has a 1-2 sentence summary of each project

**Description:**
This ticket is to add the project description that NWBC comes up with 

# Acceptance Criteria:

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdlY2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e558e2cff75e35f3a0eece06d4f8ac99981e0e7c/image.png)
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdmQ2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--58600824fa94edc1d51b1ba7bb43b861576abcc8/image.png)
- [x] field is inline editing
- [x] max 320 characters
- [x] character count changes in real-time
- [x] max height of read only is 3 lines
- [x] Below small screen breakpoint, truncate text.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaGQ3IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--c7680b618084bd40f8e911e2c636e238456c090e/image.png)
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaGg3IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--40db68dbf29e46bef2878c591433fbcea231e1a6/image.png)

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2170?planning-poker) with Zenhub @marcellmueller @rafasdc @VB-at-Bis

elliottSlee:
@marcellmueller nice touch with the expanding animation. Looks great!

marcellmueller:
> @marcellmueller nice touch with the expanding animation. Looks great!

[...]

PR #2347, opened 2023-09-29
2170 project description

marcellmueller:
Implements #2170


marcellmueller:
@AntBush thanks again for the heads about about the internal project description. This is updated whenever you have time!

app/schema/mutations/application/createNewFormData.ts
```diff
@@ -7,7 +7,14 @@ const mutation = graphql`
   mutation createNewFormDataMutation($input: CreateNewFormDataInput!) {
     createNewFormData(input: $input) {
       formData {
+        id
```
AntBush, line 10: Do we still need this change now that we're not creating a new form data?

AntBush (approved):
Works well, just a small but inconsequential nit

app/schema/mutations/application/createNewFormData.ts
marcellmueller: No I'll remove it thanks!

PR #2375, opened 2023-10-06
feat: get cbc projects spreadsheet columns by key name

marcellmueller:
Implements # \<issue number\>


marcellmueller:
@rafasdc this is that alternate solution to getting the data from the excel. Though I don't really like it, it seems very breakable if any column titles change. Let me know if you have any other ideas.

Issue #2376, opened 2023-10-06
Track History for Internal Project Description


MarsRomer:
As a communications director
I want to know if the internal project description has changed since the last update
So that I can announce projects accurately

Description:
The current implementation of the internal project description does not keep a history of changes in the DB.  This ticket is to implement the solution differently so that history is kept and can be reported on later on.

AC:
- [x] In the database the history of changes to the internal project description is kept.  Old ones are soft deleted (archived) while new ones are not.

MarsRomer:
@marcellmueller  would you like to review the ticket details I just added and if it looks ok please estimate this ticket?

MarsRomer:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2376?planning-poker) with Zenhub @marcellmueller

marcellmueller:
Looks good, I put it as a small 3! Will get it done in a few hours thanks.

PR #2379, opened 2023-10-06
2376 project description table

marcellmueller:
PR to save `internal_description` in its own table instead of the `application` table so we can better track history of changes.

Implements #2376


AntBush (approved):
🥇

PR #2381, opened 2023-10-06
feat: analysts can't edit form data when withdrawn

AntBush:
Implements # 926

Forgotten criteria for 926


marcellmueller (approved):
This is working well! 

I'll leave it up to you whether you think this edge case is worth it but I wonder if it's worth adding `disabled={isWithdrawn}` to `<FormBase>` in `edit/[section].tsx` page since they could still edit it by going to the url.

PR #2398, opened 2023-10-12
feat: template import uses title to find data

AntBush:
Implements # 1379

Uses the title next to the column rather than the absolute position. Only worry is if that title ever changes

### DIFF
--- a/app/backend/lib/excel_import/template_one.ts
+++ b/app/backend/lib/excel_import/template_one.ts
@@ -1,8 +1,12 @@
 import XLSX, { WorkBook } from 'xlsx';
 
 const FINAL_ELIGIBLE_HOUSEHOLDS_COLUMN = 'F';
+const FINAL_ELIGIBLE_HOUSEHOLDS_COLUMN_TITLE =
+  'Final number of Eligible Households targeted by this proposal';
 const TITLE_COLUMN = 'C';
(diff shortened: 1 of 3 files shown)


marcellmueller:
Tested this and it doesn't fill the `benefits` page with Template 1 that I grabbed from the tests. However it does work with Template 1  that Mars had linked. I opened both of them and can't see a difference so I'm not sure what is going on?

AntBush:
> Tested this and it doesn't fill the `benefits` page with Template 1 that I grabbed from the tests. However it does work with Template 1 that Mars had linked. I opened both of them and can't see a difference so I'm not sure what is going on?

Turns out, the excel file that Mars has is *slightly* different, in that the excel file has word lowercase but the one I got from the website was capitalized. I'm going to do a `to_lower` and that should solve this problem

marcellmueller:
> > Tested this and it doesn't fill the `benefits` page with Template 1 that I grabbed from the tests. However it does work with Template 1 that Mars had linked. I opened both of them and can't see a difference so I'm not sure what is going on?
> 
> Turns out, the excel file that Mars has is _slightly_ different, in that the excel file has word lowercase but the one I got from the website was capitalized. I'm going to do a `to_lower` and that should solve this problem

Oh crazy, these excel uploads are useful but it's hard not to make them feel like a house of cards 💀

marcellmueller (approved):

Issue #2241, opened 2023-09-12
Milestones history


marcellmueller:
Created, updated or deleted Milestones should display on the history page

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdFNlIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--04c911750ea60eb2654d219402b296577c4eec14/image.png)
Icon is https://fontawesome.com/icons/signs-post?f=classic&s=solid colour: 1A5A96

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2241?planning-poker) with Zenhub @rafasdc @AntBush

PR #2421, opened 2023-10-16
2241 milestones history

rafasdc:
Implements #2241


marcellmueller (commented):
<HistoryIcon> needs to be updated with this icon:
https://fontawesome.com/icons/signs-post?f=classic&s=solid

Also if I edit a milestone it is still showing up as `created`. Not sure how that is being updated here. I handled this in the claims one by setting `history_operation` to `updated` if there was previous data in `create_application_claims_data`.

rafasdc:
> `<HistoryIcon>` needs to be updated with this icon: https://fontawesome.com/icons/signs-post?f=classic&s=solid
> 
> Also if I edit a milestone it is still showing up as `created`. Not sure why it's not working using the `previousHistoryItem`, I handled this in the claims one by setting `history_operation` to `updated` if there was previous data in `create_application_claims_data`.
> 
> These two should be the same milestone created then updated: Screenshot 2023-10-17 at 3 13 15 PM

Thanks for this, no clue how I missed the icon. 

On the other topic, I totally misunderstood the use of the `excel_data_id` assuming it was a constant way of keeping the milestones together, looks I will have to keep them together via quarter to be able to get the previous history item.

Issue #2259, opened 2023-09-14
Assessment assignment table


elliottSlee:
As an analyst 
I want to see project information & assign analysts to particular assessments in the portal
So that I can see how many everyone has and 

**Question: How does this behave if everyone's on the page at the same time? Does it auto-refresh?**

**Description:**
Trying to recreate the project assignment spreadsheet.

# Acceptance Criteria:

[wireframe 1]

- [ ] Acceptance Criteria 

[wireframe 2]

- [ ] Acceptance Criteria

elliottSlee:
This card will likely need to be broken into multiple such as the table, editing assessment columns, filtering, toggle to show more.

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2259?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc

elliottSlee:
Please note that the styling of the table in the designs is arbitrary. It's best to not do any custom styling for first iteration. We can suggest styling changes after implementation.

AayushAnand007:
design updated.

marcellmueller:
@MarsRomer where was the target date column data supposed to come from?

MarsRomer:
lets go with the tech assessment target date for now

MarsRomer:
Looks good, I see that you went above and beyond by including filtering and sorting which is awesome, I will create follow up tickets to put the finishing touches on it

PR #2428, opened 2023-10-17
2259 assessment assignment table

marcellmueller:
Implements #2259 

I was able to pull and filter all the data using GraphQL, let me know if you think there is a good reason to use a computed column instead.


rafasdc (approved):
This is awesome!

PR #2430, opened 2023-10-17
2142-alternate-method-form-data-entry

AntBush:
Implements # 2142

To test, I'd recommend creating an internal intake using the `ccbc_public.create_hidden_intake(now(), now + interval '1 day');`. Copy the UUID and enter it in the dashboard as such `localhost:3000/applicantportal/dashboard?code=<code>` and you will see the create intake open. Then you can create and edit an internal application. Revisiting the dashboard you'll see that the application is linked to intake number 99.


marcellmueller (approved):
Nice work! 🥇

rafasdc (approved):
Double approval! 🍾

PR #2431, opened 2023-10-17
Update sp auto import cron

rafasdc:
Implements # \<issue number\>


marcellmueller (approved):
I thought we were also adding a value to disable this in prod for now?

PR #2444, opened 2023-10-18
fix: no longer use old id to archive old claim excel data

AntBush:
Implements #  2434

To replicate the error:   
1. save a claim
2. edit the claim and replace the file and save
3. delete the claim
4. create a new claim and attempt to upload a claim with the same claim number and it will return the above error. 

This does not fix the issue with already erroneous data, but will prevent the issue from occurring further


marcellmueller (approved):
Nice! 🚀

PR #2447, opened 2023-10-18
fix/claims history

marcellmueller:
- fix: deleted claims history details
- fix: claims file updated diff name
- feat: add old/new labels to history file component


rafasdc (approved):
😄

PR #2448, opened 2023-10-18
chore: set dev cronsp to once a year

rafasdc:
Sets dev schedule to only run once every year.

Implements # \<issue number\>


marcellmueller (approved):
🥳

Issue #2293, opened 2023-09-18
Bug: Other funding source does not update when 'No' is selected


MarsRomer:
**Describe the bug**
When analyst OR applicant is editing the application, if they change the 'Will you have other funding Sources?' radio button from Yes to No, the data that was entered previously in the form when yes was selected persists.

**To Reproduce**
1. Go to any application where 'Will you have other funding Sources?' radio button = Yes and there is data in the form below it 
2. Click the edit pencil
3. Select the 'No' radio button and see the form disappear
4. In metabase the data persists

**Expected behaviour**
- [x] When user applicant selects 'No' then all the conditional form data should be deleted so that those numbers don't show up in subsequent reporting
- [x] When analyst clicks edit on the application form and they select 'No' then all the conditional form data should be deleted so that those numbers don't show up in subsequent reporting

**Screenshots**
Currently this is selected as No
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaW12IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e1bf8175999068892172ba9ed607dde82f89a21e/image.png)
But it shows up in metabase
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBamV2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--759c6393f1d8bd1a63f5c6854bc7d49875c1d456/image.png)

and when I click yes, the data is still there in the form too
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBam12IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--cecbd9ffc102d8ea4c3a9616f52c322dd4342a09/image.png)

**Logs/tracebacks**

**Additional context**

PR #2466, opened 2023-10-19
fix: remove other funding sources array data when 'no' is selected

marcellmueller:
#2293

### DIFF
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -287,6 +287,7 @@ const ApplicationForm: React.FC<Props> = ({
   const isSubmitted = status === 'submitted';
   const isSubmitPage = sectionName === 'submission';
   const isAcknowledgementPage = sectionName === 'acknowledgements';
+  const isOtherFundingSourcesPage = sectionName === 'otherFundingSources';
 
   const isSubmitEnabled = useMemo(() => {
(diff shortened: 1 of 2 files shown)


AntBush (commented):
I think there's a bug when using the `mergeFormSectionData`, the otherFundingSources are still added in again after and is still being sent in the request. Should probably do the `delete` on the `newFormData` there. 
Do something like the following:
```javascript
    if (isOtherFundingSourcesPage && !newFormData.otherFundingSources.otherFundingSources) {
      delete newFormData.otherFundingSources.otherFundingSourcesArray;
      console.log(newSectionData.otherFundingSources);
      console.log(newSectionData.otherFundingSourcesArray);
    }
```

marcellmueller:
> I think there's a bug when using the `mergeFormSectionData`, the otherFundingSources are still added in again after and is still being sent in the request. Should probably do the `delete` on the `newFormData` there. Do something like the following:
> 
> ```js
>     if (isOtherFundingSourcesPage && !newFormData.otherFundingSources.otherFundingSources) {
>       delete newFormData.otherFundingSources.otherFundingSourcesArray;
>     }
> ```

Oh wow nice catch, I was just checking the `newFormSectionData`. I updated it and verified it's working in db for applicant and analyst. Thanks so much 🙇‍♂️

AntBush (approved):

Issue #2457, opened 2023-10-19
Applicant RFI uploads will auto-update benefits and budget info in db


MarsRomer:
User story:
As an analyst
I want the budget details and the benefits to update when and applicant upload new Template 1 and 2 for an RFI
So that I don't have to manually change the numbers in the application

Description:
If applicant uploads template 1 or 2 then the budget details and hh counts that are currently being extracted into the databaes from the applicant portal will upload here to.

AC:

- [x] When applicant uploads Template 1 or 2 for RFI (files that were emailed to NWBC) the associated budget or HH counts numbers are updated same way as #1379
- [x]  Autogenerate Reason for change "Auto updated from upload for RFI: [RFI number]"

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2457?planning-poker) with Zenhub @AntBush @rafasdc

rafasdc:
@MarsRomer are we ok to change this "Auto updated from upload of [template file name] for RFI: [RFI number]" to "Auto updated from upload for RFI: [RFI number]", simplifies the logic since we only have one entry for reason for change when doing both at the same time, as in both template 1 and 2 were requested as part of the RFI.

MarsRomer:
I think it's ok to reword as you suggest.  I was initially thinking it would be good to see the file name in the history, but I don't think it's necessary

PR #2499, opened 2023-10-24
2457 applicant rfi uploads will auto update benefits and budget info in db

rafasdc:
Implements #2457


marcellmueller (approved):
This is working well, nice work!!

PR #2509, opened 2023-10-26
feat: update command to use oc

AntBush:
Implements # 2295


rafasdc (commented):
Could you also add the command to only extract the schema with no data?

marcellmueller (approved):
Thanks so much for the updates!

PR #2517, opened 2023-10-26
feat: use feature flag for internal intake

AntBush:
Implements # 2495

Had some trouble with the test I added, if the test is added at the bottom of the file the test was erroring out for me. The wrong return value was being given in the `useFeature` spy.


marcellmueller (approved):
🥇

PR #2520, opened 2023-10-26
fix: add custom sorting for assessment table analysts

marcellmueller:
Fixes for a few sorting and filtering issues. 

- The `CCBC ID` column filtering will not return any value that contains an `01` anymore if you type say `ccbc-01`, it will return all values that start with `ccbc-01`.
- All four assessment columns sorting and filtering is fixed. This was due to the value being an object since it has the custom dropdown for editing so by default the sorting/filtering was broken as it didn't know which values to look for.


AntBush (approved):
Working well! Good stuff

PR #2522, opened 2023-10-26
feat: change login based on feature flag

rafasdc:
Would like to try two things, feature flagging them so we can try them as needed and without code changes. 

1. Use the custom login route or use the sso default one. SSO middleware changes included checking for previous session and clearing them if stale, want to try not clearing the cookie on first login  as currently done by the custom login route. 
2. Use direct IDIR login for the top-right login button to one-click login for gov machines inside the gov network, would like to test this with the custom login and without to check the behaviour on the double login. 

Implements # \<issue number\>


AntBush (approved):
🤞

marcellmueller (approved):
🥇

PR #2528, opened 2023-10-27
feat: only validate ccbc id for community report

AntBush:
Implements # 2417


app/backend/lib/excel_import/community_progress_report.ts
```diff
@@ -165,7 +165,7 @@ const ValidateData = async (data, applicationRowId, req) => {
 
   data.communityData.forEach((comData) => {
     if (
-      comData.communityName === undefined ||
```
marcellmueller, line 168: @AntBush This PR looks good to me, though I thought we were removing these validations all together other than the CCBC ID?

marcellmueller (approved):

app/backend/lib/excel_import/community_progress_report.ts
AntBush: I assumed we still wanted type validations (and the validation for the summary at the top). This right now is checking that if it's not empty then the type is correct. Let me check in with Mars though

AntBush: Checked in Mars, it's all good 👍

PR #2532, opened 2023-10-27
fix: do not allow applicant to click create application multiple times

AntBush:
Implements fix for being able to create multiple applications while on the applicant dashboard.

The offending area lies in the fact that the [page] is not cached (since it is dynamic), and so if there is a slow connection to the pod that is serving this chunk/page the application will allow you hit the Create Application multiple times. 

The fix here is that, if an application has been created then the button is disabled and a user cannot hit the button multiple times. The problem is that there is nothing else to communicate to the user that something is loading until the browser attempts to load the page/chunk.

### DIFF
--- a/app/pages/applicantportal/dashboard.tsx
+++ b/app/pages/applicantportal/dashboard.tsx
@@ -77,6 +77,7 @@ const Dashboard = ({
 
   const closeTimestamp = openIntake?.closeTimestamp;
   const isInternalIntakeEnabled = useFeature('internal_intake').value ?? false;
+  const [isApplicationCreated, setIsApplicationCreated] = React.useState(false);
 
   const sub: string = session?.sub;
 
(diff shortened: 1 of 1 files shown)


marcellmueller (approved):
Working well nice work!!

app/pages/applicantportal/dashboard.tsx
```diff
@@ -77,14 +77,23 @@ const Dashboard = ({
 
   const closeTimestamp = openIntake?.closeTimestamp;
   const isInternalIntakeEnabled = useFeature('internal_intake').value ?? false;
+  const [isApplicationCreated, setIsApplicationCreated] = React.useState(false);
```
marcellmueller, line 80: Total, completely insignificant nit but we don't use the pattern `React.useState` anywhere that I am aware of 😅 Not a huge deal though.

AntBush: Good point, for some reason I remembered that being used. Updated

PR #2534, opened 2023-10-27
feat: remove hh fields from applicant

AntBush:
Implements # 2456


rafasdc:
Hey, it looks like this is also removing it for the analyst, generating some empty rows, when inspecting them they are the HH fields:

image

They are also fully gone from the edit side for applicants.

Not sure if that was intended?

AntBush:
> Hey, it looks like this is also removing it for the analyst, generating some empty rows, when inspecting them they are the HH fields:
> 
> image
> They are also fully gone from the edit side for applicants.
> 
> Not sure if that was intended?

Nope definitely unintended, I went ahead and added it back for the edit and review page

rafasdc (approved):
👍

Issue #2494, opened 2023-10-24
Styling of Assessment table


MarsRomer:
- [x] Remove border around drop downs
<img width="294" alt="Screenshot 2023-10-24 at 10 33 56 AM" src="https://github.com/bcgov/CONN-CCBC-portal/assets/93939894/141a3dad-710c-4755-891b-63a40630492f">
 
- [x] Fix font - change to BC Sans
- [x] make each CCBC ID color: #1A5A96
- [x] hover state for CCBC ID hyperlink is color: blue <s>and underlined</S>
- [x] Fix permitting assessment filter alignment
- [x] After clicking on the hyperlink & on the assessments page, when "Dashboard" is clicked in the sidebar, it brings the user back to the assessment table, the filtering & sorting that were applied before they navigated away remain
- [x] Remove words 'Assessment'  from headers <S> and narrow columns to show first 9 characters</S>

MarsRomer:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2494?planning-poker) with Zenhub @marcellmueller

MarsRomer:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2494?planning-poker) with Zenhub @AntBush

MarsRomer:
@marcellmueller  can you please see if this AC is working for you:  After clicking on the hyperlink & on the assessments page, when "Dashboard" is clicked in the sidebar, it brings the user back to the assessment table, the filtering & sorting that were applied before they navigated away remain

marcellmueller:
> @marcellmueller can you please see if this AC is working for you: After clicking on the hyperlink & on the assessments page, when "Dashboard" is clicked in the sidebar, it brings the user back to the assessment table, the filtering & sorting that were applied before they navigated away remain


Just tried it in test and sorting and filtering is saved for me when I click the link and return, or refresh the page.

<img width="1422" alt="Screenshot 2023-10-31 at 9 42 26 AM" src="https://github.com/bcgov/CONN-CCBC-portal/assets/14259474/80cfb7ce-9326-492d-80cd-26124aaf991d">

MarsRomer:
[...]

PR #2536, opened 2023-10-27
2494 styling of assessment table

marcellmueller:
Implements #2494

Saving table state based off the Material React Table docs example:
https://www.material-react-table.com/docs/guides/state-management


rafasdc (approved):
This is working great and looks pretty slick! 👍

Issue #2157, opened 2023-08-25
Yaml Config - Scaling


MarsRomer:
Description: Discuss Yaml config  with the platform services team regarding scaling (see note below)

AC: 
- [x] Yaml config is reviewed with someone from Platform Services team
- [x] Recommended changes are made to our Yaml config

From OCIO:
Thank you again for the demo.  That was fun for me.

As an action item, I mentioned having a look at the OpenShift configurations for your application.  I poked around the dashboards in OpenShift for your main namespace, “ff61fb-prod”, “CONN CCBC Portal”, and saw a configuration you should bring forwards that should be adjusted.

What’s the chart below telling us?  It’s basically saying that your application is only using 6.5 percent of the CPU that’s ‘guaranteed’ (known as requested), and the remaining 76.4% it’s ‘borrowing’ (known as limit) from the node the pod is sitting on when it needs to burst for resources.  The ratios here appear to be off significantly, and you will get much better performance if the ratio of requested is higher to that which is limit is requesting when bursting.  

In plain language, it would be better to use more of the CPU you have been dedicated in your quota, than borrowing from the ‘store’ when your app needs more CPU.

I’ve seen this many times.  Please reach out to Jason Leach with Olena Mitovska’s team for a review.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcWREIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--ec0b1e49224c3684c4efad29e4d48d7f5d10b8d9/image.png)

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2157?planning-poker) with Zenhub @marcellmueller @rafasdc @VB-at-Bis

rafasdc:
From Jason: 
That sounds ok to me. You should be keeping request.cpu very low, close to what is actually being used but not lower than about 20m. Then, use limit.cpu to give it some more CPU for performance.

PR #2556, opened 2023-10-30
chore: update resources and hpa

rafasdc:
Implements #2157 

Note: also updating the HPAs so that the app is not always scaling as soon as one user starts using the app as it based on the `requests` of the cpu. Set to 200% on prod and 400% on dev/test, so it will auto scale when CPU usage gets around 40m on all environments.


marcellmueller (approved):
🥇

PR #2557, opened 2023-10-30
2535 add column to cbc data upload

marcellmueller:
- feat: add change request pending column to cbc data import
- test: update cbc_project excel import tests for new column


rafasdc (approved):
😄 👍 🍾

Issue #2455, opened 2023-10-19
Remove Budget Details Page for Applicant


MarsRomer:
As an applicant
I don't want to enter in the budget details on the form
because I already put them in the template

Description:
This ticket is to remove the Budget Details page for the Applicant, but shows on the Analyst review page where they can edit it if needed.

AC
- [ ] Budget Details page does not show for applicant
- [ ] Budget Details page does not show in the stepper
- [ ] Budget Details page does not show in the summary page for the APPLICANT
- [ ] Budget Details page does show for ANALYST in the Application page
- [ ] Budget Details page is editable for ANALYST like other applications on the Application page

MarsRomer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCBC---Delivery-Board-61c0c1204c8bcb001491c5f3/issues/bcgov/CONN-CCBC-portal/2455?planning-poker) with Zenhub @AntBush @marcellmueller @rafasdc

PR #2562, opened 2023-10-30
2455 remove budget page for applicant

marcellmueller:
#2455


rafasdc (commented):
This is throwing a 500 error on edit pages/analyst/application/[applicationId]/edit/[section].tsx. I think its still trying to use the latest schema on that page. 

I think you need to do the same thing you did on the application page to force it to add budgetDetails to the schema.

marcellmueller:
> This is throwing a 500 error on edit pages/analyst/application/[applicationId]/edit/[section].tsx. I think its still trying to use the latest schema on that page.
> 
> I think you need to do the same thing you did on the application page to force it to add budgetDetails to the schema.

Great catch thanks! This should be fixed now.

rafasdc (approved):

PR #2579, opened 2023-10-31
feat: metabase link is not visible while in edit mode

AntBush:
Implements # 2328 comment for not showing metabase link while in edit mode


marcellmueller (approved):
Good to know, I will update my claims metabase link ticket as well

Issue #2613, opened 2023-11-08
Bug: Fix Typo on withdrawl modal


MarsRomer:
says "Connection Communities BC", change to "Connected Communities BC"

PR #2655, opened 2023-11-23
fix: change withdrawal modal description

RRanath:
Implements #2613 

Change withdrawal modal description to fix typo `Connection` to `Connected`


rafasdc (approved):
👍

Issue #2634, opened 2023-11-17
make Zones single selection


MarsRomer:
Make zone selection allowable for only one zone per application.  Check boxes turn into radio button?

AC:
- [x] User can only select one zone, selecting a different zone will unselect the previously selected zone.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdzRIQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1752d444d286f25f606bc289ab54d1a3a77a7307/image.png)

PR #2663, opened 2023-11-28
chore: allow zone to be a single selection

RRanath:
Implements #2634 

Updated the checkboxesWidget to accept a flag/option - `singleSelection`
once allowed will only allow a single selection at a time
Added unit tests


rafasdc:
This is working great, I think we just need to discuss if we want it on the analyst side for any applications that might have been submitted with more than one region.

rafasdc (approved):

PR #2687, opened 2023-12-01
feat: show star for latest received rfi by field

rafasdc:
Implements: [NDT-2]

[NDT-2]: https://connectivitydivision.atlassian.net/browse/NDT-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/Review/widgets/FileWidget.tsx
+++ b/app/components/Review/widgets/FileWidget.tsx
@@ -62,11 +62,15 @@ const FileWidget: React.FC<WidgetProps> = ({
         const isFileField =
           rfiFiles && Object.keys(rfiFiles).includes(fieldName);
         if (!isFileField) return null;
+        const fileFields = rfiFiles[fieldName] || [];
+        // Sorting the file fields by id, latest file will always have higher id
+        // not sorting by uploadedAt as it is not available for all files
+        const sortedFileFields = fileFields.slice().sort((a, b) => b.id - a.id);
         const attachments = rfi?.attachments?.nodes;
 
         return (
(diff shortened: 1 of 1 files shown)


app/components/Review/widgets/FileWidget.tsx
```diff
@@ -53,24 +53,29 @@ const FileWidget: React.FC<WidgetProps> = ({
...
         const attachments = rfi?.attachments?.nodes;
 
         return (
           <StyledFile key={rfi.id}>
-            {rfiFiles[fieldName]?.map((el, index) => {
+            {sortedFileFields?.map((el, index) => {
               // loop through the list of files for the current field
               const isSingleFile = !options?.allowMultipleFiles;
-              const isDisplayIcon = isSingleFile && i === 0;
+              const isDisplayIcon =
```
RRanath, line 76: just wondering why we push to `filesWithIcon` array, instead can't we use  `const isDisplayIcon = isSingleFile && i === 0;` ?

rafasdc: because the same file might there for multiple RFIs, so the intention is to track which types of files we have added the star to, otherwise it will duplicate stars by RFI, this is what happens: 

image

RRanath: Thanks Rafael, Got it, was thinking of rfi categories, not accounting multiple raised rfis.

RRanath (approved):

PR #2707, opened 2023-12-13
fix: handle error when attempting to access growthbook

AntBush:
Implements NDT-8

It seems the bug/error that occured with the upload of the SOW may have been related to this bug that has been occuring on and off. 

Tested the build locally, worked fine. If using this locally with asdf remember to use "asdf install" to install the updated version of node.


rafasdc (approved):
👍

PR #2720, opened 2023-12-15
fix: change hotfix ci to use certs

AntBush:
Implements #NDT-29

Tested the recreate-hotfix workflow and it made the expected change: https://github.com/bcgov/CONN-CCBC-portal/tree/hotfix


rafasdc (approved):
This is fantastic, love the instructions. 👍 🥇

PR #2740, opened 2023-12-15
chore: making a spatial data upload a multiple upload

RRanath:
Implements [#NDT-51](https://connectivitydivision.atlassian.net/browse/NDT-51)

✅On Analyst portal, the kmz upload for signed agreement is now a multifile upload, and should operate the same way as other multifile uploads in the portal
image
image
image

✅Update wording for Map upload as per screen design

✅Update wording in history to reflect wording change for the map upload
image

✅update map upload in change request to a multiple upload as well
image


rafasdc (approved):
👍  great job!

PR #2741, opened 2023-12-15
feat: refresh pods nightly

rafasdc:
Implements [NDT-44]

[NDT-44]: https://connectivitydivision.atlassian.net/browse/NDT-44?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helm/app/templates/cronJobs/restart-ccbc.yaml
+++ b/helm/app/templates/cronJobs/restart-ccbc.yaml
@@ -4,6 +4,7 @@ metadata:
   name: restart-ccbc
 spec:
   schedule: "0 2 * * *" # Run at 2 AM every day
+  timeZone: America/Vancouver
   jobTemplate:
     spec:
       template:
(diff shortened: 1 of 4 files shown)


helm/app/templates/cronJobs/restart-ccbc.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: restart-ccbc
+spec:
+  schedule: "0 2 * * *" # Run at 2 AM every day
```
AntBush, line 6: I was under the impression the cron timezones were set to UTC - looking at the other cronjob we've got setup and checking the time it says it was last run I think we'll want to change this. 2 AM PST is 10 AM UTC according to google

rafasdc: Yes, totally slipped, it will run according to UTC. There seems to be a new timeZone parameter added in 4.12+, let's attempt and use that and check if it runs at 2am PST. If not then we can just adjust like the other ones.

https://docs.openshift.com/container-platform/4.12/rest_api/workloads_apis/cronjob-batch-v1.html

AntBush (approved):

PR #2763, opened 2023-12-21
feat: disable/enable submit based on project area data

rafasdc:
disable or enables the submit button on the form if the project area is fn led and has any zone, or if not fn led then it must be based on the zones selected on growthbook

Implements [NDT-63]

[NDT-63]: https://connectivitydivision.atlassian.net/browse/NDT-63?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RRanath (approved):
👍

PR #2776, opened 2023-12-28
chore: update dev data to create relevant intake

AntBush:
Purely out of convenience whenever I go about deploying dev data. If the other devs find it's not useful feel free to close the PR


RRanath (approved):
👍

PR #2777, opened 2023-12-28
fix: use filteredUiSchema to give correct lastEditedPage

AntBush:
Implements fix for NDT-89 and also NDT-91, the crux of the issue is that the current updateApplicationForm query doesn't trigger the out of sync error if the user is on the same page. The change I made here is to consistently return the correct page, thus reducing the chance of the out of sync error happening erroneously. 

I explored other options, one that I believe would solve this issue without needing to use a sync-check on the database is to use the subscription feature of graphQL and adding a subscription to the formData on this page. The work required for that though, I believe, is of far greater scope than what I believe this ticket intended so I went with this approach.

### DIFF
--- a/app/components/Dashboard/Row.tsx
+++ b/app/components/Dashboard/Row.tsx
@@ -1,6 +1,8 @@
 import Link from 'next/link';
 import styled from 'styled-components';
 import statusStyles from 'data/statusStyles';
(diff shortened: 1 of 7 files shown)


RRanath (approved):
Approved 👍

PR #2788, opened 2023-12-29
fix: submit and review page now consistent with each other

AntBush:
Implements NDT-97 along with another fix for the initial state of the FN-led question and the submission text at the end.


RRanath (approved):
looks good 👍

PR #2798, opened 2024-01-03
fix: save and continuie button is disabled while a save operation is in flight

AntBush:
Implements # \<issue number\>


RRanath (approved):
Approved 👍

PR #2802, opened 2024-01-04
feat: make geographic area required for submission

RRanath:
Implements [NDT-100](https://connectivitydivision.atlassian.net/browse/NDT-100)

image
image

[NDT-100]: https://connectivitydivision.atlassian.net/browse/NDT-100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/Form/ApplicationForm.tsx
+++ b/app/components/Form/ApplicationForm.tsx
@@ -421,51 +421,44 @@ const ApplicationForm: React.FC<Props> = ({
     }
     if (isProjectAreaPage) {
       const firstNationsLed = newFormSectionData?.firstNationsLed || false;
-      let projectAreaAccepted =
-        firstNationsLed ||
-        acceptedProjectAreasArray.includes(
-          newFormSectionData?.geographicArea?.[0]?.toString()
-        );
-
+      const isGeographicAreaEmpty =
+        newFormSectionData?.geographicArea?.[0] === 'undefined' ||
+        newFormSectionData?.geographicArea?.length === 0;
+      const projectAreaAccepted =
+        !isGeographicAreaEmpty &&
+        (firstNationsLed ||
+          acceptedProjectAreasArray.includes(
+            newFormSectionData?.geographicArea?.[0]?.toString()
+          ));
       const geographicAreaInputChanged =
         typeof newFormSectionData?.geographicArea?.[0] !== 'undefined' &&
         newFormSectionData?.geographicArea[0] !==
           jsonData.projectArea?.geographicArea?.[0];
       const firstNationsLedInputChanged =
         firstNationsLed !== jsonData.projectArea?.firstNationsLed;
-      const isGeographicAreaEmpty =
-        newFormSectionData?.geographicArea?.[0] === 'undefined' ||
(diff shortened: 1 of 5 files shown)


app/components/Form/ApplicationForm.tsx
```diff
@@ -442,13 +453,19 @@ const ApplicationForm: React.FC<Props> = ({
...
       }
-
-      setProjectAreaOpen(!projectAreaAccepted);
-
       setProjectAreaModalOpen(
         !projectAreaAccepted &&
           (geographicAreaInputChanged || firstNationsLedInputChanged)
       );
+      projectAreaAccepted =
+        firstNationsLed ||
```
AntBush, line 461: Why are we re-assigning this here? I'm a bit confused as to the purpose here. If this is due to the case of it already being submitted, I'd put this in the `if` statement above. I also don't think we need to separate `if (isSubmitted && (isGeographicAreaEmpty || !isProjectAreaAccepted))` and `if (isSubmitted && isProjectAreaAccepted)` since isProjectAreaAccepted should be false if `geographicArea` is either undefined or empty.

Additionally, I know this isn't part of the ticket and will be put for a new ticket, but the logic here is getting a bit much and could use a refactor/some comments.

app/components/Form/ApplicationForm.tsx
```diff
@@ -442,13 +453,19 @@ const ApplicationForm: React.FC<Props> = ({
       } else if (!isSubmitEnabled && !projectAreaAccepted) {
```
AntBush, line 453: Additionally, I know this was probably my mistake but just a logic error I saw here, I think this `else if` here should be `!isSubmitted` and not `!isSubmitEnabled` since that will also trigger for orgName not being filled out, etc.

We can fix it here/now or we can just open it in another ticket.

RRanath: yes, we can handle that change here.

app/components/Form/ApplicationForm.tsx
RRanath: Yes I am reassigning to avoid erroneous validation errors showing(eg: submit page)
but yes you are right, we can combine `isGeographicAreaEmpty || !isProjectAreaAccepted together.

I am not sure if we can add the recalculation logic inside if statement, cause we still need to handle modal logic after all the changes, but anyways extracted to another variable since we are changing isProjectAreaAccepted logic with new changes and for clarity as well.

AntBush: I meant moreso because we're doing this calculation for the portion surrounding the submitted portion, that re-check should be in that if statement surrounding the submitted check and encapsulating that logic in one area.

RRanath: yes that is correct, but if we do so, won't we loose the correct functionality that we need for `setProjectAreaModalOpen`?
but anyways, with new changes I have kind of separated the `isProjectAreaAccepted` from `isProjectAreaSelected`

AntBush (approved):

rafasdc (approved):
🥇

PR #2893, opened 2024-01-25
feat: use mrt for main analyst dashboard

AntBush:
Implements # \<issue number\>


rafasdc (commented):
I keep getting hydration errors on dashboard as soon as any kind of filters (cookie) gets set. 

I think we need a separate useEffect for each of the cookie.sets as well as a isFirstRender similar to the Assessment Assignment Table,  similarly we might want to set initial state with a [] useEffect to "restore" the saved filters of the cookies instead of setting them at initial state and set first render then.

Please let me know if I'm out to lunch or wanna chat.

rafasdc (approved):
👍

PR #2912, opened 2024-01-29
chore: add a banner to dev and test to indicate environment

RRanath:
Implements [NDT-128](https://connectivitydivision.atlassian.net/browse/NDT-128)


app/components/HeaderBanner.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const publicRuntimeConfig = getConfig()?.publicRuntimeConfig;
+  const namespace = publicRuntimeConfig?.OPENSHIFT_APP_NAMESPACE;
+  const isTest = namespace?.endsWith('-test');
+  return (
+    <>
+      {environmentIndicator && (
+        <StyledBaseHeaderBanner header="sub">
+          <StyledDiv>
+            Connected Communities BC portal{' '}
+            {isTest ? ' Test ' : ' Development '} environment.
```
AntBush, line 59: I think we should have a failsafe here in case this gets accidentally turned on in prod. We don't want any applicants or analysts assuming they're in dev should this be enabled

RRanath: Yeah, better to have another extra layer. 
Fixed

AntBush (approved):
Looks good!

PR #2921, opened 2024-01-30
feat: show internal and external status on the dashboard

AntBush:
Implements # \<issue number\>


RRanath (commented):
Columns look Good, but one thing though, in the ticket it says to add a `feature flag` and to `widen the table/margins` too. (unless we decided not to 🤓)

AntBush:
> Columns look Good, but one thing though, in the ticket it says to add a `feature flag` and to `widen the table/margins` too. (unless we decided not to 🤓)

Ah I completely missed that, thanks! Yeah I'll go ahead that and add it in

RRanath (approved):
Looks Good 🚀

PR #2924, opened 2024-01-31
fix: use 'full' option to include unchanged value

AntBush:
Implements fix for history page, problem occured as the docs doesn't explicitly state that scalar values aren't included in "keepUnchangedValues"


rafasdc:
The update made it create a few extra null test-ids on some tests, needed to update the array index on them to be + 1

PR #2925, opened 2024-01-31
feat: add remaining character count for text area widgets application portal

RRanath:
Implements [NDT-131](https://connectivitydivision.atlassian.net/browse/NDT-131)


AntBush (approved):
🚀

PR #2976, opened 2024-02-21
chore: upgrade rjsf version

RRanath:
Implements [#NDT-138](https://connectivitydivision.atlassian.net/browse/NDT-138)

This tickets upgrades the version to rjsf 5.17.1
main changes
- changing `JSONSchema7` to `RJSFSchema`
- changing imports to fix new utils package migration
- changes associating with `FieldProps`
- using rjsf ajv8 for validation

validation issues 
- arrays are now assigned with a empty array by default, so making validations break. possible to fix with a custom keyword passed to `customizedValidator` or sanitizing data to remove empty arrays from data
`const ignoreEmptyArrays: FuncKeywordDefinition = {
  keyword: 'ignoreEmptyArrays',
  type: 'array',
  schema: false,
  validate: (data) => {
    if (Array.isArray(data) && data.length === 0) {
      return false;
    }
    return true;
  },
  errors: true,
};`

since there is a possibility of a solution from rjsf keeping it simple with `minItems: 1` in schema for now

- still keeping `novalidate` in form, will be replaced by an optional validator property in future


rafasdc (approved):
Huge work, fantastic! 🍾 🎆

PR #2991, opened 2024-02-23
feat: add assessment status colours to dashboard

rafasdc:
Implements [NDT-184]

[NDT-184]: https://connectivitydivision.atlassian.net/browse/NDT-184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/AnalystDashboard/AssessmentLead.tsx
+++ b/app/components/AnalystDashboard/AssessmentLead.tsx
@@ -1,4 +1,5 @@
 import styled from 'styled-components';
+import assessmentPillStyles from 'data/assessmentPillStyles';
 import { useCreateAssessmentMutation } from '../../schema/mutations/assessment/createAssessment';
 
 const StyledDropdown = styled.select`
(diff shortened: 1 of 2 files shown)


app/components/AnalystDashboard/AssessmentLead.tsx
```diff
@@ -50,8 +47,37 @@ const AssignLead: React.FC<Props> = ({
 
   const analystList = Object.keys(allAnalysts);
 
+  let backgroundColor: string;
```
RRanath, line 50: just wondering if we can use the `assessmentPillStyles.ts`?, since they seems to be almost the same and we get to control colors from a single source 🤔

rafasdc: Yup, I'll refer to the file directly. Great catch

RRanath (approved):
Approved

PR #3015, opened 2024-02-29
test: fix assessment assignment table flaky tests

rafasdc:
Tests were being flakey due to some timeouts and the changes to the table (it looks like), removed async from zone to make it more consistent and increase timeout to 5000ms just in case.

Implements # \<issue number\>


RRanath (approved):
Approved 😎

PR #3041, opened 2024-03-05
fix: ches export secrets

rafasdc:
Fixes a typo on the export secrets job for the new CHES secrets that was making it fail the export.


RRanath (approved):
Approved

PR #3057, opened 2024-03-12
chore: hide Lead with feature flag

RRanath:
Implements  [NDT-216](https://connectivitydivision.atlassian.net/browse/NDT-216)

[NDT-216]: https://connectivitydivision.atlassian.net/browse/NDT-216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/AnalystDashboard/AllDashboard.tsx
+++ b/app/components/AnalystDashboard/AllDashboard.tsx
@@ -22,6 +22,7 @@ import Link from 'next/link';
 import ClearFilters from 'components/Table/ClearFilters';
 import type { AllDashboardTable_query$key } from '__generated__/AllDashboardTable_query.graphql';
 import { TableCellProps } from '@mui/material';
+import { useFeature } from '@growthbook/growthbook-react';
 import { filterZones } from './AssessmentAssignmentTable';
 
 type Application = {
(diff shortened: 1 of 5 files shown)


app/components/AnalystDashboard/AllDashboard.tsx
```diff
@@ -185,8 +186,9 @@ const AllDashboardTable: React.FC<Props> = ({ query }) => {
   const [columnFilters, setColumnFilters] = useState<MRT_ColumnFiltersState>(
     []
   );
+  const showLeadFetureFlag = useFeature('show_lead').value ?? false;
   const [columnVisibility, setColumnVisibility] = useState<MRT_VisibilityState>(
-    {}
+    { Lead: showLeadFetureFlag }
```
AntBush, line 191: I think this is causing an error, since we're instantiating the state with the growthbook flag that isn't evaluated until it reaches the client side so there's a mismatch (at least when I set the growthbook flag to false).

It also reverts back to the lead being shown on page load

AntBush (changes requested):
It looks like there's a bug when the feature flag defaults to false

AntBush (approved):
Awesome work!

PR #3064, opened 2024-03-15
feat: add legend to assesmsnt assignment table

RRanath:
Implements [NDT-228](https://connectivitydivision.atlassian.net/browse/NDT-228)

[NDT-228]: https://connectivitydivision.atlassian.net/browse/NDT-228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AntBush (commented):
I think this needs a bit more spacing between the pills, they seem to be quite close together

RRanath:
New spacing: 
image

RRanath:
@AntBush Yes, seems like breaking the line will only lengthen the header.
btw, missed a criteria to keep unassigned as an empty label, so fixed that as well.
since color change is fixed in a [separate ticket ](https://connectivitydivision.atlassian.net/browse/NDT-227)will not be touching that.

PR #3075, opened 2024-03-21
fix: trim analyst names

AntBush:
Implements NDT-259


RRanath (approved):
Approved 🚀

AntBush:
> Approved 🚀

Sorry @RRanath can you give this a second review? I updated it so that the trim only happens on the comparison.

RRanath (approved):
Approved 🚀

PR #3083, opened 2024-03-25
feat: enable real time color change

AntBush:
Implements NDT-227


RRanath:
Looks Good😎 , but in the tickets it says `When user selects Null value the colour should change to white without having to refresh` which I guess we still have to address,

AntBush:
> Looks Good😎 , but in the tickets it says `When user selects Null value the colour should change to white without having to refresh` which I guess we still have to address,

It should be blanking out the option now if you want to try it.

RRanath:
@AntBush it seems to work for some statuses, not for some,
image

AntBush:
@RRanath looks like I wasn't checking for when we were assigning the literal text "Unassigned", after copying over the test data it looks like it's working now.

RRanath (approved):
Approved

PR #3097, opened 2024-04-02
feat: add email notification to second review eligibility screening

RRanath:
Implements [NDT-252](https://connectivitydivision.atlassian.net/browse/NDT-252)

- Keeping emailer IDs in the codebase as discussed, to be removed once we implement a email action page

[NDT-252]: https://connectivitydivision.atlassian.net/browse/NDT-252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/backend/lib/emails/handleEmailNotification.ts
```diff
@@ -0,0 +1,99 @@
...
+    { _rowIds: ids },
+    req
+  );
+  return results?.data?.allAnalysts?.edges.map((edge) => edge.node.email);
+};
+
+const handleEmailNotification = async (
+  req,
+  res,
+  template: (
```
AntBush, line 57: I think making this function an interface here so it can be extended for the template functions with some descriptions of the parameter is a good idea. Also eventInitiator and params should be optional parameters so we don't get type errors.

AntBush (approved):
Working well, just a comment to help increase type coherence/code quality

app/backend/lib/emails/handleEmailNotification.ts
RRanath: Extracted to an `EmailTemplateProvider` and marked optional for eventInitiator and params

AntBush (approved):

PR #3123, opened 2024-04-08
feat: ci/cd merge process

rafasdc:
Implements [NDT-236]

- [ ] Check me to trigger release process.

[NDT-236]: https://connectivitydivision.atlassian.net/browse/NDT-236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/release-process.yaml
```diff
@@ -0,0 +1,127 @@
...
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const checkboxText = "[x] Check to trigger automatic release process";
+            const prNumber = context.payload.pull_request.number;
+            const currentPR = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: prNumber
```
AntBush, line 119: Seeing as we're unchecking after creating the release, what would happen if something else is merged? I'm guessing we'll get a conflict so we'd have to resolve this manually. Maybe add some documentation (in the other PR) on what to do in this case for new folks and so all current devs run it the same way.

rafasdc: The release process will fail as release-it is smart enough to not allow two releases as it checks the existing tags before. Added info on the two exceptions that can occur on the docs PR :)

AntBush (approved):

PR #3149, opened 2024-04-11
feat: intake 4 uses new map

AntBush:
Implements NDT-264 for new map


app/lib/theme/widgets/ZoneMapWidget.tsx
```diff
@@ -40,10 +40,28 @@ const ZoneMapWidget: React.FC<WidgetProps> = ({ formContext }) => {
     ? '/images/zone-map.png'
     : '/images/zone-map-intake-3.png';
 
+  const updatedMapIntakeFour = '/images/zone-map-intake-4.png';
+
+  const intakeNumber = formContext?.intakeNumber;
```
rafasdc, line 45: Should we default to 4 to prevent broken image when intake is neither of them? 
image

AntBush: Good point, made a change now defaults for intake 4

rafasdc (approved):
👍

PR #3230, opened 2024-05-01
chore: update applciation_history_test to use mock time

AntBush:
Uncomments using the mock time which for some reason was commented out, allowing tests to pass


AntBush:
@RRanath @rafasdc Let's merge this in first/soon so we don't hold up any release processes

rafasdc (approved):

PR #3251, opened 2024-05-07
feat: redirect to requested link after login

rafasdc:
Implements [NDT-263]


AntBush (approved):
Great work! Much simpler than I expected

PR #3296, opened 2024-05-22
fix: undefined project labels

rafasdc:
adds a condition to check for the existence of the label or title before turning it into a string, otherwise null is returned so the h4 is not rendered. It was causing unintended "undefined" to be shown as labels on conditional approval

Implements [NDT-]


RRanath (approved):
Looks Good. 🚀

PR #3312, opened 2024-05-27
chore: move change request pending to cbc header

RRanath:
Implements [[NDT-339](https://connectivitydivision.atlassian.net/browse/NDT-339)]

[NDT-339]: https://connectivitydivision.atlassian.net/browse/NDT-339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rafasdc (commented):
A few things: 

- There seems to be a problem with the mutation function create_pending_change_request when deploying to an already deployed DB (test), you'll likely need to drop the current function and then recreate it due to uniqueness. 
- Are we sure we want the cbc and ccbc change requests combined into one application table, I see the value but thing this might get confusing. I was thinking this was going to go into a cbc_pending_changing_request table or similar to keep them separated. PRobably something we should discuss with the wider group for metabase queries too.

rafasdc (approved):
🚀

PR #3369, opened 2024-06-17
fix: use shortened feature name for helm and route

rafasdc:
Implements [NDT-]


RRanath (approved):
Approved

PR #3380, opened 2024-06-24
fix: feature db name to match release name

rafasdc:
Implements [NDT-]

### DIFF
--- a/.github/workflows/deploy_feature.yaml
+++ b/.github/workflows/deploy_feature.yaml
@@ -62,8 +62,8 @@ jobs:
           openshift_token: ${{ secrets.OPENSHIFT_TOKEN }}
       - name: Setup database
         run: |
-          FEATURE_NAME_LOWER=$(echo $FEATURE_NAME | tr '[:upper:]' '[:lower:]')
-          FEATURE_NAME_LOWER_SHORT=$(echo $FEATURE_NAME_LOWER | cut -c -30)
+          FEATURE_NAME_LOWER=$(echo $FEATURE_NAME | tr '[:upper:]' '[:lower:]' | sed 's/-*$//')
+          FEATURE_NAME_LOWER_SHORT=$(echo $FEATURE_NAME_LOWER | cut -c -30 | sed 's/-*$//')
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_feature.yaml
```diff
@@ -62,6 +62,8 @@ jobs:
           openshift_token: ${{ secrets.OPENSHIFT_TOKEN }}
       - name: Setup database
         run: |
+          FEATURE_NAME_LOWER=$(echo $FEATURE_NAME | tr '[:upper:]' '[:lower:]')
+          FEATURE_NAME_LOWER_SHORT=$(echo $FEATURE_NAME_LOWER | cut -c -30)
```
AntBush, line 66: Aren't we missing the `sed 's/-*$//'` command here? 
```suggestion
          FEATURE_NAME_LOWER_SHORT=$(echo $FEATURE_NAME_LOWER | cut -c -30 | sed 's/-*$//')
```

.github/workflows/deploy_feature.yaml
```diff
@@ -62,6 +62,8 @@ jobs:
...
           chmod +x ./lib/feature_envs/create_feature_db.sh
           ./lib/feature_envs/create_feature_db.sh ccbc ${{ env.FEATURE_NAME }} ${{ secrets.OPENSHIFT_APP_NAMESPACE }}
```
AntBush, line 68: I think we're supposed to pass the `FEATURE_NAME_LOWER_SHORT` here right? 

```suggestion
          ./lib/feature_envs/create_feature_db.sh ccbc "$FEATURE_NAME_LOWER_SHORT" ${{ secrets.OPENSHIFT_APP_NAMESPACE }}
```

rafasdc: Yup, fixed!

.github/workflows/deploy_feature.yaml
rafasdc: Yes!

AntBush (approved):

PR #3384, opened 2024-06-26
fix: add cbc_admin role to allowed download of attachments

AntBush:
Implements [NDT-402]


RRanath (approved):
Approved

PR #3397, opened 2024-07-02
feat: generate gcpe report

rafasdc:
Implements [NDT-338]

[NDT-338]: https://connectivitydivision.atlassian.net/browse/NDT-338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/backend/lib/reporting/gcpe.ts
```diff
@@ -0,0 +1,592 @@
...
+      // project complete
+      { value: node?.jsonData?.projectComplete },
+      // phase
+      { value: node?.jsonData?.phase },
+      // project #
+      { value: node?.projectNumber },
+      // ubf/federal project #
+      { value: node?.jsonData?.federalProjectNumber },
+      // applicant
+      { value: node?.jsonData?.applicantContractualName },
```
AntBush, line 193: Just to be sure, we're not using the current operating name here vs the contractual name?

app/components/AnalystDashboard/DashboardTabs.tsx
```diff
@@ -43,6 +43,7 @@ const DashboardTabs = ({ session }) => {
   const isAdmin = authRole === 'ccbc_admin';
   const router = useRouter();
   const showGisUpload = useFeature('show_gis_upload').value;
+  const showReporting = true;
```
AntBush, line 46: I'm guessing this was meant to be converted into a feature flag?

app/backend/lib/reporting/reporting.ts
```diff
@@ -0,0 +1,58 @@
...
+} from './gcpe';
+
+const reporting = Router();
+
+const limiter = RateLimit({
+  windowMs: 1 * 60 * 1000,
+  max: 2000,
+});
+
+reporting.get('/api/reporting/gcpe', limiter, async (req, res) => {
```
AntBush, line 17: I notice for these we're not using the same authentication check we're using for other api routes. Is the expectation here for the DB to deny the request here and let it fail vs a 404 (as we're doing now) or a 403

AntBush (changes requested):
Solid work, but there's a couple things that need some clarification here

app/backend/lib/reporting/reporting.ts
rafasdc: I was not attempting to restrict it and rather refine it later, but yes the call will fail if you don't have proper auth and will result in an error.

AntBush (approved):

PR #3409, opened 2024-07-09
fix: grant service account to cbc change request table

rafasdc:
Fixes the error on the import of cbc data by the cron job by granting access to the cbc_application_pending_change_request table to the ccbc_service_account so the graphql query and mutation can succeed after initial import.

Implements [NDT-423]

[NDT-423]: https://connectivitydivision.atlassian.net/browse/NDT-423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RRanath (approved):
Approved

PR #3411, opened 2024-07-10
fix: onAuthCallBack re-enabled

AntBush:
Implements [NDT-384]

[NDT-384]: https://connectivitydivision.atlassian.net/browse/NDT-384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AntBush:
Just as a note, when reviewing, if you're using a fresh DB you'll need to login 3 times for your user to be created since we put out the mock users first I believe.

rafasdc (approved):

PR #3421, opened 2024-07-15
feat: refine cbc header

rafasdc:
Implements [NDT-386]

[NDT-386]: https://connectivitydivision.atlassian.net/browse/NDT-386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/Analyst/CBC/CbcEditProjectNumber.tsx
+++ b/app/components/Analyst/CBC/CbcEditProjectNumber.tsx
@@ -220,8 +220,6 @@ const CbcEditProjectNumber: React.FC<Props> = ({
       {error && (
         <StyledMessage>
           {error && <StyledError>{error}</StyledError>}
-
-          {/* {help && <Label>{help}</Label>} */}
         </StyledMessage>
       )}
     </StyledDiv>
--- a/app/components/Analyst/CBC/CbcEditProjectNumber.tsx
+++ b/app/components/Analyst/CBC/CbcEditProjectNumber.tsx
@@ -139,7 +139,6 @@ const CbcEditProjectNumber: React.FC<Props> = ({
           'Content-Type': 'application/json',
         },
       });
-      console.log('response', response);
       if (!response.ok) return false;
       const data = await response.json();
       return data;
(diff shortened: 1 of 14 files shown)


app/components/Analyst/CBC/CbcEditProjectNumber.tsx
```diff
@@ -0,0 +1,229 @@
...
+    } catch (e) {
+      setError(
+        'An error occurred while attempting to validate the project number.'
+      );
+      return false;
+    }
+  };
+
+  const validateProjectNumber = async (val) => {
+    setIsValidating(true);
```
RRanath, line 154: I think we need to validate if the project number is <4 as well?

> When editing and on read only display If “Project Number” is not 4 digits integer, validate with red box outline and text.  “Invalid number, please use a 4 digit number”.

db/revert/tables/cbc_add_fk_update_constraint.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert ccbc:tables/cbc_add_fk_update_constraint from pg
+
+BEGIN;
+
+-- XXX Add DDLs here.
```
RRanath, line 5: Do we need the revert script here?

app/components/Analyst/CBC/CbcEditProjectNumber.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const [isSaving, setIsSaving] = useState(false);
+  const [updateProjectNumber] = useUpdateCbcProjectNumberMutation();
+
+  const [projectNumber, setProjectNumber] = useState(
+    queryFragment?.projectNumber
+  );
+
+  const [error, setError] = useState(null);
+
+  const handleSubmit = () => {
```
RRanath, line 85: Noticed that after update the project number in dashboard it still shows the old project number.

app/components/Analyst/CBC/CbcEditProjectNumber.tsx
rafasdc: I thought so too at first, but AC states "Allow user to save the the project number with >=1 digit.", so current validation should be ok.

RRanath (approved):

PR #3422, opened 2024-07-15
chore: update crunchy postgres

rafasdc:
updates crunchy postgres version which includes updates to the cron schedule for backups so s3 and volume do not conflict with each other

Implements [NDT-]


RRanath (approved):
Approved

PR #3530, opened 2024-09-06
feat: cbc history

AntBush:
Implements [NDT-341]

[NDT-341]: https://connectivitydivision.atlassian.net/browse/NDT-341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rafasdc:
This is working great, just a few things: 

1. Need a key for `projectStatus` and `projectDescription` on the history schema so they show in proper case.
2. When editing ER and RD they don't show in the history. I'm thinking this is something we will handle on another ticket? 
3. Same as above for project number. When the project number is changed the column `project_number` on `cbc_data` and `cbc` tables gets updated but there is no change on the json_data, so the change is showing up as empty for this one. Not sure if this something we handle now or as part of a different ticket.

AntBush:
> This is working great, just a few things:
> 
> 1. Need a key for `projectStatus` and `projectDescription` on the history schema so they show in proper case.
> 2. When editing ER and RD they don't show in the history. I'm thinking this is something we will handle on another ticket?
> 3. Same as above for project number. When the project number is changed the column `project_number` on `cbc_data` and `cbc` tables gets updated but there is no change on the json_data, so the change is showing up as empty for this one. Not sure if this something we handle now or as part of a different ticket.

I think 1. and 3. we can make as part of the ticket, but for 2 I'll ask Mars during standup see what he thinks for this ticket

rafasdc (approved):

PR #3542, opened 2024-09-10
feat: ccbc er rd job

rafasdc:
Implements [NDT-360]

[NDT-360]: https://connectivitydivision.atlassian.net/browse/NDT-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cron/shp/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
...
+CREATE TABLE IF NOT EXISTS ccbc_public.ccbc_applications_coverages (
+    id SERIAL PRIMARY KEY
+);
+"
+
+echo "Running shp2pgsql and sql scripts..."
+# Run shp2pgsql for each shapefile
+for dir in /data/economic_regions /data/regional_districts /data/ccbc_applications_coverages; do
+    base_dir=$(basename $dir)
+    for shp in $dir/*.shp; do
```
AntBush, line 35: When trying this locally, the files in ER and RDs are nested so the data doesn't get populated. I added some `mv`s after the zip to get this to run as expected. But doing a replace of this with 
```suggestion
    find $dir -name '*.shp' | while read -r shp; do
```
fixed any of the nested-file issues. I don't know if this is from a difference in environments and if the `unzip` package in ubuntu/debian are different or if it's from a difference in the zips used.

AntBush (approved):

PR #3603, opened 2024-10-04
feat: ccbc summary page communities from template 9

rafasdc:
Implements [NDT-503]

[NDT-503]: https://connectivitydivision.atlassian.net/browse/NDT-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/helpers/ccbcSummaryGenerateFormData.tsx
```diff
@@ -295,13 +335,28 @@ const getFormDataFromApplication = (
...
   return {
     formData: {
       counts: {
-        communities: null,
-        benefitingCommunities: null,
-        indigenousCommunities: null,
-        nonIndigenousCommunities: null,
+        communities:
+          (applicationData?.applicationFormTemplate9DataByApplicationId
+            ?.nodes[0]?.jsonData?.communitiesToBeServed || 0) +
```
AntBush, line 351: If, for whatever reason, there is no communitiesToBeServed do we want it to be 0 or just be null? Small thing we can address later though

AntBush (approved):
Looks good, just a small question for the summary page. Can be addressed in a future ticket if necessary

app/lib/helpers/ccbcSummaryGenerateFormData.tsx
rafasdc: Yeah, I think that is a good point, since it is defaulting to 0 to avoid weird JS summation. I've created [NDT-579] to handle it if needed.

PR #3637, opened 2024-10-24
fix: add application id when archiving

AntBush:
Implements [NDT-571]

[NDT-571]: https://connectivitydivision.atlassian.net/browse/NDT-571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/deploy/mutations/archive_application_sow.sql
```diff
@@ -14,7 +14,7 @@ begin
     sow_data_ids := array(
         select sd.id
         from ccbc_public.application_sow_data as sd
-        where sd.amendment_number = _amendment_number
+        where sd.amendment_number = _amendment_number and sd.application_id = _application_id
     );
 
     -- Update archived_at for non-archived records in application_sow_data table
```
RRanath, line 20: I think we need the same check here as well?

AntBush: Ah yes you're right, let me update that. Thanks!

RRanath (approved):
looks good 🚀

PR #3725, opened 2024-12-04
feat: add all dashboard export

rafasdc:
Implements [NDT-626]

[NDT-626]: https://connectivitydivision.atlassian.net/browse/NDT-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/components/AnalystDashboard/AllDashboard.tsx
```diff
@@ -325,6 +331,67 @@ const AllDashboardTable: React.FC<Props> = ({ query }) => {
...
+          // eslint-disable-next-line no-param-reassign
+          result[program.toLowerCase()] = [];
+        }
+
+        result[program].push(item.rowId);
+        return result;
+      },
+      {} as Record<string, number[]>
+    );
+    await fetch('/api/dashboard/export', {
```
AntBush, line 370: Usually shouldn't mix await with a `.catch`/`.then`. I think either remove the `await` or refactor into a try/catch with multiple awaits for the `blob()` call.

AntBush (approved):
Looks good/works well! Just a small (optional) change for column_options and a small refactor for calling the endpoint for the export

PR #3833, opened 2025-01-30
Ndt 729 further refine ci cd

rafasdc:
Implements [NDT-729]

[NDT-729]: https://connectivitydivision.atlassian.net/browse/NDT-729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/test-db.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+  pgtap:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Run Database Unit Tests
+        run: |
+          docker run --rm --name ccbc-pgtap -p 5432:5432 -d ghcr.io/bcgov/conn-ccbc-portal/ccbc-unit-test-db:latest
```
AntBush, line 67: I'm a bit confused by the intention to use latest - isn't it possible to pick up another image that is built from another PR?

.github/workflows/build.yaml
```diff
@@ -63,6 +67,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build and push Docker image
+        if: ${{ matrix.build_on_main != true || github.ref == 'refs/heads/main' || matrix.name == 'app' }}
```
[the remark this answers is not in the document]
rafasdc: Agreed, will change/add

.github/workflows/test-db.yaml
rafasdc: We don't use any other type of tags aside from the SHA of the commit or latest, since this image will almost never change using latest will ensure it always find an existing image (since it is only built on main). We could decide we want to add a "main" tag to all images built on main though?

AntBush (approved):

PR #3864, opened 2025-02-11
chore: disable sentry in local dev

AntBush:
Implements [NDT-]


rafasdc (approved):

app/sentry.client.config.ts
```diff
@@ -5,29 +5,31 @@
 import * as Sentry from '@sentry/nextjs';
 import getConfig from 'next/config';
 
-const { SENTRY_ENVIRONMENT, SENTRY_RELEASE } =
+const { SENTRY_ENVIRONMENT, SENTRY_RELEASE, OPENSHIFT_APP_NAMESPACE } =
   getConfig().publicRuntimeConfig || {};
 
-Sentry.init({
-  dsn: 'https://d6719b95640e48e28369cd152b9ea9e5@o4504057698320384.ingest.us.sentry.io/4504079285944320',
+if (OPENSHIFT_APP_NAMESPACE && OPENSHIFT_APP_NAMESPACE.includes('ff61fb')) {
```
rafasdc, line 11: Might be worth it to disable it in ff61fb-dev? Or we can wait and see if this change alleviates the storage use

AntBush: Disabling it in dev should be another easy change, but yes it's likely that the dev instance is likely what's also eating it up so I can go ahead and make that change. This issue is mostly with replays, which afaik is not something we use often? I've never bothered to try out a replay

PR #3970, opened 2025-04-17
fix: schema spy

rafasdc:
Implements [NDT-845]

[NDT-845]: https://connectivitydivision.atlassian.net/browse/NDT-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/schemaspy.yaml
+++ b/.github/workflows/schemaspy.yaml
@@ -45,7 +45,7 @@ jobs:
           mkdir output
           chmod a+rwx -R output
       - name: Run Schemaspy
-        run: docker run --network host -v "$PWD/output:/output" schemaspy/schemaspy:latest -t pgsql -db ccbc -host 127.0.0.1 -port 5432 -u postgres -p postgres -schemas ccbc_public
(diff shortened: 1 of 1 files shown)


.github/workflows/schemaspy.yaml
```diff
@@ -1,42 +1,49 @@
...
-        run: make start_pg && make drop_db && make deploy_db_migrations
+        run: |
+          docker run --name ccbc-db -p 5432:5432 -e POSTGRES_PASSWORD=mysecretpassword -d postgres:15
+          # Wait for PostgreSQL to be ready
+          until docker exec ccbc-db pg_isready -U postgres; do
+            echo "Waiting for PostgreSQL to be ready..."
+            sleep 2
+          done
+          # Additional wait to ensure PostgreSQL is fully ready
+          sleep 5
```
osokolovbc, line 28: Isn't it ready by the condition from the prev loop?

rafasdc: Should be, but that was not what was observed from other actions that do the same thing, sometimes it is ready on the first time and it reports as ready and works but other times the connection still fails, even when "ready", so the second check and sleep was needed to ensure it is actually "fully ready", I think this is likely due to the runner being slower at times.

osokolovbc (approved):

PR #3978, opened 2025-04-25
chore: Added project number to project detail panel

osokolovbc:
Implements [NDT-708]


osokolovbc:
Screenshot 2025-04-29 at 2 06 19 PM

rafasdc (changes requested):
I've gone through this, a few things: 

For the project number, you should be targeting the field `originalProjectNumber` not `projectNumber` this will likely not exist on most CBC projects. 

The ticket also has the AC: "When a user searches by the original project number in global search, the project row should expand and highlight the original project number." so you will need add logic to search under the customGlobalFilter function which starts at line 559

app/components/AnalystDashboard/AllDashboardDetailPanel.tsx
```diff
@@ -71,6 +72,8 @@ const AllDashboardDetailPanel: React.FC<Props> = ({ row, filterValue }) => {
       ) : (
         <StyledSpan>N/A</StyledSpan>
       )}
+      <StyledSpan>Original Project Number</StyledSpan>
+      <StyledSpan>{row.originalProjectNumber ?? "N/A"}</StyledSpan>
```
rafasdc, line 76:
```suggestion
      <StyledSpan>{row.original.originalProjectNumber ?? 'N/A'}</StyledSpan>
```

rafasdc: You still neeed the row.original as it is referring to the original value of the row in MRT

app/components/AnalystDashboard/AllDashboard.tsx
```diff
@@ -536,9 +536,16 @@ const AllDashboardTable: React.FC<Props> = ({ query }) => {
       .join(',')
       ?.toLowerCase();
     const detailsMatch = communitiesString?.includes(filterValue.toLowerCase());
-    expandedRowsRef.current[row.id] = detailsMatch;
 
-    return defaultMatch || detailsMatch;
+    // Match for original project number
+    const originalProjectNumber = row.originalProjectNumber?.toLowerCase();
+    const projectNumberMatch = originalProjectNumber?.includes(filterValue.toLowerCase());
```
rafasdc, line 542:
```suggestion
    // Match for original project number
    const originalProjectNumber = row?.original?.originalProjectNumber;
    // Ensure both are numbers for comparison, or fallback to string includes if not
    let projectNumberMatch = false;
    if (
      originalProjectNumber !== undefined &&
      originalProjectNumber !== null &&
      filterValue !== undefined &&
      filterValue !== null &&
      filterValue !== ''
    ) {
      // Try numeric comparison if both are numbers or can be parsed as numbers
      const origNum = Number(originalProjectNumber);
      const filterNum = Number(filterValue);
      if (!Number.isNaN(origNum) && !Number.isNaN(filterNum)) {
        projectNumberMatch = origNum === filterNum;
      } else {
        // fallback to string includes for partial match
        projectNumberMatch = originalProjectNumber
          .toString()
          .includes(filterValue.toString());
      }
    }
```

rafasdc: These are currently always numbers or might be null or undefined, added a few more cases in case somehow it is a string due to imports from Excel. This makes search work for original project number but still missing the highlighting part.

image

rafasdc (changes requested):

rafasdc (approved):

PR #4009, opened 2025-05-15
feat: logout redirect based on roles

rafasdc:
Implements [NDT-757]


RRanath (approved):
Approved 🚀

PR #4023, opened 2025-05-20
Ndt 798 make expanded view of map size customizable

rafasdc:
Implements [NDT-798]


RRanath (approved):
Looks Good! Approved 🚀

PR #4179, opened 2025-08-05
fix: skip unsaved warning on project navigation

rafasdc:
Implements [NDT-961]


app/components/Analyst/ProjectNavigationSidebar.tsx
RRanath, line 1: Issue in the ticket step 4 `Then navigate to a different page`(using left nav bar) still seems to exists.

RRanath (approved):
Looks Good 😎

PR #4194, opened 2025-08-14
Ndt 1024 modify data imported from claims report

rafasdc:
Implements [NDT-1024]


RRanath (approved):
🚀Approved

